PR #4, opened 2020-09-24
Document to ensure the gem is added to production environment

wildmaples:

Clarify that Packwerk needs to be grouped in the production environment in the gemfile. 

https://github.com/Shopify/shopify/pull/261333#discussion_r492978297

Update documentation in README and USAGE. 

## Type of Change

### Additional Release Notes

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,8 @@ Watch a [1-minute video demo](https://www.youtube.com/watch?v=NwqlyBAxVpQ&featur
 gem 'packwerk'
 ```
 
+_Note: Packwerk has to be grouped in production environment as it contains logic to load custom inflectors._
+
 2. Install the gem
 
 Execute:
--- a/USAGE.md
+++ b/USAGE.md
@@ -43,6 +43,8 @@ After including Packwerk in the Gemfile, you can generate the necessary files to
 
     bundle exec packwerk init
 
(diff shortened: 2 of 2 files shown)


rafaelfranca (approved):

paracycle (approved):

README.md
```diff
@@ -28,6 +28,8 @@ Watch a [1-minute video demo](https://www.youtube.com/watch?v=NwqlyBAxVpQ&featur
 gem 'packwerk'
 ```
 
+_Note: Packwerk has to be grouped in production environment as it contains logic to load custom inflectors._
```
tomstuart, line 31: I’m too late for this PR, but this advice is only true if the app uses custom inflections, which I think we should be clear on so that people don’t add it as a production dependency unnecessarily. My guess would be that most Rails apps don’t configure any custom inflections.

tomstuart: In fact, I think this advice should live in the [Inflections section](https://github.com/Shopify/packwerk/blob/main/USAGE.md#inflections) of `USAGE.md`, since it’s adding that initializer which creates the production dependency on `Packwerk::Inflections::Custom`.

exterm: @wildmaples can you follow up on this? If you have already, ignore me

PR #21, opened 2020-09-26
Ignore paths from vendored gems inside the Rails root

tonymarklove:

Packwerk already ignores autoload paths outside of the Rails root directory, which means paths added by gems are usually ignored. However, when gems are packaged into the `vendor/cache` directory, this sits inside the Rails root, meaning that gem-added paths are not ignored. This change makes sure that these paths are ignored.

I have extended the code which selects paths in the Rails root directory to exclude paths which also match the load path returned by `Bundler.bundle_path`.

This is where the existing filtering was taking place, so it seems like a sensible place to add an additional check.

Testing this is not very simple, since it relies on building an app with vendored gems. The approach I have taken is to stub `Bundler.bundle_path` to return a path under the Rails root. I have tested this change against a real app with vendored gems and `Bundler.bundle_path` seems to be the correct method to use. But any thoughts on making the automated test more robust, and less reliant on stubbing Bundler, would be useful.

## Type of Change

### Additional Release Notes

## Checklist

### DIFF
--- a/lib/packwerk/configuration.rb
+++ b/lib/packwerk/configuration.rb
@@ -28,7 +28,7 @@ module Packwerk
 
     DEFAULT_CONFIG_PATH = "packwerk.yml"
     DEFAULT_INCLUDE_GLOBS = ["**/*.{rb,rake,erb}"]
-    DEFAULT_EXCLUDE_GLOBS = ["{bin,node_modules,script,tmp}/**/*"]
+    DEFAULT_EXCLUDE_GLOBS = ["{bin,node_modules,script,tmp,vendor}/**/*"]
 
     attr_reader(
       :include, :exclude, :root_path, :package_paths, :custom_associations, :load_paths, :inflections_file,
--- a/lib/packwerk/configuration.rb
+++ b/lib/packwerk/configuration.rb
@@ -58,18 +58,19 @@ module Packwerk
         (engine.config.autoload_paths + engine.config.eager_load_paths + engine.config.autoload_once_paths).uniq
       end
 
-      rails_root_match = Rails.root.join("**").to_s
       bundle_path_match = Bundler.bundle_path.join("**").to_s
 
-      all_paths = all_paths.map do |path_string|
-        # ignore paths outside of the Rails root and vendored gems
+      all_paths = all_paths.each_with_object([]) do |path_string, paths|
+        # ignore paths inside gems
         path = Pathname.new(path_string)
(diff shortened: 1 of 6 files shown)


lib/packwerk/configuration.rb
```diff
@@ -58,10 +58,13 @@ def all_application_autoload_paths
         (engine.config.autoload_paths + engine.config.eager_load_paths + engine.config.autoload_once_paths).uniq
       end
 
+      rails_root_match = Rails.root.join("**").to_s
+      bundle_path_match = Bundler.bundle_path.join("**").to_s
+
       all_paths = all_paths.map do |path_string|
-        # ignore paths outside of the Rails root
+        # ignore paths outside of the Rails root and gems
```
exterm, line 65:
```suggestion
        # ignore paths outside of the Rails root and vendored gems
```

exterm:
Interesting one - we don't usually vendor gems so didn't encounter this.

Thank you for your contribution Tony.

lib/packwerk/configuration.rb
```diff
@@ -58,10 +58,13 @@ def all_application_autoload_paths
...
         path = Pathname.new(path_string)
-        if path.exist? && path.realpath.fnmatch(Rails.root.join("**").to_s)
+        if path.exist? && path.realpath.fnmatch(rails_root_match) && !path.realpath.fnmatch(bundle_path_match)
```
exterm, line 67: Heh. We introduced the restriction to paths under `Rails.root` mostly to eliminate gem load paths, but now we have a separate clause for gem load paths.

In this case do we even still need the `rails_root_match`?

tonymarklove: Good point. I can't think of anything other than gems which could add themselves to the load paths, so simply ignoring the `bundle_path` might be sufficient.

tonymarklove: I have tried this out on a large project which vendors gems. You are correct that now we are checking the bundle path, there is no need for the separate check of `Rails.root`. I have added a new commit with that change.

I have also added the `vendor` directory to the default `exclude` list. I can't think of a reason you'd want Packwerk to scan files inside the gems.

exterm:
We should test this on our core monolith before merging.

wildmaples:
@exterm this change looks good on the core monolith 👍

exterm (approved):

shioyama (approved):

Issue #17, opened 2020-09-25
Add unit tests for ApplicationValidator checks


wildmaples:
We have `ApplicationValidatorTest` which are integration tests. However, we should find a way to unit test the individual checks. 

We could make unit testing this file easier by refactoring it so a fake filesystem can be injected. 
This would make it easier to test but would require a larger refactor. https://github.com/Shopify/packwerk-old/pull/262#discussion_r455956863

TO DO:

- [ ] check_autoload_path_cache,
- [ ] check_package_manifests_for_privacy,
- [ ] check_package_manifest_syntax,
- [ ] check_application_structure,
- [ ] check_inflection_file,
- [ ] check_acyclic_graph,
- [ ] check_package_manifest_paths,
- [ ] check_valid_package_dependencies,
- [ ] check_root_package_exist,

rochlefebvre:
Addressed in #84

PR #24, opened 2020-09-29
Turn expected and actual inflections into a set.

gy:
Fix a bug in the `bin/packwerk validate` that incorrectly reports a mismatch in inflections. This scenario seems to occur when the inflection has been previously defined from a gem.

For example:
graphql-rails gem [defines a GraphiQL acronym here](https://github.com/rmosolgo/graphiql-rails/blob/0dee95827c5f1661238290f5873f70ea2bae413a/lib/graphiql/rails.rb#L6). The following inflection config will fail

```ruby
acronym:
  - "OAuth"
  - "GraphiQL" # <- this needs to come first.
```

Convert inflections into a set instead of an array to avoid false validation errors. Sorting the array won't work due to its heterogeneous nature.

Testing this is a little tricky. One simple but I think bad idea is to load a [second fixture file](https://github.com/Shopify/packwerk/blob/master/test/unit/application_validator_test.rb#L63) that contains the same acronyms in a different order.

A better but more involved approach is to break down the [application validator into individual validators ](https://github.com/Shopify/packwerk/blob/master/lib/packwerk/application_validator.rb#L26) which can help us write unit  tests.

## Type of change

## Checklist

### DIFF
--- a/lib/packwerk/application_validator.rb
+++ b/lib/packwerk/application_validator.rb
@@ -170,8 +170,8 @@ module Packwerk
       Packwerk::Inflections::Custom.new(inflections_file).apply_to(test_inflections)
 
       results = %i(plurals singulars uncountables humans acronyms).map do |type|
-        expected = ActiveSupport::Inflector.inflections.public_send(type).to_a
-        actual = test_inflections.public_send(type).to_a
+        expected = ActiveSupport::Inflector.inflections.public_send(type).to_set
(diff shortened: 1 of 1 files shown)


exterm (approved):

wildmaples (approved):
Yeah, we need unit tests for these validators -  https://github.com/Shopify/packwerk/issues/17 

Thanks for your patience on this @gy

Issue #19, opened 2020-09-25
Improve message output for parsing errors


wildmaples:
While trying to fix a parsing error shown on Packwerk, we had to run rubocop locally since Packwerk did not display any specifics of the error (line number, for example).

PR #32, opened 2020-10-06
Pass the file path to the formatter

ngan:
I can't get the `check` to finish on our repo because there are parsing errors (on both packwerk and our code). But when it errors, I can't tell where it's at. I've been adding `puts` to the cli. While I was there, I thought it'd make sense for the cli to pass along the path to the formatter in case someone wants to make a document-style formatter.

This is similar to how RSpec (and others) formatter works.

## Type of Change

- [x] New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/lib/packwerk/cli.rb
+++ b/lib/packwerk/cli.rb
@@ -130,7 +130,7 @@ module Packwerk
       all_offenses = T.let([], T.untyped)
       execution_time = Benchmark.realtime do
         all_offenses = files.flat_map do |path|
-          @run_context.file_processor.call(path).tap { |offenses| mark_progress(offenses) }
+          @run_context.file_processor.call(path).tap { |offenses| mark_progress(path, offenses) }
         end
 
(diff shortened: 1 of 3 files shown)


tomstuart:
Thank you for your contribution @ngan! It’s a good idea but I’m inclined not to merge it yet because we don’t know whether anyone is planning to contribute a formatter that needs this information, and in the meantime it’s not serving any purpose in the codebase. Do you have any immediate plans to use it?

ngan:
> because we don’t know whether anyone is planning to contribute a formatter that needs this information

That's fair! I was running into parsing errors and was trying to find the cleanest place to `puts` the error info (line num, etc) and I thought it was best in the formatter.

I'll take a crack at this: https://github.com/Shopify/packwerk/issues/19 since we actually need it...

tomstuart:
Cool, thanks. I’ll close this PR for now to help us with triage, but we can reopen later if you need it to be merged for whatever comes out of #19.

Issue #36, opened 2020-10-13
[Bug Report]


apeacock1991:
**Description**

Currently the public folder is hard-coded as `app/public`, which may not work for all users of the gem. They may wish to define their own public folder, and if none is defined, default to `app/public`.

** Why? **

Depending on the architecture of the application, `app/public` might just not make sense, or at least not be cohesive with the rest of the folder structure. Consider a layered architecture for example, where you break up the layers of your application via technical partition (i.e. presentation layer, domain layer, infrastructure layer etc.). Your top-level folders in your package would typically be `/presentation`, `/domain` etc., so logically in that architecture you would want to select the layer that makes sense to expose the package from, and `app/public` conflicts with that vision.

I've raised a PR [here]() that I believe introduces that feature, and seems to work nicely.

wildmaples:
This feature request as been merged in https://github.com/Shopify/packwerk/pull/37

PR #37, opened 2020-10-13
Add the ability to specify a custom public path for a package

apeacock1991:

See [this issue](https://github.com/Shopify/packwerk/issues/36) for details.

tl;dr I would like to be able to configure what a package's public path is.

The config was already available in the Package class, so I've simply added the documentation for that, amended the validator for this new parameter, and then used the user-defined public path if it's set, else the default.

Have I fully covered the tests required?

## Type of Change

- [x] New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/lib/packwerk/application_validator.rb
+++ b/lib/packwerk/application_validator.rb
@@ -113,7 +113,7 @@ module Packwerk
         hash = YAML.load_file(f)
         next unless hash
 
-        known_keys = %w(enforce_privacy enforce_dependencies dependencies metadata)
+        known_keys = %w(enforce_privacy enforce_dependencies public_path dependencies metadata)
         unknown_keys = hash.keys - known_keys
 
         unless unknown_keys.empty?
@@ -131,6 +131,16 @@ module Packwerk
         if hash.key?("enforce_dependencies")
           unless [TrueClass, FalseClass].include?(hash["enforce_dependencies"].class)
             errors << "Invalid 'enforce_dependencies' option in #{f.inspect}: #{hash['enforce_dependencies'].inspect}"
+          end
+        end
+
+        if hash.key?("public_path")
+          unless hash["public_path"].is_a?(String)
+            errors << "'public_path' option must be a string in #{f.inspect}: #{hash['public_path'].inspect}"
+          end
+
+          unless hash["public_path"].try(:last) == "/"
+            errors << "'public_path' option must end with a forward slash #{f.inspect}: #{hash['public_path'].inspect}"
           end
         end
 
--- a/lib/packwerk/application_validator.rb
+++ b/lib/packwerk/application_validator.rb
@@ -138,10 +138,6 @@ module Packwerk
           unless hash["public_path"].is_a?(String)
             errors << "'public_path' option must be a string in #{f.inspect}: #{hash['public_path'].inspect}"
           end
-
-          unless hash["public_path"].try(:last) == "/"
-            errors << "'public_path' option must end with a forward slash #{f.inspect}: #{hash['public_path'].inspect}"
-          end
         end
 
         next unless hash.key?("dependencies")
--- a/lib/packwerk/generators/templates/package.yml
+++ b/lib/packwerk/generators/templates/package.yml
@@ -13,7 +13,7 @@ enforce_privacy: false
 
 # By default the public path will be app/public/, however this may not suit all applications' architecture so
 # this allows you to modify what your package's public path is.
-public_path: app/public/
+# public_path: app/public/
 
 # A list of this package's dependencies
 # Note that packages in this list require their own `package.yml` file
--- a/lib/packwerk/generators/templates/package.yml
+++ b/lib/packwerk/generators/templates/package.yml
@@ -11,6 +11,10 @@ enforce_dependencies: true
 # We recommend enabling this for any new packages you create to aid with encapsulation.
 enforce_privacy: false
 
+# By default the public path will be app/public/, however this may not suit all applications' architecture so
+# this allows you to modify what your package's public path is.
+public_path: app/public/
+
 # A list of this package's dependencies
 # Note that packages in this list require their own `package.yml` file
 # dependencies:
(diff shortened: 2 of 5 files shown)


lib/packwerk/application_validator.rb
```diff
@@ -134,6 +134,16 @@ def check_package_manifest_syntax
           end
         end
 
+        if hash.key?("public_path")
+          unless hash["public_path"].is_a?(String)
+            errors << "'public_path' option must be a string in #{f.inspect}: #{hash['public_path'].inspect}"
+          end
+
+          unless hash["public_path"].try(:last) == "/"
```
thegedge, line 142: `last` doesn't look like a core string method (https://ruby-doc.org/core-2.6/String.html) so, depending on the environment this runs in, this could always raise a `NoMethodError`. Best to index by -1 or `.chars.last`.

That being said, I'm thinking we should drop this anyways, since it doesn't guarantee a directory on disk. If the slash is necessary elsewhere in packwerk (I think so, because of [this](https://github.com/Shopify/packwerk/blob/da99ef83da6ba9fd0879c85501c6bd7ed4fa9ec0/lib/packwerk/package.rb#L40)), we should ensure it's there (perhaps in `Package#initialize`). I think this provides a slightly better user experience.

What do you think?

thegedge (commented):
👍 we discussed this early in the packwerk days. I believe we opted to not implement it because we wanted to be opinionated internally. Now that we're open source, I think this is a good configuration value to have 🙂

lib/packwerk/application_validator.rb
apeacock1991: Ah, good spot on the `#last`! Thanks for the review 🤓 

As for the validation... I assumed that the code I amended in application_validator would run on startup or something, is that not the case? And it's just run when the validate command is used via the CLI?

If it's not run outside of that, then I agree it would be nice to fail loudly, perhaps something like:

```
    def validate_public_path!
      return unless @config["public_path"]

      raise PublicPathMustEndInForwardSlash unless @config["public_path"].chars.last == "/"
    end
```

In package.rb, which is then called from the initializer?

I presume if we go with this, we'd also want to keep the validation in the application_validator too so that if people run that without the slash, it also intuitively lets them know?

thegedge: > As for the validation... I assumed that the code I amended in application_validator would run on startup or something, is that not the case? And it's just run when the validate command is used via the CLI?

You got it! This only gets executed when one runs `packwerk validate`. Just to explain why there's a separation here: `packwerk validate` uses Rails application state to help validate the configuration. `packwerk check` and `packwerk update` don't need to run with application state, which means they boot up and run much more quickly.

> In package.rb, which is then called from the initializer?

As I mentioned, I think it may be best to not even check for the slash but simply add it if it's missing. Imagine a scenario where it was configured to be `my_public`, but `my_public` was either a file or not the right directory.

First, we'd fail loudly saying you need to add a `/`, after which `packwerk check` would likely fail because the `start_with?` check (linked in my previous comment) would always fail. Once we correct that mistake, we'd likely have many privacy violations. It may be misleading to the end user to think that the validation is protecting them in case of mistake.

That being said, I can't think of a _great_ solution here. We _could_ check for the directory's existence (under the package path) when validation happens. The problem there is that maybe it's configured and there's no public directory at all? We could suggest, in the validator, to create the directory (possibly adding an empty file in there to ensure it is made available by your VCS). Every solution I can think of has pros and cons 🙈 

I'll poke someone on my team to chime in too. It looks like there's no perfect solution, so let's aim for at least a couple of people to agree on some approach, and go with that 🙂

apeacock1991: Personally I would say they are two different issues (and thus _probably_ two PRs). I think adding the slash in automatically makes sense, because before it was handled by the fact it was hard-coded, so we then have the same behavior with the custom paths.

Then the secondary quality of life change is around making sure that public folder exists, as realistically I imagine it's currently possible for people to not create the `app/public/` folder and it would have the same issue as if a custom path was used and that folder didn't exist (unless I am missing something, which is possible as I am new to the codebase 😅 ).

apeacock1991: I've pushed a change to automatically add the slash if it's missing, and removed the `/` handling in the validator (as there is no real need to loudly fail now we're handling it)

lib/packwerk/generators/templates/package.yml
```diff
@@ -11,6 +11,10 @@ enforce_dependencies: true
 # We recommend enabling this for any new packages you create to aid with encapsulation.
 enforce_privacy: false
 
+# By default the public path will be app/public/, however this may not suit all applications' architecture so
+# this allows you to modify what your package's public path is.
+public_path: app/public/
```
tomstuart, line 16: My gut feeling is that this should be commented out rather than set to the default. WDYT @thegedge?

lib/packwerk/application_validator.rb
tomstuart: I think it’s okay to configure a public folder path that doesn’t yet exist on disk. It’s a little weird but shouldn’t be treated as any more “wrong” than not having an `app/public` folder, which is fine.

There’s a deeper problem here which is that we shouldn’t really be manipulating paths with low-level string operations like concatenation and `#start_with?`, but that’s outside of the scope of this PR. I think the implementation looks good!

lib/packwerk/generators/templates/package.yml
thegedge: 👍 Yeah, commented out would be good.

I think it makes sense to have it not commented out if there's no default in code, but (for the time being) we will want to keep the default in code (a lot of internal things will depend on that). It's also not useful unless `enforce_privacy: true` was the default 🙂

(sorry, I type a lot of words to say I agree...)

thegedge (approved):
Just one little comment on the configuration template, but implementation is great. Thanks for your contribution, @apeacock1991 ! ❤️

apeacock1991:
@thegedge Thanks, change made! I read somewhere about squashing commits, is that something whoever merges can do? (which I believe needs to be someone from Shopify 😅 )

lib/packwerk/generators/templates/package.yml
apeacock1991: Done!

tomstuart (approved):

PR #38, opened 2020-10-14
Ignore package.yml found inside vendored gems

tonymarklove:

Following up on my previous PR (https://github.com/Shopify/packwerk/pull/21), which prevented loading of paths inside the `vendor/cache` directory, the `packwerk validate` command can still fail due to the presence of `package.yml` files inside gems.

Searching for `package.yml` files should follow the same rules, and ignore anything inside the Bundler gem directory.

I made `PackageSet.package_paths` public so it could be shared with `ApplicationValidator` to avoid duplication. Then made sure that any paths inside `Bundler.bundle_path` are excluded from the set of package file paths.

I also noticed that `RunContext` was not correctly passing along the `package_paths` configuration option. So there is a one-line fix for that, plus 2 test stub changes to support it.

## Type of Change

## Checklist

### DIFF
--- a/lib/packwerk/application_validator.rb
+++ b/lib/packwerk/application_validator.rb
@@ -325,7 +325,7 @@ module Packwerk
     end
 
     def package_manifests(glob_pattern)
-      Dir.glob(File.join(glob_pattern, Packwerk::PackageSet::PACKAGE_CONFIG_FILENAME)).map { |f| File.realpath(f) }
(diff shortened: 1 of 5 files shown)


tomstuart (approved):
Thank you for this, @tonymarklove!

The change looks good to me. I’ve tested it against the Shopify codebase and didn’t see any changes in behaviour. 👍

thegedge (approved):

Issue #15, opened 2020-09-25
Extract constant names from fixtures


wildmaples:
## What?

In test files, when we see:

```ruby
class MyTest < ActiveSupport::TestCase
  setup do
    @thing = things(:thing_fixture)
  end
end
```

We want to extract the constant name `::Thing` from the call `things(:thing_fixture)`.

## Why?

Tests should also be appropriately packaged (in other words, no boundary violations), and fixtures are an easy way for packages to (currently) depend on other packages without packwerk realizing. By registering these constant references we have a better signal to help packwerkers have proper isolation for their tests.

## How?

I'm thinking this will be a few PRs:

1. Introduce an interface for getting a constant name from a node.
2. Refactor `ReferenceExtractor` to accept a list of instances of anything that implements the "constant name from node" interface in (1).
3. Extract existing "constant name from node" from `ReferenceExtractor` (the `node.const_node?` path).
4. Add new class to get constant name from a node if the node looks like a fixtures call.

For (4), to reduce false positives, I'm thinking the criteria for a "fixtures function call" would be the following:

- Directory of the file being parsed matches `**/test/**/*_test.rb` or `**/spec/**/*_spec.rb`
- AST node is of type `:send` with one or more arguments that are all symbols, and
- Name of function maps to a file in (a configurable) fixtures path.
  - We can preload all fixtures files to get the model class name too (via `_fixture.model_class`, if it exists, otherwise just use the inflector on the filename).

cc: @thegedge

doodzik:
[We tried to add fixture detection into Packwerk](https://github.com/Shopify/packwerk/pull/50) because it is responsible for ensuring that code has no boundary violations. During the implementation process, I noticed that this approach is tying Packwerk directly with a fixture implementation. To detect the fixtures, Packwerk has to assume the name of a fixture method, how many arguments it takes, and what type of arguments. 

# s(:send, nil, :shop,
#   s(:sym, :snowdevil))
sig { params(node: ::AST::Node).returns(T::Boolean) }
def potential_fixture_call?(node)
   arguments = Node.method_arguments(node)
   arguments.length == 1 && Node.symbol?(arguments[0])
end

[...]

PR #39, opened 2020-10-21
Clarify ConstNodeInspector code

doodzik:

I was going through the code to implement #15 and I had a hard time following the code in the ConstNodeInspector class. As such I refactored it to be a bit more self-explanatory.

## Type of Change

## Checklist


lib/packwerk/association_inspector.rb
```diff
@@ -1,13 +1,13 @@
...
 # frozen_string_literal: true
 
-require "packwerk/constant_name_inspector"
+require "packwerk/constant_name_inspector_interface"
 require "packwerk/node"
 
 module Packwerk
   # Extracts the implicit constant reference from an active record association
   class AssociationInspector
-    include ConstantNameInspector
```
doodzik, line 10: I found it really confusing that the interfaces aren't postfixed with interface and made it hard to understand the code initially.

lib/packwerk/const_node_inspector.rb
```diff
@@ -1,44 +1,54 @@
...
+      name = Node.parent_module_name(ancestors: ancestors)
+      name ||= generate_qualified_constant(node, ancestors)
+      "::" + name
+    end
+
+    def generate_qualified_constant(node, ancestors:)
+      namespace_path = Node.enclosing_namespace_path(node, ancestors: ancestors)
+      constant_name = Node.constant_name(node)
+      namespace_path.push(constant_name).join("::")
+    end
```
doodzik, line 52: Some of these private methods could live in another class, but tackling that issue in this PR would increase the scope of the PR too much.

lib/packwerk/reference_extractor.rb
```diff
@@ -17,7 +17,7 @@ class ReferenceExtractor
     sig do
       params(
         context_provider: Packwerk::ConstantDiscovery,
-        constant_name_inspectors: T::Array[Packwerk::ConstantNameInspector],
```
doodzik, line 20: When I did my first pass through the code, not having the interface postfix for this constant made the code harder to understand. So I'm adding one here.

tomstuart:
Do we normally use an `…Interface` suffix for modules which provide Sorbet interfaces?

exterm:
> do we normally use an …Interface suffix for modules which provide Sorbet interfaces?

I don't think a convention exists, at least at Shopify, and I find it very confusing. IMO, we should have a different method to include it (`implement` instead of `include`), but that would have to be supplied by sorbet.

doodzik:
> I don't think a convention exists, at least at Shopify, and I find it very confusing. IMO, we should have a different method to include it (implement instead of include), but that would have to be supplied by sorbet.

> The problem is not in the interface definition; it's obvious from the first few lines of the modules that it's an interface. It's in implementing the interface. Having a more specific method to do that would be better than an Interface suffix on the class name IMO. 

We don't have a company convention around this, but in my previous team, it was a team convention. My guiding principle here is that we want to write our code that is easy to understand. When reading the code that includes an interface, it is not clear what is happening. We could either include methods from a module or implement an interface. Granted that this is a minor inconvenience, those add up and contribute to code that is hard to follow.

Having a dedicated keyword would be the best solution, but that doesn't strike me as the most practical solution here. I could try to add it to sorbet if this is something we desire. @Morriar Do you think this would be a good addition?

> We don't use Class and Module suffixes either.

Yes, we don't suffix Classes nor Modules, but those constructs are first-class citizens in Ruby. Interfaces are not. We are following similar conventions for constructs like Controllers, Services, and Views. The only exceptions are Models, which is possible because every other class is suffixed in rails and the models live in a dedicated model folder.

doodzik:
I opened an issue for the interface issue on sorbet https://github.com/sorbet/sorbet/issues/3578. For the time being, I'm going to drop the changes to the interface in the PR. I also split the commit into multiple commits as @tomstuart suggested.

lib/packwerk/const_node_inspector.rb
```diff
@@ -9,12 +9,12 @@ class ConstNodeInspector
     include ConstantNameInspector
 
     def constant_name_from_node(node, ancestors:)
-      return nil unless Node.type(node) == Node::CONSTANT
+      return nil unless Node.constant?(node)
```
tomstuart, line 12: There’s a fair amount of discussion about this design decision on the (Shopify-only) Shopify/packwerk-old#266 PR. The idea is that there were many places in Packwerk where we wanted to test for multiple node types, e.g. “if this is a class or module or constant assignment, then …, otherwise …”, so the API was designed to be usable in a `case` statement:

```ruby
case Node.type(node)
when Node::CLASS, Node::MODULE, Node::CONSTANT_ASSIGNMENT
  # …
when Node::BLOCK
  # …
else
  # …
end
```

If we provide predicates instead then this becomes a slightly less convenient `if Node.class?(node) || Node.module?(node) || Node.constant_assignment?(node)`. I was reluctant to provide both because that would mean our `Packwerk::Node` API offers two different ways of doing the same thing, which makes it (again slightly) less straightforward to use.

Looking at the code now, I can see that essentially all of these `case` statements have been moved into `Packwerk::Node` itself, so perhaps there’s a case for making `Node.type` private and exposing only predicates to the rest of Packwerk. WDYT?

lib/packwerk/const_node_inspector.rb
```diff
@@ -40,5 +35,19 @@ def constant_in_module_or_class_definition?(node, parent:)
         parent_name && parent_name == Node.constant_name(node)
       end
     end
+
+    def fully_qualify_constant(node, ancestors:)
```
tomstuart, line 40: Extracting these private methods is a good idea. I would love to suggest better names for them to more clearly describe what they’re doing, but I actually don’t understand what the code does in the first place (why is it sometimes returning a “parent module name” generated only from the ancestor nodes, but sometimes a ”qualified constant” name generated from the node itself?) so can’t begin to think of anything better. Fixing the deeper problem is a job for another PR. 😄

lib/packwerk/const_node_inspector.rb
```diff
@@ -10,13 +10,11 @@ class ConstNodeInspector
 
     def constant_name_from_node(node, ancestors:)
       return nil unless Node.constant?(node)
-
       # Only process the root `const` node for namespaced constant references. For example, in the
       # reference `Spam::Eggs::Thing`, we only process the const node associated with `Spam`.
-      parent = ancestors.first
-      return nil if parent && Node.constant?(parent)
+      return nil if root_constant?(ancestors)
```
tomstuart, line 15: The name `root_constant?` is the opposite of what this check is doing — as the comment says, we want to ignore every node _except_ `Spam` in `Spam::Eggs::Thing`, so we return `nil` if we find ourselves trying to process a node like `Eggs` or `Thing` whose parent is another constant. If you want to keep the name `root_constant?` then the conditional should be `unless root_constant?` and the behaviour of the predicate should be reversed.

lib/packwerk/const_node_inspector.rb
```diff
@@ -10,13 +10,11 @@ class ConstNodeInspector
...
 
-      if constant_in_module_or_class_definition?(node, parent: parent)
+      if constant_in_module_or_class_definition?(node, ancestors: ancestors)
```
tomstuart, line 17: I don’t think this part of the change is worth it, because we’re now intentionally passing more information into `constant_in_module_or_class_definition?` than it needs to do its job. For the caller it’s useful to be able to see at a glance that this decision depends only upon the parent node, not the whole ancestor chain, and for the implementer it’s helpful to be prevented from accidentally poking around in the rest of the ancestor chain unless you really mean it (and deliberately change the arguments to reflect that). WDYT?

lib/packwerk/const_node_inspector.rb
```diff
@@ -17,12 +17,10 @@ def constant_name_from_node(node, ancestors:)
...
         fully_qualify_constant(node, ancestors: ancestors)
       else
-        begin
-          Node.constant_name(node)
-        rescue Node::TypeError
-          nil
-        end
+        Node.constant_name(node)
       end
+    rescue Node::TypeError
```
tomstuart, line 22: I’m on the fence about this one. The `rescue` was [originally added](https://github.com/Shopify/packwerk-old/commit/8315ad638d2ad414dd43569ac270542d92f4594d) to deal with an occurrence of `self.class::HEADERS` in core. I agree that `Node.constant_name` can raise `TypeError` elsewhere in this method but that’s not something we’ve seen happen so I’m not able to decide whether returning `nil` would be the right thing to do in those unknown cases.

tomstuart (commented):
Thank you for doing this @doodzik! It’s a positive change and I’m glad to see the code becoming clearer.

In a few places I’ve added some historical context which I hope helps you to decide whether you want to commit to these changes. It’s all just food for thought; let me know what you think and I’ll come back and approve if you’re happy.

lib/packwerk/const_node_inspector.rb
doodzik: I'm going to remove this change from the PR because another PR was opened that deals with this issue more concretely https://github.com/Shopify/packwerk/pull/54

lib/packwerk/const_node_inspector.rb
doodzik: Agreed, I'm providing too much in the private method and making the intention behind the code harder to understand 👍

lib/packwerk/const_node_inspector.rb
doodzik: Agreed, I'm not confident in why the code does what it does. I think it would be good to go through the test cases and document how the data gets manipulated in those methods. I would add to this that we can probably extract those private methods into an Ancestor module in a separate PR and document it then. WDYT?

lib/packwerk/const_node_inspector.rb
doodzik: > API offers two different ways of doing the same thing

Agreed, I think it makes sense to keep the API as straightforward to use as possible.

> Looking at the code now, I can see that essentially all of these case statements have been moved into Packwerk::Node itself, so perhaps there’s a case for making Node.type private and exposing only predicates to the rest of Packwerk. 

I did a quick [experiment to verify that we use it only in `Packwerk::Node`](https://github.com/Shopify/packwerk/pull/55/commits/84746c5b4d7772e869311ceb5b380bd527ca8d15) and we do. So it shouldn't be an issue to change `Node.type` to be private.

Looking at the Node Module and how it uses the switch statements it seems to me that we are missing a Factory here. What I would suggest for a separate PR is to move as much logic out of the Node module into dedicated modules converting the Node module into a Factory module. And then as you mentioned change the `Node.type` method to be private. I think this would clean up the code and simplify the API of `Packwerk::Node`. WDYT?

lib/packwerk.rb
```diff
@@ -14,7 +14,6 @@
 require "packwerk/configuration"
 require "packwerk/const_node_inspector"
 require "packwerk/constant_discovery"
-require "packwerk/constant_name_inspector"
```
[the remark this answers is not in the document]
doodzik: I prefer requiring a file where it is actually used because it makes the dependencies clearer. In this case, the `requires` in `lib/package.rb` would only contain the entry points into the application and what we intend to expose to our users. This also means that you will not get into a situation where you don't know what any of the `requires` is for. 

I think there is an argument for convenience by requiring every file in the `lib/package.rb` file, but I think having clearly defined `requires` makes the codebase easier to understand in the long run.

I would like us to align one approach since this allows us to be consistent in the codebase, but I'm happy to go with either approach. So thank you for raising this issue :)

lib/packwerk/const_node_inspector.rb
tomstuart: > Looking at the Node Module and how it uses the switch statements, it seems to me that we are missing a Factory here. I would suggest a separate PR to move as much logic out of the Node module into dedicated modules, converting the Node module into a Factory module.

When we discussed this on a call, I think we agreed that a) it’s a good idea but b) you’re not going to do it at the moment.

lib/packwerk/const_node_inspector.rb
tomstuart: Again, it’s a good idea — let’s decide separately whether it’s worth doing in another PR.

lib/packwerk/const_node_inspector.rb
```diff
@@ -29,11 +26,15 @@ def constant_name_from_node(node, ancestors:)
 
     private
 
+    # Only process the root `const` node for namespaced constant references. For example, in the
+    # reference `Spam::Eggs::Thing`, we only process the const node associated with `Spam`.
+    def root_constant?(parent)
+      !parent || !Node.constant?(parent)
```
tomstuart, line 32: For me this would be 1% easier to understand as `!(parent && Node.constant?(parent))` (per [de Morgan](https://en.wikipedia.org/wiki/De_Morgan%27s_laws)) but there’s not much in it.

tomstuart (approved):
Looks good. Thanks for your patience! 🙏

Morriar:
> Having a dedicated keyword would be the best solution, but that doesn't strike me as the most practical solution here. I could try to add it to sorbet if this is something we desire. @Morriar Do you think this would be a good addition?

I agree with the answer given by the Sorbet team on this: it would be hard to enforce using the static checker for retro-compatibility purposes so we would end up with the two forms in our codebase which would be super confusing for the developers.

PR #42, opened 2020-10-22
Clarify AssociationInspector class

doodzik:

I was going through the code to implement https://github.com/Shopify/packwerk/issues/15 and I had a hard time following the code in the AssociationInspector class. As such I refactored it to be a bit more self-explanatory.

## Type of Change

## Checklist

### DIFF
--- a/lib/packwerk/association_inspector.rb
+++ b/lib/packwerk/association_inspector.rb
@@ -26,14 +26,11 @@ module Packwerk
       return unless association?(node)
 
       arguments = Node.method_arguments(node)
-      association_name = Node.literal_value(arguments[0]) if Node.symbol?(arguments[0])
-      return nil unless association_name
+      return unless (association_name = association_name(arguments))
 
-      association_options = arguments.detect { |n| Node.hash?(n) }
-      class_name_node = Node.value_from_hash(association_options, :class_name) if association_options
-
-      if class_name_node
-        Node.literal_value(class_name_node) if Node.string?(class_name_node)
+      if (class_name_node = custom_class_name(arguments))
+        return unless Node.string?(class_name_node)
+        Node.literal_value(class_name_node)
       else
         @inflector.classify(association_name.to_s)
       end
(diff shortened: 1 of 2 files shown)


tomstuart:
This is probably a good change but I can’t tell from the commit. Are you able to break it into separate commits with messages that explain the changes you’re making?

lib/packwerk/association_inspector.rb
```diff
@@ -22,20 +22,20 @@ def initialize(inflector: Inflector.new, custom_associations: Set.new)
     end
 
     def constant_name_from_node(node, ancestors:)
-      return unless Node.type(node) == Node::METHOD_CALL
+      return unless Node.method_call?(node)
```
tomstuart, line 25: My [comment on #39](https://github.com/Shopify/packwerk/pull/39#discussion_r513322686) delves into this a bit. If we can make these predicates the only type-related API that `Node::Packwerk` exposes then I’m in favour, but I’m unsure about adding them while also keeping `Node.type`.

lib/packwerk/association_inspector.rb
```diff
@@ -23,9 +23,7 @@ def initialize(inflector: Inflector.new, custom_associations: Set.new)
 
     def constant_name_from_node(node, ancestors:)
       return unless Node.method_call?(node)
-
-      method_name = Node.method_name(node)
-      return nil unless @associations.include?(method_name)
+      return unless association?(node)
```
tomstuart, line 26: Cool! 😎

lib/packwerk/association_inspector.rb
```diff
@@ -45,5 +42,18 @@ def association?(node)
       method_name = Node.method_name(node)
       @associations.include?(method_name)
     end
+
+    def association_name_when_custom_class_name(arguments)
```
tomstuart, line 46: This method returns a const node containing the class name, not the association name, so its name is a little misleading. How about something like `#custom_class_name_for_association` or just `#custom_class_name`?

lib/packwerk/association_inspector.rb
```diff
@@ -45,5 +42,18 @@ def association?(node)
...
+      association_options = arguments.detect { |n| Node.hash?(n) }
+      return unless association_options
+
+      Node.value_from_hash(association_options, :class_name)
+    end
+
+    def association_name(arguments)
```
tomstuart, line 53: Great!

tomstuart (approved):
Looks good, thank you! A couple of nits for you to consider, but no blockers.

lib/packwerk/association_inspector.rb
doodzik: I might be projecting, but I think you are raising a critical point: the Node module's API is a bit messy. To your point, let's make these predicates the only type-related API that `Node` exposes. I want to do that change in another PR since it seems a bit more practical.

PR #43, opened 2020-10-22
Add Checker Interface for PrivacyChecker and DependencyChecker

doodzik:

When I was going through the codebase I noticed that PrivacyChecker and DependencyChecker conform to an interface but don't have one defined. In this PR I'm adding an interface to codify that interface. This will make it easier to add additional sorbet signatures and makes sure that the contract for the checker classes is enforced.

## Type of Change

## Checklist


lib/packwerk/dependency_checker.rb
```diff
@@ -2,9 +2,12 @@
 # frozen_string_literal: true
 
 require "packwerk/violation_type"
+require "packwerk/checker_interface"
 
 module Packwerk
   class DependencyChecker
+    require "packwerk/checker_interface"
```
tomstuart, line 9:
```suggestion
    include CheckerInterface
```

tomstuart:
Ditto [my comment on #39](https://github.com/Shopify/packwerk/pull/39#issuecomment-715157905) about commits and interface names.

I think it’s a good idea to add an interface for `PrivacyChecker` and `DependencyChecker` 🎉

exterm:
I agree - this refactor is a good idea, and we should try to align on a way to solve the problem that interface implementation looks like module inclusion and it's confusing.

I added a comment to https://github.com/Shopify/packwerk/pull/39 as well.

doodzik:
I made some adjustments to the PR and split the commits up. Please have a look again :)

lib/packwerk/checker.rb
```diff
@@ -0,0 +1,20 @@
+# typed: true
+# frozen_string_literal: true
+
+require "sorbet-runtime"
+require "packwerk/reference_lister"
+
+module Packwerk
+  module Checker
```
tomstuart, line 8: This interface is a good idea, and `Checker` looks like a good name based on my current understanding of our naming conventions 👍

tomstuart (approved):
Good to go, as long as you convince yourself that the `lib/packwerk.rb` change is right 👍

lib/packwerk.rb
```diff
@@ -34,7 +34,6 @@
 require "packwerk/parsers"
 require "packwerk/privacy_checker"
 require "packwerk/reference_extractor"
-require "packwerk/reference_lister"
```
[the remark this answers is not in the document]
doodzik: I think it makes sense, but if we decide that we don't want to do it I will change it back 👍

PR #44, opened 2020-10-22
Refactor RunContext to NodeProcessorBuilder

doodzik:

When I was reading through the code I got a bit confused about the responsibility of the `RunContext` class. The particular issue I saw was that it is used to build a NodeProcessor and a FileProcessor. The way the latter one is build is a bit backward because it uses gets initialized in with a RunContext, but used in the CLI class. So I moved the initialization of the FileProcessor out of the RunContext, which made the responsibilities a bit clearer and code easier to follow.

After I moved the FileProcessor out of the RunContext it made was easier to see that it is used to build a NodeProcessor, so I changed the name and made some changes to the class structure that make the code easier to understand.

## Type of Change

## Checklist


test/unit/configuration_test.rb
```diff
@@ -8,7 +8,7 @@ module Packwerk
   class ConfigurationTest < Minitest::Test
     test ".from_path raises ArgumentError if path does not exist" do
       File.expects(:exist?).with("foo").returns(false)
-      error = assert_raises ArgumentError do
```
doodzik, line 11: rubocop

lib/packwerk/node_processor_builder.rb
```diff
@@ -0,0 +1,92 @@
...
+    ]
+
+    const :root_path, String
+    const :reference_lister, T.nilable(ReferenceLister)
+    const :package_paths, T.nilable(String)
+    const :node_processor_class, T.untyped, default: ::Packwerk::NodeProcessor
+    const :load_paths, T.nilable(T::Array[String])
+    const :inflector, T.nilable(T.untyped)
+    const :custom_associations, T::Array[String], default: []
+    const :checker_classes, T.untyped, default: DEFAULT_CHECKERS
```
doodzik, line 33: @paracycle I couldn't find it in the docs, but do you know if we can specify a class and not an object as a type in sorbet?

lib/packwerk/cli.rb
```diff
@@ -9,19 +9,27 @@
...
+      out: $stdout,
+      err_out: $stderr,
+      style: OutputStyles::Plain
+    )
       @out = out
       @err_out = err_out
       @style = style
       @configuration = configuration || Configuration.from_path
-      @run_context = run_context || Packwerk::RunContext.from_configuration(@configuration)
+      reference_lister = ::Packwerk::CheckingDeprecatedReferences.new(@configuration.root_path)
```
doodzik, line 30: From my understanding of the `CheckingDeprecatedReferences` and `UpdatingDeprecatedReferences` class, both are concerned with producing a type of output for the user of packwerk. As such the initialization of the classes should happen outside the `NodeProcessorBuilder` and rather in the `Cli` class.

tomstuart:
[Same again](https://github.com/Shopify/packwerk/pull/42#issuecomment-715155769), sorry. Is this a whole new class or a modification of the previous one?

doodzik:
@tomstuart I split this PR into separate commits, so it should be easier to review.

lib/packwerk/cli.rb
```diff
@@ -177,6 +185,11 @@ def check(paths)
       all_offenses.empty?
     end
 
+    sig { returns(FileProcessor) }
+    def file_processor
+      @file_processor ||= FileProcessor.new(node_processor_builder: @node_processor_builder)
```
exterm, line 190: The CLI already knows way too much and I don't think it should know about `FileProcessor`.

exterm: Ideally the CLI would just be a thin layer that translates one command line call into one to three ruby method calls.

lib/packwerk/node_processor_builder.rb
exterm: I think you want [`T.class_of`](https://sorbet.org/docs/class-of)

exterm:
I agree that `RunContext` is due for a refactor, but I'm not sure this is the right one.

For context, `RunContext` is a leftover hack from when we were still running on rubocop. Rubocop does all processing per file and there is no way to carry state over from one file to the next, meaning we wouldn't be able to memoize or accumulate data across files.

We created `RunContext` and put it into a class attribute on the cop to solve that problem. This hack is no longer needed, so we may want to think about what `RunContext` is in this new world.

Intuitively it looks to me like we should just have a `Run` that we instantiate with all our configuration and options and then just `run.execute`. That could also allow us to move a lot of complexity out of `CLI`.

tomstuart:
This PR is in my todo list but I’m waiting for the outcome of @exterm‘s comment on the big picture before I dig into the detail.

doodzik:
closing in favour of https://github.com/Shopify/packwerk/pull/63

PR #45, opened 2020-10-22
ignore .byebug_history in git

doodzik:

I was looking into the code with byebug and accidentally committed the .byebug_history on another branch. Since it is pretty common to use byebug to debug code I think it makes sense to add the `.byebug_history` file to the `.gitignore` list.

## Type of Change

- [x] Configuration Change

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,5 @@
 /.bundle/
 /tmp/
 .rubocop-*
+.byebug_history
 sorbet/rbi/hidden-definitions/errors.txt


rochlefebvre (approved):
Awesome! I was sitting on the same change. Thanks!

tomstuart:
Since this is the kind of file you want to ignore in every repo, not just `packwerk`, you might want to put it in your [global `.gitignore`](https://docs.github.com/en/free-pro-team@latest/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer).

doodzik:
> Since this is the kind of file you want to ignore in every repo, not just packwerk, you might want to put it in your global .gitignore.

I think it is an interesting question to ask what belongs to a .gitignore file or not. IMO any file that was added by a dependency of a project should be ignored. In our case we are [defining .byebug as a dependency in the Gemfile](https://github.com/Shopify/packwerk/blob/main/Gemfile#L21). So I think it makes sense to add it to the project's .gitignore file rather than a global one.

exterm:
I agree with Frederik's reasoning here. If it's generated by a dependency specified in this repository, everyone using this repository will have the file so it makes sense to ignore it in the repo.

On the other hand, things like the temporary files that your editor (or any other local tooling) generates should be ignored on your machine, across repos.

PR #46, opened 2020-10-22
Update pull_request_template.md

doodzik:

I noticed in https://github.com/Shopify/packwerk/pull/45 that we didn't have a `Type of Change` defined for configuration changes in our PullRequest template. So I'm adding it to the list.

## Type of Change

- [x] Configuration change

## Checklist

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -7,16 +7,15 @@
 ## What should reviewers focus on?
 
 
-
 ## Type of Change
 
-- [ ] Bug fix (non-breaking change which fixes an issue)
-- [ ] New feature (non-breaking change which adds functionality)
-- [ ] Code refactor (non-breaking change that doesn't add functionality)
-- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
-- [ ] This change requires a documentation update
+- [ ] Bugfix
+- [ ] New feature
+- [ ] Non-breaking change (a change that doesn't alter functionality - i.e., code refactor, configs, etc.)
 
 ### Additional Release Notes
+
+- [ ] Breaking change (fix or feature that would cause existing functionality to change)
 
 Include any notes here to include in the release description. For example, if you selected "breaking change" above, leave notes on how users can transition to this version.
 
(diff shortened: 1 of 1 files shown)


rochlefebvre (approved):
Makes sense as a new category.

wildmaples:
What other configuration changes do you foresee that would fall under this category?

doodzik:
> What other configuration changes do you foresee that would fall under this category?

Mainly changes to any other configuration file in this repository. For example, `dev.yml`, `.github` folder, changes to the shipit file, etc. Any change in those categories isn't captured by any other category that we have defined at the moment.

.github/pull_request_template.md
```diff
@@ -15,6 +15,7 @@
 - [ ] Code refactor (non-breaking change that doesn't add functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
 - [ ] This change requires a documentation update
+- [ ] Configuration change
```
lastgabs, line 18: The goal of this list is for us to get a glimpse into what to write in our Release Notes.
Ultimately more important than the labels we attributed to each one of these points is the description, so maybe we should change that instead. It might be best renaming `Code refactor` to `Non-breaking change (non-breaking change that doesn't add functionality - ie. code refactor, configs, etc.)`

My reasoning is that a configuration change that changes no behavior for users of the gem should be a `non-breaking change that doesn't add functionality`. It is possible that a config change does break functionality for users, and so it should go into the `breaking change` category.

Wdyt?

doodzik: I think your reasoning makes sense here. I wonder if it makes sense to reword the list a bit if we use it to help us write our release notes 🤔 What do you think of this:

```markdown
## Type of Change

- [ ] Bugfix
- [ ] New feature

## Checklist

```

.github/pull_request_template.md
```diff
@@ -7,21 +7,22 @@
...
-- [ ] Bug fix (non-breaking change which fixes an issue)
-- [ ] New feature (non-breaking change which adds functionality)
-- [ ] Code refactor (non-breaking change that doesn't add functionality)
-- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
-- [ ] This change requires a documentation update
+- [ ] Bugfix
+- [ ] New feature
+- [ ] Non-breaking change (a change that doesn't alter functionality - i.e., code refactor, configs, etc.)
+
+## Checklist
```
lastgabs, line 16: Thanks for making the changes! Only thing I'd say is to move this section below the additional release notes section, so that it doesn't break the flow (non-breaking change and breaking change). Otherwise it's a 🚢 for me.

lastgabs (approved):
GG! Thanks Frederik!

Issue #48, opened 2020-10-23
[Bug Report] `packwerk update` requires `inflections.yml` file to be loaded from Rails


wildmaples:
**Description**
When `packwerk update` is run (without any loaded files) from the command line, Packwerk fails to interprets some constant association. 

**To Reproduce**
1. Add a custom association to the inflection file:
```
uncountable:
- 'crazy_details'
```

2. Create a mock violation in a package
```
has_many :crazy_details
```

3. Run `packwerk update` to get Packwerk to update the shitlist with a violation. You will see that Packwerk does not recognize the association. 

4. Run `bin/packwerk update` to get Packwerk to update. You will see that the violation is captured because the files are loaded through `bin/packwerk` script. 

**Expected Behaviour**

See 4 above. 

**Screenshots**

**Version Information**
 - Packwerk: [e.g. v0.1.7]
 - Ruby [e.g. v1.7]

**Additional Context**

Potential solutions: 
- Add a warning for `packwerk update` runs
- Update documentation to ensure that it is clear that files need to be loaded if there are custom associations used in the application.

tomstuart:
I don’t fully understand the problem yet, but is there no solution which is “make it work properly”? Why can’t we load the inflections ourselves?

tomstuart:
(More generally, I’m confused about `bundle exec packwerk` vs `bin/packwerk` vs `bin/spring packwerk` and don’t understand why we have all three.)

PR #53, opened 2020-10-28
Fix inflections bug that resulted in missed constant associations violations 

wildmaples:

https://github.com/Shopify/packwerk/issues/48

**TL;DR:** 

Packwerk should be reading from the `inflections.yml` file (or whatever custom association file specified) of an application directly, even when Packwerk is run without any files loaded (i.e. via `bundle exec packwerk check`). 

Recently, we found that constant associations were not read properly, which resulted in missing violations in a shitlist. However, if you run it with the application files loaded (i.e. with `bin/packwerk` or `bin/spring`) you would get the expected behaviour. 

`RunContext` was using `ActiveSupport::Inflector`, which passes it along to the `AssociationInspector`. When the files are not loaded from the application, no custom associations gets registered because `ActiveSupport::Inflector` has no knowledge of any custom inflector. 

The bug was initially missed because when files are loaded from the application, `inflections.rb` runs `ActiveSupport::Inflector.inflections` to [apply the custom inflections to Packwerk](https://github.com/Shopify/shopify/blob/98237b83a87d2726b78b5844986b9c12dc225f46/config/initializers/004_inflections.rb#L8-L11). 

We should ensure Packwerk's run behaviour is the same whether or not the application files are loaded. In order to do that, we should be using `Packwerk::Inflector` in `RunContext`, which would read from the custom inflections file. 

Changes in this PR: 

- Some improvements to `Packwerk::Inflector` making it more testable. 
  - Add a constructor that takes in a file path and creates a `Packwerk::Inflections::Custom` object
  - Add test cases ensuring custom inflections are applied
- The changes to `RunContext` to fix the bug
  - Use `Packwerk::Inflector`'s constructor, so `RunContext` doesn't need to know about `Packwerk::Inflections::Custom`
  - Add integration test case that specifically tests the constant association bug we found
- Fix the broken `ApplicationValidator` check for inflections as it was also using `ActiveSupport::Inflector` 
  - Swap `ActiveSupport::Inflector` to `Packwerk::Inflector` in application validation check 
- Remove a test case in `ApplicationValidator` which I believes is now irrelevant as it was testing the incorrect behaviour of for custom inflections 

Everything. This change looks like a huge ass PR so I broke down the changes by commits. I believe it makes sense as a single PR though. 

I'll also add comments to guide your review.

Does this change make sense? Is there another test I can add?  

## Type of Change

### Additional Release Notes

This bug fix may result in additional valid violations being found / recorded in applications with custom inflections, such as Core. 

## Checklist

### DIFF
--- a/lib/packwerk/application_validator.rb
+++ b/lib/packwerk/application_validator.rb
@@ -170,14 +170,12 @@ module Packwerk
     def check_inflection_file
       inflections_file = @configuration.inflections_file
 
-      test_inflections = ActiveSupport::Inflector::Inflections.new
-
-      Packwerk::Inflections::Default.apply_to(test_inflections)
-      Packwerk::Inflections::Custom.new(inflections_file).apply_to(test_inflections)
+      application_inflections = ActiveSupport::Inflector.inflections
+      packwerk_inflections = Packwerk::Inflector.from_file(inflections_file).inflections
 
       results = %i(plurals singulars uncountables humans acronyms).map do |type|
-        expected = ActiveSupport::Inflector.inflections.public_send(type).to_set
-        actual = test_inflections.public_send(type).to_set
+        expected = application_inflections.public_send(type).to_set
+        actual = packwerk_inflections.public_send(type).to_set
 
         if expected == actual
           Result.new(true)
--- a/lib/packwerk/association_inspector.rb
+++ b/lib/packwerk/association_inspector.rb
@@ -16,7 +16,7 @@ module Packwerk
       has_and_belongs_to_many
     ).to_set
 
-    def initialize(inflector: Inflector.new, custom_associations: Set.new)
+    def initialize(inflector:, custom_associations: Set.new)
       @inflector = inflector
       @associations = RAILS_ASSOCIATIONS + custom_associations
     end
--- a/lib/packwerk/inflector.rb
+++ b/lib/packwerk/inflector.rb
@@ -8,20 +8,31 @@ require "packwerk/inflections/custom"
 module Packwerk
   class Inflector
     class << self
+      extend T::Sig
+
       def default
         @default ||= new
       end
+
+      sig { params(inflections_file: String).returns(::Packwerk::Inflector) }
+      def from_file(inflections_file)
+        new(custom_inflector: Inflections::Custom.new(inflections_file))
+      end
     end
 
-    # For #camelize, #classify, #pluralize, #singularize
-    include ::ActiveSupport::Inflector
+    extend T::Sig
+    include ::ActiveSupport::Inflector # For #camelize, #classify, #pluralize, #singularize
 
-    def initialize(custom_inflection_file: nil)
+    sig do
+      params(
+        custom_inflector: Inflections::Custom
+      ).void
+    end
+    def initialize(custom_inflector: Inflections::Custom.new)
       @inflections = ::ActiveSupport::Inflector::Inflections.new
 
       Inflections::Default.apply_to(@inflections)
-
-      Inflections::Custom.new(custom_inflection_file).apply_to(@inflections)
+      custom_inflector.apply_to(@inflections)
     end
 
     def pluralize(word, count = nil)
--- a/lib/packwerk/inflector.rb
+++ b/lib/packwerk/inflector.rb
@@ -43,8 +43,6 @@ module Packwerk
       end
     end
 
-    private
-
     def inflections(_ = nil)
       @inflections
     end
--- a/lib/packwerk/inflector.rb
+++ b/lib/packwerk/inflector.rb
@@ -11,7 +11,7 @@ module Packwerk
       extend T::Sig
 
       def default
-        @default ||= new
+        @default ||= new(custom_inflector: Inflections::Custom.new)
       end
 
       sig { params(inflections_file: String).returns(::Packwerk::Inflector) }
@@ -28,7 +28,7 @@ module Packwerk
         custom_inflector: Inflections::Custom
       ).void
     end
-    def initialize(custom_inflector: Inflections::Custom.new)
+    def initialize(custom_inflector:)
       @inflections = ::ActiveSupport::Inflector::Inflections.new
 
       Inflections::Default.apply_to(@inflections)
--- a/test/fixtures/skeleton/different_inflections.yml
+++ b/test/fixtures/skeleton/different_inflections.yml
@@ -1,0 +1,4 @@
+acronym:
+  - 'TLA'
+  - 'WTF'
+  - 'LOL'
--- a/test/integration/custom_executable_test.rb
+++ b/test/integration/custom_executable_test.rb
@@ -35,6 +35,19 @@ module Packwerk
         assert_match(/1 offense detected/, captured_output)
       end
 
+      test "'packwerk check' with violations accounts for custom inflections, fails and displays violations" do
+        Tempfile.create(["timeline_comment", ".rb"], timeline_path("app", "models")) do |file|
+          # payment_details is an uncountable inflection
+          file.write("class TimelineComment; has_one :payment_details; end")
+          file.flush
+
+          refute_successful_run("check")
+        end
+
+        assert_match(/Privacy violation: '::PaymentDetails'/, captured_output)
+        assert_match(/1 offense detected/, captured_output)
+      end
+
       test "'packwerk update-deprecations' with no violations succeeds and updates no files" do
         deprecated_reference_content = read_deprecated_references
 
--- a/test/unit/application_validator_test.rb
+++ b/test/unit/application_validator_test.rb
@@ -42,7 +42,7 @@ module Packwerk
     test "returns error for mismatched inflections.yml file" do
       config_path = "test/fixtures/skeleton/packwerk.yml"
       configs = YAML.load_file(config_path)
-      configs["inflections_file"] = "custom_inflections.yml"
+      configs["inflections_file"] = "different_inflections.yml"
 
       configuration = Packwerk::Configuration.new(configs, config_path: config_path)
 
--- a/test/unit/application_validator_test.rb
+++ b/test/unit/application_validator_test.rb
@@ -9,6 +9,9 @@ require "packwerk/application_validator"
 # make sure PrivateThing.constantize succeeds to pass the privacy validity check
 require "fixtures/skeleton/components/timeline/app/models/private_thing.rb"
 
+# make sure the application has a chance to load its inflections
+require "fixtures/skeleton/config/environment"
+
 module Packwerk
   class ApplicationValidatorTest < Minitest::Test
     setup do
--- a/test/unit/cli_test.rb
+++ b/test/unit/cli_test.rb
@@ -117,7 +117,8 @@ module Packwerk
         root_path: @temp_dir,
         load_paths: ["path"],
         package_paths: "**/",
-        custom_associations: ["cached_belongs_to"]
+        custom_associations: ["cached_belongs_to"],
+        inflections_file: "config/inflections.yml"
       )
       cli = ::Packwerk::Cli.new(configuration: configuration, out: string_io)
 
@@ -134,7 +135,8 @@ module Packwerk
         root_path: @temp_dir,
         load_paths: ["path"],
         package_paths: "**/",
-        custom_associations: ["cached_belongs_to"]
+        custom_associations: ["cached_belongs_to"],
+        inflections_file: "config/inflections.yml"
       )
       cli = ::Packwerk::Cli.new(configuration: configuration, out: string_io)
 
--- a/test/unit/inflector_test.rb
+++ b/test/unit/inflector_test.rb
@@ -8,7 +8,7 @@ require "packwerk/inflector"
 module Packwerk
   class InflectorTest < Minitest::Test
     def setup
-      @inflector = Inflector.new
+      @inflector = inflector_for(file: "config/inflections.yml")
     end
 
     test "acts like activesupport inflector" do
@@ -37,6 +37,27 @@ module Packwerk
     test "#pluralize will singularize when count is 1" do
       assert_equal "thing", @inflector.pluralize("thing", 1)
       assert_equal "thing", @inflector.pluralize("things", 1)
+    end
+
+    test "#initialize will apply custom inflections from file" do
+      inflector = inflector_for(file: "test/fixtures/skeleton/custom_inflections.yml")
+
+      assert_equal "graphql", inflector.underscore("GraphQL")
+      assert_equal "payment_details", inflector.singularize("payment_details")
+    end
+
+    test "#initialize will not apply custom inflections if there aren't any" do
+      inflector = inflector_for(file: "no_inflections_here.yml")
+
+      assert_equal "graph_ql", inflector.underscore("GraphQL")
+      assert_equal "payment_detail", inflector.singularize("payment_details")
+    end
+
+    private
+
+    def inflector_for(file:)
+      custom_inflector = Packwerk::Inflections::Custom.new(file)
+      Inflector.new(custom_inflector: custom_inflector)
     end
   end
 end
--- a/lib/packwerk/run_context.rb
+++ b/lib/packwerk/run_context.rb
@@ -1,7 +1,6 @@
 # typed: true
 # frozen_string_literal: true
 
-require "active_support/inflector"
 require "constant_resolver"
 
 require "packwerk/association_inspector"
@@ -10,6 +9,7 @@ require "packwerk/constant_discovery"
 require "packwerk/const_node_inspector"
 require "packwerk/dependency_checker"
 require "packwerk/file_processor"
+require "packwerk/inflector"
 require "packwerk/node_processor"
 require "packwerk/package_set"
 require "packwerk/privacy_checker"
@@ -36,11 +36,13 @@ module Packwerk
       def from_configuration(configuration, reference_lister: nil)
         default_reference_lister = reference_lister ||
           ::Packwerk::CheckingDeprecatedReferences.new(configuration.root_path)
+        inflector = ::Packwerk::Inflector.from_file(configuration.inflections_file)
+
         new(
           root_path: configuration.root_path,
           load_paths: configuration.load_paths,
           package_paths: configuration.package_paths,
-          inflector: ActiveSupport::Inflector,
+          inflector: inflector,
           custom_associations: configuration.custom_associations,
           reference_lister: default_reference_lister,
         )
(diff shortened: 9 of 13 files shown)


lib/packwerk/inflector.rb
```diff
@@ -32,8 +49,6 @@ def pluralize(word, count = nil)
       end
     end
 
-    private
```
wildmaples, line 35: Removed this as a private method because the `ApplicationValidator` needs to [check the inflections ](https://github.com/Shopify/packwerk/pull/53/files#diff-5ca9dcd54fd0b9eeef94e00ba6adc7822dc83e5dd2bfb9e8d96fd3c356de8b8aR179)of this class.

lib/packwerk/inflector.rb
```diff
@@ -8,20 +8,37 @@
 module Packwerk
   class Inflector
     class << self
+      extend T::Sig
+
       def default
```
wildmaples, line 13: FYI - The instance of `Packwerk::Inflector` returned by `default` is used internally to inflect words output by Packwerk (e.g. "5 violations found"). It is a different use case and irrelevant to this bug.

test/integration/custom_executable_test.rb
```diff
@@ -35,6 +35,19 @@ class CustomExecutableTest < Minitest::Test
         assert_match(/1 offense detected/, captured_output)
       end
 
+      test "'packwerk check' with violations accounts for custom inflections, fails and displays violations" do
+        Tempfile.create(["timeline_comment", ".rb"], timeline_path("app", "models")) do |file|
+          # payment_details is an uncountable inflection
```
wildmaples, line 40: I added this comment because it may not be clear to someone who is just reading the code. You would have to know that `payment_details` in `custom_inflections.yml` is an "uncountable", which means it should not be inflected when found in a constant association. 

Without the fix for this bug, this `check` run would pass because Packwerk is unable to resolve the constant and treats it as a non-violation. In reality, `payment_details` exists as a constant, which makes it a legitimate violation.

test/unit/application_validator_test.rb
```diff
@@ -67,7 +49,6 @@ class ApplicationValidatorTest < Minitest::Test
 
       config_path = "test/fixtures/skeleton/packwerk.yml"
       configs = YAML.load_file(config_path)
-      configs["inflections_file"] = "custom_inflections.yml"
```
wildmaples, line 70: We don't need this line anymore as it's [specified directly](https://github.com/Shopify/packwerk/pull/53/files#diff-3f56f44fb50daf4959f7aa276e090b22307338992ad49247e9540b96151e456eR12) in the file.

test/unit/application_validator_test.rb
```diff
@@ -39,24 +39,6 @@ class ApplicationValidatorTest < Minitest::Test
       refute result.ok?, result.error_value
     end
 
-    test "returns error for mismatched inflections.yml file" do
```
wildmaples, line 42: This test case seems to be validating the bug.

Previously `custom_inflections.yml` was not specified in the skeleton configuration file that is used for our integration tests. In this test, the file was added to the configuration and expected to fail. This was because the validator was using the `ActiveSupport::Inflector` and not checking for any custom inflections.

test/unit/cli_test.rb
```diff
@@ -117,7 +117,8 @@ class CliTest < Minitest::Test
         root_path: @temp_dir,
         load_paths: ["path"],
         package_paths: "**/",
-        custom_associations: ["cached_belongs_to"]
+        custom_associations: ["cached_belongs_to"],
+        inflections_file: "config/inflections.yml"
```
wildmaples, line 121: Along with [the lines below](https://github.com/Shopify/packwerk/pull/53/files#diff-efc360b012815c4277ebe4bf8fdde379c5596bacc8dbc6cc139419abeac71c54R139), this key is now required in the `RunContext` because we now gotta know the inflection file path.

test/unit/inflector_test.rb
```diff
@@ -38,5 +38,33 @@ def setup
...
+      file = "./test/fixtures/skeleton/custom_inflections.yml"
+      custom_inflector = custom_inflector_for(file)
+
+      expected_inflections = {
+        "acronym" => ["GraphQL", "MRuby", "TOS"],
+        "irregular" => [["analysis", "analyses"], ["reserve", "reserves"]],
+        "uncountable" => ["payment_details"],
+      }
+
+      assert_equal custom_inflector.inflections, expected_inflections
```
tomstuart, line 52: I think this is testing the wrong thing. `#custom_inflector_for` goes to the trouble of creating a new `Packwerk::Inflector` and applying the custom inflections to it, but then the inflector gets thrown away and we’re just looking at the `@inflections` instance variable on the `Packwerk::Inflections::Custom` object to see what it contains. So we’re not testing whether those inflections have been applied to anything, only that they were successfully read from the file. We should either change the test description to reflect what it’s really doing or rewrite the test so that `#custom_inflector_for` actually returns the inflector and we try it out by seeing what e.g. `custom_inflector.underscore('GraphQL')` and `custom_inflector.singularize('payment_details')` return.

test/integration/custom_executable_test.rb
tomstuart: This is good information which deserves a promotion into the commit message 😄

lib/packwerk/application_validator.rb
```diff
@@ -176,7 +176,7 @@ def check_inflection_file
       Packwerk::Inflections::Custom.new(inflections_file).apply_to(test_inflections)
 
       results = %i(plurals singulars uncountables humans acronyms).map do |type|
-        expected = ActiveSupport::Inflector.inflections.public_send(type).to_set
+        expected = Packwerk::Inflector.from_file(inflections_file).inflections.public_send(type).to_set
```
tomstuart, line 179: What is `Packwerk::ApplicationValidator#check_inflection_file` supposed to be checking? I don’t understand under what circumstances this check is intended to fail, which means I don’t know enough to say whether this change is right.

tomstuart (commented):
Thank you Maple! I need more time to actually run this code and confirm that it works as expected, but I’m submitting my review comments so far so that you get them ASAP.

test/unit/application_validator_test.rb
tomstuart: Again, an excellent piece of documentation that deserves to live in the commit message.

lib/packwerk/inflector.rb
wildmaples: I am not quite sure why we used `Packwerk::Inflector` for general UI inflections either. To add on to that, using `Packwerk::Inflector` in two different use cases also confused me when I was digging around. We should probably just give it one concern. 

I will create another PR for this after.

exterm: Might be worth digging for the PR that introduced `Packwerk:Inflector.default` to see if a convincing reason was given.

lib/packwerk/inflector.rb
```diff
@@ -8,20 +8,37 @@
...
 
-    def initialize(custom_inflection_file: nil)
+    sig do
+      params(
+        custom_inflector: Inflections::Custom
+      ).void
+    end
+    def initialize(custom_inflector: Inflections::Custom.new)
+      @custom_inflector = custom_inflector
+      @default_inflector = Inflections::Default
```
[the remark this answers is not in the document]
exterm: `inflections`? That's also not great but it follows the class naming.

lib/packwerk/inflector.rb
```diff
@@ -8,20 +8,37 @@
...
+    extend T::Sig
+    include ::ActiveSupport::Inflector # For #camelize, #classify, #pluralize, #singularize
 
-    def initialize(custom_inflection_file: nil)
+    sig do
+      params(
+        custom_inflector: Inflections::Custom
+      ).void
+    end
+    def initialize(custom_inflector: Inflections::Custom.new)
```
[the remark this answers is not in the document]
wildmaples: Previously we were setting up the [default behaviour of `Packwerk::Inflector` to be nil](https://github.com/Shopify/packwerk/pull/53/files#diff-444596239abec8bf66645e487b66687f3942e9b28fab77a7a03483870d7fdb30L19), which could (in part) have led to the development of this bug. So I think you are right, I will change this to be a required parameter.

lib/packwerk/inflector.rb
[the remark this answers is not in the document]
wildmaples: How do you feel about being more verbose and name spacing it `Packwerk::Inflections::Helpers::Custom` or alternatively `Packwerk::Inflections::CustomHelper`?

lib/packwerk/inflector.rb
wildmaples: It seems like `Packwerk:Inflector.default` was used for `Packwerk:Inflector#pluralize`. `ActiveSupport::Inflector`'s [version of pluralize does not take in an argument](https://github.com/Shopify/packwerk-old/pull/150#discussion_r422123662), so we created the pluralization logic in `Packwerk:Inflector#pluralize`.

lib/packwerk/application_validator.rb
wildmaples: Perhaps @exterm can give a better response since you wrote it? I am not quite sure either. From what I understand, we've previously used `ActiveSupport::Inflector` as the inflector, and now just switching over to the Packwerk's version.

lib/packwerk/inflector.rb
```diff
@@ -8,20 +8,37 @@
...
+      @custom_inflector = custom_inflector
+      @default_inflector = Inflections::Default
       @inflections = ::ActiveSupport::Inflector::Inflections.new
 
-      Inflections::Default.apply_to(@inflections)
+      apply_all_inflections
+    end
 
-      Inflections::Custom.new(custom_inflection_file).apply_to(@inflections)
+    def apply_all_inflections
```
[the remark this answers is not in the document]
wildmaples: Fixed in: [`cda9947` (#53)](https://github.com/Shopify/packwerk/pull/53/commits/cda994752819ae919689704b08351f8e5ddd72e7)

test/unit/inflector_test.rb
wildmaples: Fixed in: [`cda9947` (#53)](https://github.com/Shopify/packwerk/pull/53/commits/cda994752819ae919689704b08351f8e5ddd72e7)

lib/packwerk/inflector.rb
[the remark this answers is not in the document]
wildmaples: Fixed in: [`9165078` (#53)](https://github.com/Shopify/packwerk/pull/53/commits/9165078bb04ce409349d0aee2192506f70fe11f4)

lib/packwerk/inflector.rb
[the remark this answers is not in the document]
exterm: I think the class names are good, but the variable names could be improved. `@default_inflector` should probably be `@default_inflections`.

lib/packwerk/application_validator.rb
exterm: Ah, this change is wrong - sorry for not writing the code in a more obvious way!

The purpose of this check is to compare the inflections that are configured for packwerk with the actual inflections present in the application. So this line really should not be changed.

exterm: Wait no I am confused

exterm: Yeah, this whole check is super confusing.

But suffice it to say, my explanation is correct, the previous version was correct, this change breaks it.

exterm: Further up we have 

```
      test_inflections = ActiveSupport::Inflector::Inflections.new
      Packwerk::Inflections::Default.apply_to(test_inflections)
      Packwerk::Inflections::Custom.new(inflections_file).apply_to(test_inflections)
```

and then down here

```
expected = ActiveSupport::Inflector.inflections.public_send(type).to_set
actual = test_inflections.public_send(type).to_set
```

So `expected` is the application's inflections and `actual` is packwerk's inflections.

exterm: It can be simplified to (pseudocode)

```
expected = application_inflections = ActiveSupport::Inflector.inflections.public_send(type).to_set
actual = packwerk_inflections = Packwerk::Inflector.from_file(configuration.inflections_file).inflections.public_send(type).to_set
```

exterm (changes requested):
Breaks the validator for inflections

lib/packwerk/inflector.rb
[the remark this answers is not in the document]
wildmaples: I removed the parameters name completely - making it similar to what we had in the init method before.

lib/packwerk/application_validator.rb
```diff
@@ -170,14 +170,15 @@ def check_application_structure
     def check_inflection_file
       inflections_file = @configuration.inflections_file
 
-      test_inflections = ActiveSupport::Inflector::Inflections.new
+      application_inflections = ActiveSupport::Inflector::Inflections.new
+      Packwerk::Inflections::Default.apply_to(application_inflections)
+      Packwerk::Inflections::Custom.new(inflections_file).apply_to(application_inflections)
```
exterm, line 175: I'm pretty sure this is wrong.

The validator is running in the application context, so `application_inflections = ActiveSupport::Inflector.inflections`.

tomstuart: Thanks @exterm. Since Maple’s away this week I’m going to try to resolve this on her behalf.

tomstuart: Okay, I agree — this is the crux of my [confusion](https://github.com/Shopify/packwerk/pull/53#discussion_r514148499) over this bit of code (“I don’t understand under what circumstances this check is intended to fail”).

I think we’re trying to confirm that the application has [correctly wired up](https://github.com/Shopify/packwerk/blob/main/USAGE.md#inflections) its Packwerk inflections in `config/initializers/inflections.rb` or wherever. But if we ourselves apply the Packwerk inflections on top of whatever the application has already loaded, this check will always succeed regardless of whether the application is wired up correctly.

As far as I can tell we have no unit or acceptance tests covering the case where `#check_inflection_file` fails, so it’s unsurprising that it doesn’t work. It’s a separate bug than the one being fixed by this PR, so I’ll open another PR and rebase this one on top of it.

tomstuart: Oh, surprise, I was confused again. The code on `main` does the right thing, it’s just this change that’s wrong. Sorry. I shall fix this PR then.

lib/packwerk/inflector.rb
```diff
@@ -28,17 +28,11 @@ def from_file(inflections_file)
         custom_inflector: Inflections::Custom
       ).void
     end
-    def initialize(custom_inflector: Inflections::Custom.new)
-      @custom_inflector = custom_inflector
-      @default_inflector = Inflections::Default
+    def initialize(custom_inflector:)
```
tomstuart, line 31: This change to the `custom_inflector` parameter is in the wrong commit. It belongs in “Make custom_inflector argument in Inflector required”.

tomstuart:
I have fixed the above problems and rebased this branch to be up to date with `main`. Tidying aside, the main change I made was to add a `config/environment.rb` to the `test/fixtures/skeleton` app so that there was a way for the tests to “boot” it in order for the application validator (specifically `#check_inflection_file`) to do its job.

@exterm, how does this look now?

cjhutchi (approved):
LGTM! I'm not 💯 on the `packwerk` code base and how it's all wired together. But I'm not seeing any glaring errors that standout. Would recommend a second set of eyes from someone possibly more in the know.

Thanks @wildmaples and @tomstuart for clearing this up so quickly! ❤️

lib/packwerk/application_validator.rb
```diff
@@ -170,14 +170,12 @@ def check_application_structure
     def check_inflection_file
       inflections_file = @configuration.inflections_file
 
-      test_inflections = ActiveSupport::Inflector::Inflections.new
-
-      Packwerk::Inflections::Default.apply_to(test_inflections)
-      Packwerk::Inflections::Custom.new(inflections_file).apply_to(test_inflections)
+      application_inflections = ActiveSupport::Inflector.inflections
+      packwerk_inflections = Packwerk::Inflector.from_file(inflections_file).inflections
```
exterm, line 174: Oh my, all the discussions were worth it. This is so much better than what's currently on `main`

lib/packwerk/association_inspector.rb
```diff
@@ -16,7 +16,10 @@ class AssociationInspector
       has_and_belongs_to_many
     ).to_set
 
-    def initialize(inflector: Inflector.new, custom_associations: Set.new)
+    def initialize(
+      inflector: Inflector.new(custom_inflector: Inflections::Custom.new),
```
exterm, line 20: Why would `association_inspector` know about the specifics of `Inflector`, namely the existence of `Inflections::Custom`?

exterm: Maybe this should just be the default argument in `Inflector`'s initializer?

exterm: Mmmh, actually I think there may be a deeper problem here - `Inflections::Custom` doesn't do anything unless fed with an inflections file.

Does it make sense to use empty custom inflections as a default?

test/fixtures/skeleton/different_inflections.yml
```diff
@@ -0,0 +1,4 @@
+acronym:
+  - 'TLA'
+  - 'WTF'
+  - 'LOL'
```
exterm, line 4: nice

lib/packwerk/association_inspector.rb
tomstuart: This was @wildmaples pushing the required argument change to `Packwerk::Inflector#initialize` up to its call sites, i.e. inlining the previous default wherever it was already implicitly used, so there shouldn’t be any change in behaviour here.

I agree that `Inflections::Custom.new` without an inflections file is basically pointless (hence the [suggestion](https://github.com/Shopify/packwerk/pull/53#discussion_r514128192) to remove the default) so I’ll clean this up too. As far as I can tell `AssociationInspector` is only constructed by `RunContext` which [supplies an inflector](https://github.com/Shopify/packwerk/blob/main/lib/packwerk/run_context.rb#L81) anyway so this default is never used in practice.

tomstuart: I’ve pushed a revised commit which removes the `inflector` default too. I think “why are we using `Inflections::Custom` without an inflections file” is an interesting design question which we should revisit but I’m choosing not to address it here.

test/fixtures/skeleton/different_inflections.yml
cjhutchi: nice

wildmaples: nice

exterm (approved):

tomstuart:
I tested this on Shopify’s monolith and everything seems to work. I’ll merge. Thank you everyone, especially @wildmaples for doing all the hard work! 🎉 🤝

PR #54, opened 2020-10-29
Add to node parsing capabilities

shageman:
In the codebase I am working on I see several errors related to edge cases where the code parsing fails. I am working to make packwerk run.

I added a bunch of defensive statements that makes packwerk continue to run even when individual nodes fail to parse.

This doesn't seem like a cohesive PR at the moment. Although the tests do pass, how should I break these changes up so merging would be considered?

## Type of Change

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [X] It is safe to simply rollback this change.

### DIFF
--- a/lib/packwerk/node.rb
+++ b/lib/packwerk/node.rb
@@ -27,7 +27,7 @@ module Packwerk
         case type_of(constant_node)
         when CONSTANT_ROOT_NAMESPACE
           ""
-        when CONSTANT, CONSTANT_ASSIGNMENT
+        when CONSTANT, CONSTANT_ASSIGNMENT, SELF
           # (const nil :Foo)
           #   "Foo"
           # (const (cbase) :Foo)
@@ -40,6 +40,8 @@ module Packwerk
           #   "::Foo = 1"
           # (casgn (lvar :a) :Foo (int 1))
           #   "a::Foo = 1"
+          # (casgn (self) :Foo (int 1))
+          #   "self::Foo = 1"
           namespace, name = constant_node.children
           if namespace
             [constant_name(namespace), name].join("::")
@@ -204,12 +206,14 @@ module Packwerk
       HASH_PAIR = :pair
       METHOD_CALL = :send
       MODULE = :module
+      SELF = :self
       STRING = :str
       SYMBOL = :sym
 
+
       private_constant(
         :BLOCK, :CLASS, :CONSTANT, :CONSTANT_ASSIGNMENT, :CONSTANT_ROOT_NAMESPACE, :HASH, :HASH_PAIR, :METHOD_CALL,
-        :MODULE, :STRING, :SYMBOL,
+        :MODULE, :SELF, :STRING, :SYMBOL,
       )
 
       def type_of(node)
--- a/lib/packwerk/node.rb
+++ b/lib/packwerk/node.rb
@@ -210,7 +210,6 @@ module Packwerk
       STRING = :str
       SYMBOL = :sym
 
-
       private_constant(
         :BLOCK, :CLASS, :CONSTANT, :CONSTANT_ASSIGNMENT, :CONSTANT_ROOT_NAMESPACE, :HASH, :HASH_PAIR, :METHOD_CALL,
         :MODULE, :SELF, :STRING, :SYMBOL,
--- a/lib/packwerk/parsed_constant_definitions.rb
+++ b/lib/packwerk/parsed_constant_definitions.rb
@@ -29,7 +29,8 @@ module Packwerk
       fully_qualified_constant_name = "::#{constant_name}"
 
       possible_namespaces = namespace_path.reduce([""]) do |acc, current|
-        acc << acc.last + "::" + current
+        acc << "#{acc.last}::#{current}" if acc.last && current
+        acc
       end
 
       possible_namespaces.map { |namespace| namespace + fully_qualified_constant_name }
--- a/lib/packwerk/parsed_constant_definitions.rb
+++ b/lib/packwerk/parsed_constant_definitions.rb
@@ -28,9 +28,8 @@ module Packwerk
 
       fully_qualified_constant_name = "::#{constant_name}"
 
-      possible_namespaces = namespace_path.reduce([""]) do |acc, current|
+      possible_namespaces = namespace_path.each_with_object([""]) do |current, acc|
         acc << "#{acc.last}::#{current}" if acc.last && current
-        acc
       end
 
       possible_namespaces.map { |namespace| namespace + fully_qualified_constant_name }
--- a/test/unit/reference_extractor_test.rb
+++ b/test/unit/reference_extractor_test.rb
@@ -50,6 +50,15 @@ module Packwerk
       assert_equal "::Order::Extension", reference.constant.name
     end
 
+    test "doesn't break on unusal constants it can't detect" do
+      references = process(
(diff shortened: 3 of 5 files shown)


shageman:
How do I complete the CLA step?

wildmaples:
> How do I complete the CLA step?

@shageman https://cla.shopify.com/

lib/packwerk/parsed_constant_definitions.rb
```diff
@@ -38,16 +38,20 @@ def self.reference_qualifications(constant_name, namespace_path:)
...
 
     def collect_local_definitions_from_root(node, current_namespace_path = [])
-      if Node.type(node) == Node::CONSTANT_ASSIGNMENT
-        add_definition(Node.constant_name(node), current_namespace_path, Node.name_location(node))
-      elsif Node.module_name_from_definition(node)
-        # handle compact constant nesting (e.g. "module Sales::Order")
-        tempnode = node
-        while (tempnode = Node.each_child(tempnode).find { |n| Node.type(n) == Node::CONSTANT })
-          add_definition(Node.constant_name(tempnode), current_namespace_path, Node.name_location(tempnode))
+      begin
```
doodzik, line 41: From what I can tell, you are only adding a rescue block here, correct?
In that case, I would suggest getting rid of the `begin` statement in this method. 
This is a good example of what I mean: http://seejohncode.com/2012/04/17/short-tip-rescuing-a-method-in-ruby/

lib/packwerk/node.rb
```diff
@@ -29,7 +29,11 @@ def class_or_module_name(class_or_module_node)
           # (module (const nil :Foo) (nil))
           #   "module Foo; end"
           identifier = class_or_module_node.children[0]
-          constant_name(identifier)
+          begin
+            constant_name(identifier)
+          rescue Node::TypeError
+            nil
+          end
```
doodzik, line 36: I don't think we want to rescue against a TypeError in this method, because we expect it to raise a TypeError as seen on line 38. I would suggest rescuing the type of error in the method that calls this method.

doodzik (commented):
Your changes look like a great addition 👍 I left two small comments but other than that it looks good.

lib/packwerk/parsed_constant_definitions.rb
shageman: Yes!

lib/packwerk/node.rb
shageman: Makes sense!

shageman:
@doodzik I believe I addressed your comments with the new changes and hope to have also appeased the linter.

shageman:
Scratch that, my changes were too aggressive that broke that I need to go back and run tests locally. I hope to have another version tomorrow

lib/packwerk/const_node_inspector.rb
```diff
@@ -24,12 +24,10 @@ def constant_name_from_node(node, ancestors:)
...
       else
-        begin
-          Node.constant_name(node)
-        rescue Node::TypeError
-          nil
-        end
+        Node.constant_name(node)
       end
+    rescue Node::TypeError
+      nil
```
doodzik, line 30: This PR looks good to me, but I would like to get the approval of @tomstuart because he raised an issue [about defensive node parsing on one of my open PRs](https://github.com/Shopify/packwerk/pull/39/files#r513344732). From my understanding, this gem's design focuses on doing a good enough job uncovering dependency/privacy violations. In my opinion, defensive node parsing would fall under that because we don't fail if packwerk cannot handle the code. The downside is that we don't learn about parsing issues.

shageman: I totally agree. I would love to rather propose a fix for the parsing errors we are running into, but am not deep enough into packwerk and parser to understand much of what's going on. 

I would be happy to pair on this or run non-defensive versions against the codebase I am looking at to give more input into what might be going on...

shageman:
If I fix the linting error as proposed by the check, all tests fail: https://github.com/Shopify/packwerk/pull/54/checks?check_run_id=1328856280

exterm:
> all tests fail

Could you link to the build with the failing tests? the one you link to only has the failing linter.

shageman:
> > all tests fail
> 
> Could you link to the build with the failing tests? the one you link to only has the failing linter.

Yes - me too 🤔

shageman:
@exterm @doodzik I took another stab at the changes needed to make it work for our codebase... this time a bit more focussed and with tests.

lib/packwerk/parsed_constant_definitions.rb
```diff
@@ -51,6 +51,8 @@ def collect_local_definitions_from_root(node, current_namespace_path = [])
       end
 
       Node.each_child(node) { |child| collect_local_definitions_from_root(child, current_namespace_path) }
+    rescue Node::TypeError
+      nil
```
exterm, line 55: I'm a bit concerned about this part. It looks like in the future, this could lead to parsing bugs being silently ignored, and the behavior of packwerk regressing.

From your tests, it looks like the case that you want to fix is dynamic namespacing (`self::SOME_CONSTANT`). The fact that packwerk chokes on that seems like a bug to me. Can we find a solution that fixes this exact bug, instead of introducing a clause that ignores all unknown nodes?

shageman: I have removed the offending lines you point to and added the ability for packwerk to parse the `self::` syntax

lib/packwerk/node.rb
```diff
@@ -192,7 +195,7 @@ def parent_module_name(ancestors:)
       end
 
       def type(node)
-        node.type
+        node&.type
```
exterm, line 198: Do you still need this now that you are handling `self::` explicitly? The only test that fails when I revert this change is the test that explicitly tests this change.

test/unit/reference_extractor_test.rb
```diff
@@ -50,6 +50,15 @@ def setup
       assert_equal "::Order::Extension", reference.constant.name
     end
 
+    test "doesn't break on unusal constants it can't detect" do
```
exterm, line 53: I guess you can make this test description more specific now, something about namespacing by `self`?

exterm:
Thank you for the adjustments! This PR is starting to look like a very valuable bugfix.

doodzik (approved):
The PR is looking good! Once Philips comment is addressed and fix the merge conflict we can ship this PR 🚢

lib/packwerk/node.rb
shageman: ok!

test/unit/reference_extractor_test.rb
shageman: good catch! Thanks!

shageman:
Thanks, @doodzik - I have addressed all open comments

exterm:
looking good! I'm going to test this on our core monolith.

exterm:
It's complicated by the fact that packwerk main is currently broken on our monolith. We may want to fix that first before merging this. Sorry for the delay Stephan.

exterm:
We fixed the main branch. @shageman , can you rebase? I'll manually test your branch on our monolith afterwards (I would have rebased myself, but there are merge conflicts and I sadly don't have the time to look into them right now)

shageman:
@exterm done. Thanks!

exterm:
Ran it on the core monolith without problems.

exterm (approved):
Thank you for your contribution Stephan!

PR #40, opened 2020-10-22
Add support for minitest-focus

rochlefebvre:
## What are you trying to accomplish?
I would like to be able to use [minitest-focus](https://github.com/seattlerb/minitest-focus) as I develop tests and debug issues.  I can mark a few tests of interest with the `focus` method and then run `dev test`.

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Code refactor (non-breaking change that doesn't add functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist

- [x] It is safe to simply rollback this change.

## 🎩 
Call the `focus` method on a line that preceeds a test you wish to run, e.g.
```ruby
# test/unit/application_validator_test.rb
    focus
    test "validity" do
    # ...
    end

    test "returns an error for unresolvable privatized constants" do
    # ...
    end
```
Only that test will run when `rake test` or `dev test` is run:
```
➜  packwerk git:(add-minitest-focus-gem) dev test
👩‍💻  Running if [[ "$*" =~ ":"[0-9]+ ]];... from dev.yml
[ ... ]
# Running:

.

Finished in 0.030783s, 32.4855 runs/s, 32.4855 assertions/s.
1 runs, 1 assertions, 0 failures, 0 errors, 0 skips
➜  packwerk git:(add-minitest-focus-gem) ✗
```

PR #59, opened 2020-11-02
[POC] A FixureHelper module for managing test artifacts between tests

rochlefebvre:
### TL;DR
This proof-of-concept branch showcases a way to set up Packwerk tests by copying a template folder into a temporary directory, modifying those configuration fixtures, running the test, and then cleaning up the temporary artifacts.  I welcome feedback on this proposed pattern and how it's implemented.

### Current test environment
It was mentioned in #17 and https://github.com/Shopify/packwerk-old/pull/262#discussion_r455956863 that it's getting increasingly harder to write unit tests for Packwerk while using the lone https://github.com/Shopify/packwerk/tree/main/test/fixtures/skeleton test app fixture. Integrators of Packwerk configure the library via a series of yaml config files rooted in the application's own directory.  Packwerk relies on a few sources to find these configuration artifacts:
- the current working directory;
- the Rails.application `Engine` railties;
- the `Rails.application` singleton's `config` object;
- the `Rails.root` variable.

Today, many tests point to test/fixtures/skeleton as the application directory.  Some tests happen to rely on the Packwerk project's own directory.  The https://github.com/Shopify/packwerk/blob/main/test/rails_test_helper.rb sets up the Rails application singleton for all tests to work within.

### Problems
- **Coupled tests**: Changing the makeup of the `fixures/skeleton` fixture in order to support a new test has the potential of affecting existing tests: load paths, inflections, dependencies, etc.  Applying a mock on top of the shared fixture can work, but pervasive mocking has its own drawbacks.
- **Rails singleton**: As soon as a test requires `rails_test_helper`, the test process will now have `Dummy` as the lone Rails application, complete with its hard-coded load paths and root path.  Spoiler alert: I worked around this, but didn't actually solve anything.
- **Complex "skeleton" fixture**: I believe the `skeleton` fixture started off being quite lean, representing a simple yet valid Packwerk-enabled application.  As Packwerk devs implement more features and tests, the shared test app is growing as well.  It's becoming harder to know exactly how this simple application behaves.

I reasoned that Packwerk's three main entry points are 1) the application folder;  2) the `Rails.application` instance; 3) the `Rails.root` variable; we should use those three interfaces in our test setup.  I implemented the FixtureHelper module, which clones a fixture folder into a temporary `tmp/fixtures/the_test_name` directory and reconfigures the lone Rails application instance around each test execution.  Tests are free to manipulate their copy of the fixture file, and even play with the Rails load and root paths.

~~The setup and teardown occur automatically by way of an `around` method, from https://github.com/splattael/minitest-around.  Test suites are free to implement their own `setup` and `teardown` hooks without having to worry about managing the fixure directory.~~

@doodzik convinced me to eschew minitest-around in favour of a more traditional, explicit setup/teardown approach.  As such, I updated the FixtureHelper and RailsFixtureHelper modules to extract the `around` method into two methods, `setup_fixture` and `teardown_fixture`. Suites are responsible for calling those int their own setup/teardown methods.

I entertained the idea of stubbing `Rail.application` to respond with a minimally viable config mock.  This works for load paths, but does not address `Rails.root`.  Unless someone can find a good way to manage that last one, I'll just stick with caching & restoring the path values around each test.

8949c4452d08b4e874a81ba48f969b5ac52805b0 updates existing tests to use the fixture helper.  3583992ad0a31d10f88fb9b99d150e4ff3c46ba9 then adds new tests to ApplicationValidator.  Consider looking at each commit separately to follow the changes more easily.

Is the pattern easy to understand? Is it safe and correct to use (especially the Rails stuff)? Are the helper methods intuitive, or surprising?

I would also like some feedback on the contents of the new `:minimal` fixture.  It's supposed to be a simple, valid, non-trivial implementation of a Packwerk-enabled project.  I copied the `components/sales` portion of `:skeleton`, but I doubt we need all of the ruby classes, erb, test file, etc.  The goal is to seldom change this minimal fixture going forward, and instead perform extra setup in each test that requires changes.

## Type of Change

## Checklist

- ~~[] It is safe to simply rollback this change.~~ I will not ship this.


Gemfile
```diff
@@ -19,4 +19,6 @@ gem("tapioca", require: false)
 
 group :development do
   gem("byebug", require: false)
+  gem("minitest-focus", require: false)
```
rochlefebvre, line 22: The minitest-focus gem is being introduced in https://github.com/Shopify/packwerk/pull/40

lib/packwerk/application_validator.rb
```diff
@@ -24,15 +24,15 @@ def initialize(config_file_path:, application_load_paths:, configuration:)
 
     def check_all
       results = [
+        check_package_manifest_syntax,
         check_autoload_path_cache,
         check_package_manifests_for_privacy,
-        check_package_manifest_syntax,
```
rochlefebvre, line 29: This feels like the first thing we should check.

lib/packwerk/application_validator.rb
```diff
@@ -209,7 +209,7 @@ def check_inflection_file
     end
 
     def check_acyclic_graph
-      packages = Packwerk::PackageSet.load_all_from(".")
```
rochlefebvre, line 212: `"."` resolves to the packwerk project's own directory, not the directory of the application under test.

lib/packwerk/package_set.rb
```diff
@@ -27,9 +27,11 @@ def load_all_from(root_path, package_pathspec: nil)
 
       def package_paths(root_path, package_pathspec)
         bundle_path_match = Bundler.bundle_path.join("**").to_s
-
-        Dir.glob(File.join(root_path, package_pathspec, PACKAGE_CONFIG_FILENAME))
```
rochlefebvre, line 31: This doesn't work when more than one package path is provided, as an array.

lib/packwerk/package_set.rb
```diff
@@ -27,9 +27,11 @@ def load_all_from(root_path, package_pathspec: nil)
...
-          .map { |path| Pathname.new(path) }.reject { |path| path.realpath.fnmatch(bundle_path_match) }
+        glob_patterns = Array(package_pathspec).map do |pathspec|
```
rochlefebvre, line 30: `Array(x)` is more flexible, as it can handle a String path e.g. `"**/"` along with an array of paths coming from a `package.yml` file
```yml
package_paths:
- components/sales
- "."
```

test/support/fixture_helper.rb
```diff
@@ -0,0 +1,133 @@
+# typed: ignore
+# frozen_string_literal: true
+
+require "rails_test_helper"
+
+module FixtureHelper
+  TEMP_FIXTURE_DIR = ROOT.join("tmp", "fixtures").to_s
+  FileUtils.mkdir_p(TEMP_FIXTURE_DIR)
+
+  Minitest.after_run do
```
rochlefebvre, line 10: The idea was to copy every test's fixture directory inside this `tmp/fixtures` directory, and have each test delete their own copy under normal operating conditions.  I intended to have `Minitest.after_run` delete `tmp/fixtures` itself once the testing concludes, even in response to early termination.

I can't get this to work.  Any suggestions?

test/support/fixture_helper.rb
```diff
@@ -0,0 +1,133 @@
...
+  end
+
+  def around
+    old_working_dir = Dir.pwd
+    cache_rails_paths
+
+    yield
+
+    restore_rails_paths
+    Dir.chdir(old_working_dir)
```
rochlefebvre, line 22: If we don't change the working directory, then parts of Packwerk use the `packwerk` project directory as the application root, not `tmp/fixtures/the_current_test`. Is it possible to always use Configuration.root_path as the application folder?

test/support/fixture_helper.rb
```diff
@@ -0,0 +1,133 @@
...
+    FileUtils.remove_entry(@app_dir, true) if defined? @app_dir
+  end
+
+  def copy_template(template)
+    copy_dir("test/fixtures/#{template}")
+    Dir.chdir(app_dir)
```
rochlefebvre, line 28: From this point on, the test uses `<your packwerk project dir>/tmp/fixtures/the_current_test` as the working directory.

test/support/fixture_helper.rb
```diff
@@ -0,0 +1,133 @@
...
+module FixtureHelper
+  TEMP_FIXTURE_DIR = ROOT.join("tmp", "fixtures").to_s
+  FileUtils.mkdir_p(TEMP_FIXTURE_DIR)
+
+  Minitest.after_run do
+    # This does not get called on early exit, e.g. `Ctrl-C` or `(byebug) exit`.
+    FileUtils.remove_entry(TEMP_FIXTURE_DIR, true)
+  end
+
+  def around
```
rochlefebvre, line 15: By using `around` instead of `setup` and `teardown`, we don't need to worry that including test classes will forget to call `super` in their own `setup` and `teardown` methods. `around` runs regardless.

test/support/fixture_helper.rb
```diff
@@ -0,0 +1,133 @@
...
+    FileUtils.remove_entry(@app_dir, true) if defined? @app_dir
+  end
+
+  def copy_template(template)
+    copy_dir("test/fixtures/#{template}")
+    Dir.chdir(app_dir)
+
+    case template
+    when :minimal
+      rails_autoload_paths(paths_to("/components/sales/app/models"))
```
rochlefebvre, line 32: This sucky part is where we reconfigure the `Rails.application` singleton's bits that Packwerk cares about.

test/support/fixture_helper.rb
```diff
@@ -0,0 +1,133 @@
...
+      rails_autoload_once_paths(
+        paths_to(
+          "components/timeline/app/models",
+          "components/timeline/app/models/concerns",
+          "vendor/cache/gems/example/models",
+        )
+      )
+      rails_root_path(app_dir)
+      # make sure PrivateThing.constantize succeeds to pass the privacy validity check
+      require "fixtures/skeleton/components/timeline/app/models/private_thing.rb"
```
rochlefebvre, line 48: Moved from test/unit/application_validator_test.rb, but I'm unsure if it makes sense to move this kludge.

test/support/fixture_helper.rb
```diff
@@ -0,0 +1,133 @@
...
+    end
+  end
+
+  def copy_dir(path)
+    root = Dir.mktmpdir(name, TEMP_FIXTURE_DIR)
+    FileUtils.cp_r("#{path}/.", root)
+    @app_dir = root
+  end
+
+  def config
```
rochlefebvre, line 60: Any test can simply call `config` to retrieve the Configuration instance for their selected fixture's copy.

test/support/fixture_helper.rb
```diff
@@ -0,0 +1,133 @@
...
+    @config ||= Packwerk::Configuration.from_path(app_dir)
+  end
+
+  def merge_into_yaml_file(relative_path, hash)
```
rochlefebvre, line 64: Either create a new yaml file at the indicated path, or update the existing one by treating its contents as a hash and merging in the new hash.

If there is an existing array value, e.g.
```yml
# path/to/file.yml
a: some_value
b:
- 1
- 2
c:
- hi
```
Then calling 
```ruby
merge_into_yaml_file("path/to/file.yml", { b: [3,4], c: "bye" })
```
Will result in this yml
```yml
# path/to/file.yml
a: some_value
b:
- 1
- 2
- 3
- 4
c:
- hi
- bye
```

test/unit/application_validator_test.rb
```diff
@@ -3,83 +3,143 @@
...
-
 require "packwerk/application_validator"
 
-# make sure PrivateThing.constantize succeeds to pass the privacy validity check
-require "fixtures/skeleton/components/timeline/app/models/private_thing.rb"
-
 module Packwerk
   class ApplicationValidatorTest < Minitest::Test
+    include FixtureHelper
+
```
rochlefebvre, line 11: 8949c4452d08b4e874a81ba48f969b5ac52805b0 updates existing tests to use the fixture helper.  3583992ad0a31d10f88fb9b99d150e4ff3c46ba9 then adds new tests to ApplicationValidator.  Consider looking at each commit separately to follow the changes more easily.

test/unit/configuration_test.rb
```diff
@@ -27,28 +29,27 @@ class ConfigurationTest < Minitest::Test
...
         "inflections_file" => "custom_inflections.yml",
       }
-      YAML.expects(:load_file).with("./packwerk.yml").returns(configuration_hash)
 
-      configuration = Configuration.from_path(".")
+      merge_into_yaml_file("packwerk.yml", configuration_hash)
 
-      assert_equal ["xyz/*.rb"], configuration.include
-      assert_equal ["{exclude_dir,bin,tmp}/**/*"], configuration.exclude
-      assert_equal File.expand_path("."), configuration.root_path
```
rochlefebvre, line 36: This set configuration.root_path to `Shopify/packwerk`, and not `Shopify/packwerk/test/fixtures/skeleton`

lib/packwerk/application_validator.rb
doodzik: I think this bug fix could live in a separate PR. That way we can keep the scope of the PR a bit contained and merge the bug fix sooner.

rochlefebvre: Definitely! None of this will be merged in the future PR that will introduce FixtureHelper.  The fixes you see in the app part of Packwerk came about while I was writing new unit tests, and found problems in the app itself.  Otherwise, I'd have to write unit tests that validate incorrect behaviour.

test/support/fixture_helper.rb
doodzik: > The setup and teardown occur automatically by way of an around method, from https://github.com/splattael/minitest-around. Test suites are free to implement their own setup and teardown hooks without worrying about managing the fixture directory.

The `around` method gets used by including the `FixtureHelper` module in the test suite. It is not clear from that include in what order or when the fixtures' setup occurs. I would also add that having multiple setup/teardown aliases makes it harder to reason about what gets run when.

I prefer to call a fixture_setup/teardown method in the setup/teardown block and know exactly what gets executed in the test suite. I'm a little bit worried that the proposed usage of the `around` gem will make the test suite harder to maintain.

doodzik:
I really like your proposal. 🎉

[I have one issue with the `around` gem's usage](https://github.com/Shopify/packwerk/pull/59/files#r516869838), but that is detail and doesn't invalidate the proposal.

I want someone else to look at the Rails side of things in this PR because I don't have enough context on that.

test/support/fixture_helper.rb
rochlefebvre: I felt the same way when I first started out writing the helper module.  Here is what I learned about modules, super classes, and the setup/teardown hierarchy.

Say we have a simple `SetupTearDownModule` that we wish to include in various test suites:
```ruby
require "test_helper"
require 'active_support/concern'

module SetupTearDownModule
  extend ActiveSupport::Concern
  included do
    setup do
      puts "SetupTearDownModule#setup"
    end

    teardown do
      puts "SetupTearDownModule#teardown"
    end
  end
end

module Packwerk
  class SomeTest < Minitest::Test
    include SetupTearDownModule

    setup do
      puts "  SomeTest#setup"
    end

    teardown do
      puts "  SomeTest#teardown"
    end

    test "SomeTest#test" do
      puts "    SomeTest#test"
    end

    test "SomeTest#other_test" do
      puts "    SomeTest#other_test"
    end
  end
end
```
Running this test gives us this output:
```
# Running:

  SomeTest#setup
    SomeTest#other_test
  SomeTest#teardown
  SomeTest#setup
    SomeTest#test
  SomeTest#teardown
```
The SetupTearDown module's hooks are not executed because the including suite didn't call `super` in its own setup/teardown hooks.  Going forward, every including suite needs to concern itself with calling `super`:
```ruby
  class SomeTest < Minitest::Test
    include SetupTearDownModule

    setup do
      super()
      puts "  SomeTest#setup"
    end

    teardown do
      puts "  SomeTest#teardown"
      super()
    end
```
Which gives us the intended behaviour:
```
# Running:

  SomeTest#setup
    SomeTest#test
  SomeTest#teardown
  SomeTest#setup
    SomeTest#other_test
  SomeTest#teardown
```
Wait... what?  Turns out the whole `Concern/included do` bit defines `setup` and `teardown` in SomeTest, not in the SetupTeardownModule ancestor:
```ruby
[29, 38] in /Users/rochlefebvre/src/github.com/Shopify/packwerk/test/unit/some_test.rb
   29: module Packwerk
   30:   class SomeTest < Minitest::Test
   31:     include SetupTearDownModule
   32: 
   33:     byebug
=> 34:     setup do
   35:       super()
   36:       puts "  SomeTest#setup"
   37:     end
   38: 
(byebug) self
Packwerk::SomeTest
(byebug) self.instance_methods.grep(/setup/)
[:setup, :before_setup, :mocha_setup, :after_setup]
(byebug) self.instance_method(:setup).source_location
["/Users/rochlefebvre/src/github.com/Shopify/packwerk/test/unit/some_test.rb", 19]
```
Which is 
image
```ruby
(byebug) n
/Users/rochlefebvre/src/github.com/Shopify/packwerk/test/support/test_macro.rb:11: warning: method redefined; discarding old setup
/Users/rochlefebvre/src/github.com/Shopify/packwerk/test/unit/some_test.rb:19: warning: previous definition of setup was here

[34, 43] in /Users/rochlefebvre/src/github.com/Shopify/packwerk/test/unit/some_test.rb
   34:     setup do
   35:       super()
   36:       puts "  SomeTest#setup"
   37:     end
   38: 
=> 39:     teardown do
   40:       puts "  SomeTest#teardown"
   41:       super()
   42:     end
   43: 
(byebug) self.instance_method(:setup).source_location
["/Users/rochlefebvre/src/github.com/Shopify/packwerk/test/unit/some_test.rb", 34]
(byebug)
```
image

So we can't use TestMacro's `setup` and `teardown` macros to define the helper module's hooks.  Back to `def setup` and `def teardown`, then:

```ruby
module SetupTearDownModule
  def setup
    puts "SetupTearDownModule#setup"
  end

  def teardown
    puts "SetupTearDownModule#teardown"
  end
end
```
Now the test hooks run as expected
```
SetupTearDownModule#setup
  SomeTest#setup
    SomeTest#other_test
  SomeTest#teardown
SetupTearDownModule#teardown
SetupTearDownModule#setup
  SomeTest#setup
    SomeTest#test
  SomeTest#teardown
SetupTearDownModule#teardown
```
At this point, you're correct that setup and teardown hooks give us what we want and are less surprising that `around`, once we use `super`.  But what if we decide to introduce a super class for some suites like SomeTest?
```ruby
class SuperclassTest < Minitest::Test
  def setup
    super()
    puts "SuperclassTest#setup"
  end

  def teardown
    puts "SuperclassTest#teardown"
    super()
  end
end

module Packwerk
  class SomeTest < SuperclassTest
    include SetupTearDownModule

    setup do
      super()
      puts "  SomeTest#setup"
    end

    teardown do
      puts "  SomeTest#teardown"
      super()
    end

    test "SomeTest#test" do
      puts "    SomeTest#test"
    end

    test "SomeTest#other_test" do
      puts "    SomeTest#other_test"
    end
  end
end
```
```
# Running:

SetupTearDownModule#setup
  SomeTest#setup
    SomeTest#other_test
  SomeTest#teardown
SetupTearDownModule#teardown
SetupTearDownModule#setup
  SomeTest#setup
    SomeTest#test
  SomeTest#teardown
SetupTearDownModule#teardown

(╯°□°)╯︵ ┻━┻
```
By including our helper in the suite, we introduced `SetupTeardownModule`  in the calling chain, and the module hooks don't call `super`.  Ok, let's do that:
```ruby
module SetupTearDownModule
  def setup
    super()
    puts "SetupTearDownModule#setup"
  end

  def teardown
    puts "SetupTearDownModule#teardown"
    super()
  end
```
```
# Running:

SuperclassTest#setup
SetupTearDownModule#setup
  SomeTest#setup
    SomeTest#other_test
  SomeTest#teardown
SetupTearDownModule#teardown
SuperclassTest#teardown
.SuperclassTest#setup
SetupTearDownModule#setup
  SomeTest#setup
    SomeTest#test
  SomeTest#teardown
SetupTearDownModule#teardown
SuperclassTest#teardown
```
Everyone's hooks get called, but the order is not great.  In fact, in the context of my FixtureHelper, it means that no test super class may use any of the helper's functionality.

Now let's replace `SetupTearDownModule` with an `AroundModule` that defines an `around <setup> yield <teardown>` method.  Let's also get rid of `super` calls in `SuperclassTest` as well, and go back to the macro form of `setup do` and `teardown do`:
```ruby
module AroundModule
  def around
    puts "AroundModule#around - before yield"
    yield
    puts "AroundModule#around - after yield"
  end
end

class SuperclassTest < Minitest::Test
  setup do
    puts "SuperclassTest#setup"
  end

  teardown do
    puts "SuperclassTest#teardown"
  end
end

module Packwerk
  class SomeTest < SuperclassTest
    include AroundModule

    setup do
      super()
      puts "  SomeTest#setup"
    end

    teardown do
      puts "  SomeTest#teardown"
      super()
    end

    test "SomeTest#test" do
      puts "    SomeTest#test"
    end

    test "SomeTest#other_test" do
      puts "    SomeTest#other_test"
    end
  end
end
```
```
# Running:

AroundModule#around - before yield
SuperclassTest#setup
  SomeTest#setup
    SomeTest#test
  SomeTest#teardown
SuperclassTest#teardown
AroundModule#around - after yield
AroundModule#around - before yield
SuperclassTest#setup
  SomeTest#setup
    SomeTest#other_test
  SomeTest#teardown
SuperclassTest#teardown
AroundModule#around - after yield
```
Now, the [around](https://github.com/splattael/minitest-around/blob/master/lib/minitest/around/unit.rb) logic neatly bookends the entire test setup - execute - teardown flow.  I feel this is a more sustainable approach.  Our macros work, no need to call `super` everywhere, and test class inheritance works.

doodzik: Oh my, that is quite a hairy issue 😅 Thank you for giving so much detail, which I appreciate.

I was thinking more along the lines of not hooking into the lifecycle methods at all, which would allow us to circumvent the issue completely. It isn't as smart and thus requires more typing, but I think it would make the setup a bit simpler and also more transparent. WDYT?

```ruby
module SetupTearDownModule
    def setup_fixture
      puts "SetupTearDownModule#setup"
    end

    def teardown_fixture
      puts "SetupTearDownModule#teardown"
    end
  end
end

module Packwerk
  class SomeTest < Minitest::Test
    include SetupTearDownModule

    setup do
      setup_fixture
      puts "  SomeTest#setup"
    end

    teardown do
      teardown_fixture
      puts "  SomeTest#teardown"
    end

    test "SomeTest#test" do
      puts "    SomeTest#test"
    end
  end
end
```

rochlefebvre: Ah, I understand your proposal now.  Let's call your pattern the _explicit approach_, v.s. the _implicit approach_ in the PR.

As you pointed out, the explicit  approach requires more typing for every test suite that choses to include the helper. At minimum, they need to do four things:
- Add a `setup` method;
- Call the module's `setup_fixtures` method before they invoke any code that touches the application directory, the Rails application singleton, or the current working directory.
- Add a `teardown` method;
- Call the module's `teardown_fixtures` method so they don't leak the temporary context into subsequent tests and suites.

There is no picking and choosing which steps to perform, only when to call `setup_fixtures` and `teardown_fixtures` relative to other setup/teardown logic.

So what do we get out of the explicit approach? You mentioned simplicity and transparency.  It's true that Packwerk's gem setup is simpler without `minitest_around`, and so is the testing runtime.   But now we need to do all of the work I listed above, every time, without fail.

Transparency is desireable when the person should be concerned with what is happening automagically.  I don't need to know how MacOS is currently managing my process memory in physical RAM.  I _may_ need to know what the current working directory is when instantiating a new Packwerk::Configuration object.  If I'm writing a new test suite in Packwerk and I had to read up on FixtureHelper's `setup_fixtures` method before calling it, I may be more aware of its side effects.

Transparency is also desirable when there's a chance that the person would need to do things differently depending on the situation.  Here, I think flexibility is neither needed nor desireable.  The test either runs in the temporary fixture sandbox (including the updated Rails singleton and application directory), or it doesn't.  In fact, if you call the helper's setup and teardown code early/late enough, then you can even mess with the Rails application load paths for your own testing purposes, knowing that the fixture helper's teardown will undo your changes.  The sandbox doesn't constrain your own test setup.

Last point: if you introduce a test superclass with its own setup and teardown, then the explicit approach requires a lot of mindfulness in calling `super`, `setup_fixtures`, along with your own setup code:

```ruby
module SetupTearDownModule
  def setup_fixture
    puts "SetupTearDownModule#setup"
  end

  def teardown_fixture
    puts "SetupTearDownModule#teardown"
  end
end

module Packwerk
  class SomeTestSuperclass < Minitest::Test
    setup do
      puts "  SomeTestSuperclass#setup"
    end

    teardown do
      puts "  SomeTestSuperclass#teardown"
    end
  end

  class SomeTest < SomeTestSuperclass
    include SetupTearDownModule

    setup do
      setup_fixture             # <-- needs to happen here
      super()                   # otherwise SomeTestSuperclass#setup will have a different context, e.g. cwd
      puts "    SomeTest#setup"
    end

    teardown do
      puts "    SomeTest#teardown"
      super()
      teardown_fixture
    end

    test "SomeTest#test" do
      puts "      SomeTest#test"
    end
  end
end
```
```
# Running:

SetupTearDownModule#setup
  SomeTestSuperclass#setup
    SomeTest#setup
      SomeTest#test
    SomeTest#teardown
  SomeTestSuperclass#teardown
SetupTearDownModule#teardown
```

lib/packwerk/application_validator.rb
exterm: Not sure this should be in the same PR as your new testing pattern.

Also, If we start reordering these, I think we should do

- all application integration tests first (autoload path cache, inflections, application structure)
- all package manifest checks afterwards (manifest syntax, manifest privacy, acyclic graph etc)

lib/packwerk/package_set.rb
exterm: Do we support an array of paths or are you adding a feature here?

test/support/fixture_helper.rb
exterm: Why is it important to delete a temporary directory that is on the temp drive? It should be cleaned up by the system at some point, right?

If you are concerned about state leaking into the next test, maybe clear the data at the start of the test?

exterm: I'd love if we could hook into the garbage collector for this, that would be the most elegant solution. Not sure that's possible in Ruby.

tomstuart: Is there a way we could use [`Dir.mktmpdir`](https://www.rubydoc.info/stdlib/tmpdir/Dir#mktmpdir-class_method) for this? In RSpec I’d put it in an `around` hook and run the example inside its block; I don’t know whether there’s an easy equivalent for minitest.

lib/packwerk/application_validator.rb
rochlefebvre: Definitely not in the same PR.  This is just a throwaway POC, where I occasionally needed to make unrelated changes in order to unblock the larger demo.

I believe one of my new unit tests could not exercise what I wanted to demonstrate until I reordered the checks.

lib/packwerk/package_set.rb
rochlefebvre: The plural form of the configuration key, `package_paths`, tells me that more than one path could be provided.  It also feels like something we should support, especially once we fix the existing logic.

But `File.join(root_path, package_pathspec, PACKAGE_CONFIG_FILENAME)` will not produce a viable path if `package_pathspec` is an array nor a comma-delimited set of paths.

test/support/fixture_helper.rb
rochlefebvre: In this POC, I'm toying with writing the temporary data in packwerk's own `/tmp` directory, instead of the system's temporary directory.  While I was debugging tests, I found it quite convenient to see the temporary files right in the project, using my code editor.  It's not so fun to figure out where the fixtures go in the system temp directory.

Also, if ever there is a leak of sorts with temporary fixture directories, it's easier to spot, being right in the project workspace.

rochlefebvre: >Why is it important to delete a temporary directory that is on the temp drive? It should be cleaned up by the system at some point, right?

To be clear, the fixures are somewhere in packwerk's own workspace:
```ruby
TEMP_FIXTURE_DIR = ROOT.join("tmp", "fixtures").to_s
# <your packwerk project dir>/tmp/fixures/the_current_test
```

>In RSpec I’d put it in an around hook and run the example inside its block so that the temporary directory gets cleaned up automatically

Under normal exit conditions, my temporary directory gets cleaned up.  I'm struggling with getting cleanup logic to fire during `Ctrl-C` or `(byebug) exit`.  I don't know if RSpec's `around` block would get to finish under equivalent circumstances?

test/support/fixture_helper.rb
exterm: This should be relevant only for the test for the application validator. No other test in packwerk should depend on anything in Rails.

test/support/fixture_helper.rb
```diff
@@ -0,0 +1,133 @@
...
+
+  def rails_root_path(path)
+    Rails.application.config.root = path
+  end
+
+  def cache_rails_paths
+    @rails_autoload_paths = Rails.application.config.autoload_paths
+    @rails_eager_load_paths = Rails.application.config.eager_load_paths
+    @rails_autoload_once_paths = Rails.application.config.autoload_once_paths
+    @rails_root_path = Rails.application.config.root
```
exterm, line 112: Again: Only the application validator test should depend on parts of a Rails application.

test/support/fixture_helper.rb
rochlefebvre: ApplicationValidator gets the paths from Configuration.  A test author could provide the correct temporary directory load paths through the configuration object as long as they know exactly what those paths are:
```ruby
def fixture_load_paths(template = :minimal)
  case template
  when :minimal
    paths_to("/components/sales/app/models")
  when :skeleton
    paths_to(
      "components/sales/app/models",
      "components/platform/app/models",
      #...
      "vendor/cache/gems/example/models"
    )
  else
    raise SomeError
  end
end
# ...
test "Something that cares about load paths" do
  config_params = read_fixture_packerk_yml_as_hash.merge(
    load_paths: fixture_load_paths(:the_fixure),
    probably_other_params: as_well
  )
  config = Packwerk::Configuration.new(config_params, config_path: app_dir)
  # ...
end
```

That decouples tests that are sensitive to load paths from the Rails singleton.  But yuck.

ApplicationValidator tests care about load paths, as you said.  The test suite for Configuration also cares, and it benefits from my proposed fixture helper setup/teardown.  To a much lesser degree, some of the Cli's tests are affected by load paths.  We currently set up that context via mocking.

I feel that the FixtureHelper's value is in setting up a complete, coherent runtime sandbox for a given test.  That includes a proper application directory, current working directory, and Rails singleton.  Although most tests will not be impacted by coherent Rails paths, test authors get to work in a setting where calling `Configuration.new` or `Configuration.from_path(app_dir)` will just work.

test/support/fixture_helper.rb
doodzik: I like the implicit approach because it is dead simple for the user to work with, but I'm not sure if it is required for this POC. I think what I was trying to get at was that it distracts from the rest of your proposal, which doesn't mean that we shouldn't implement it. I would suggest implementing your proposal without the implicit approach and then creating a follow-up PR that adds the implicit testing approach. That way, we can get into more details and see the actual impact on the codebase.

> So, what do we get out of the explicit approach? You mentioned simplicity and transparency. Packwerk's gem setup is indeed simpler without minitest_around, and so is the testing runtime. But now we need to do all of the work I listed above, every time, without fail.

It makes a big difference in how much we abstract our code and move away from a standard. This change is adding a gem and changes the default behavior of minitest, which are two changes I would like to be confident about because they have implications for this repository's maintainability. If we add a handful of test suites that are using the fixture helper, then it is probably premature to make these changes.

> Last point: if you introduce a test superclass with its own setup and teardown, then the explicit approach requires a lot of mindfulness in calling super, setup_fixtures, along with your own setup code:

I would argue that we should not rely on inheritance in defining our tests but rather use module composition. Using different approaches in defining test cases makes them brittle and hard to maintain because we increase the variables at play. This is something we learned from our test suite in core.

> Transparency is also desirable when there's a chance that the person would need to do things differently depending on the situation. Here, I think flexibility is neither needed nor desirable.

I think this is a good point, and I want to say that I'm not opposed to adding the implicit approach for all the reasons you mentioned. I want to wait and see if the explicit approach is actually going to warrant a change of minitest's default behavior and the addition of another dependency.

test/support/fixture_helper.rb
exterm: if `Minitest.after_run` doesn't do it, have you tried an `ensure` in the `around` block?

test/unit/configuration_test.rb
```diff
@@ -6,7 +6,7 @@
 
 module Packwerk
   class ConfigurationTest < Minitest::Test
-    include FixtureHelper
+    include RailsFixtureHelper
```
exterm, line 9: Love this new commit!

It makes the dependencies on Rails more obvious. We can now see that `Configuration` needs `Rails` around to work, and we should change that (obviously in a separate PR).

test/support/fixture_helper.rb
rochlefebvre: [Three days later] I just tried it and it works! Nice.  There will be no need for this Minitest.after_run nor garbage collection hook after all.  And it made me realize how I should generally write `around` methods as
```ruby
def around
  setup_the_stuff
  yield
ensure
  tear_down_the_stuff
end
```

test/support/fixture_helper.rb
rochlefebvre: Here is a pattern that gives us most of the implicit approach's benefits without committing to the minitest-around gem:

- per-test fixture copying and runtime caching + configuration;
- per-test cleanup, even during early termination (Ctrl+C or byebug#exit)
- support for including unrelated modules in a test suite, not dependent on inclusion order (i.e. ancestry).

We can move to an `around` model to automate many of the remaining details at a later time.

```ruby 
# typed: ignore
# frozen_string_literal: true

require "test_helper"

module SetupTearDownModule
  def setup_fixture
    puts "SetupTearDownModule#setup"
  end

  def teardown_fixture
    puts "SetupTearDownModule#teardown"
  end
end

module Packwerk
  module SomeOtherHelperModule
    def setup
      puts "  SomeOtherHelperModule#setup"
    end

    def teardown
      puts "  SomeOtherHelperModule#teardown"
    end
  end

  class SomeTest < Minitest::Test
    include SetupTearDownModule
    include SomeOtherHelperModule

    setup do
      setup_fixture             # <-- needs to happen here
      super()                   # otherwise SomeOtherHelperModule#setup will have a different context, e.g. cwd
      puts "    SomeTest#setup"
    end

    teardown do
      puts "    SomeTest#teardown"
      super()
      teardown_fixture
    end

    test "SomeTest#test" do
      puts "      SomeTest#test"
    end
  end
end
```
```
# Running:

SetupTearDownModule#setup
  SomeOtherHelperModule#setup
    SomeTest#setup
      SomeTest#test
    SomeTest#teardown
  SomeOtherHelperModule#teardown
SetupTearDownModule#teardown
```

test/unit/configuration_test.rb
exterm: Here's the [separate PR](https://github.com/Shopify/packwerk/pull/68)

Issue #12, opened 2020-09-25
Fail check if shitlisted violation is removed


wildmaples:
Fail check if shitlisted violation is removed or not updated.

We have a fix for this bug on Shopify/shopify - https://github.com/Shopify/shopify/pull/250880. However, we need to introduce a permanent fix for other repositories.

Options: 

- Have `packwerk update --exit-status` or maybe `packwerk update --check-stale`, which returns an exit status: 0 if there are no stale violations; and 1 if there is a stale violation. This can be easily used on CI to ensure there are no stale violations. The flag will run `packwerk update` and check to see if there is a difference in the shitlists. Though I am concerned that this check would also be flagging new violations as stale and encouraging users to just run `packwerk update` to update the shitlist.

- Another option is to have `packwerk check` understand stale references. This would require more refactoring and unifying the idea of `Shitlists` and `UpdatingShitlists`. This option is preferred as the former would cost us some memory.

wildmaples:
We should also be considering https://github.com/Shopify/packwerk/issues/49 while solving this issue. Both involve modifying how `packwerk update` currently runs.

wildmaples:
This issue is fixed in #61 . 

You can now run `packwerk detect-stale-violations` and the exit code will indicate whether or not the current system has stale violations in the `deprecated_references.yml` file. If so, run `packwerk update`.

PR #61, opened 2020-11-03
Packwerk checks for stale violations

cindygshopify:
Adding feature for Packwerk checking stale violations. This should be incorporated into CI in the future to make sure that PRs don't pass CI if there are stale violations. Linked issue here: https://github.com/Shopify/packwerk/issues/12

We had to refactor the code a bit to create this feature. We needed a lot of functionality in the UpdatingDeprecatedReferences.rb class but putting the change in that class didn't seem right. So CacheDeprecatingReferences.rb was created as a parent class that both UpdatingDeprecatedReferences.rb and DetectStaleDeprecatedReferences.rb now inherit from. Functionality common to both now belongs in CacheDeprecatingReferences.rb and functionality specific to updating vs detecting stale violations are now in their respective classes.

When executing the command, Packwerk will go through all the loaded files and check for violations in the files which it then stores in DetectStaleDeprecatedReferences.rb. At the end when all files were processed, we compare the deprecated_references.yml file of each component loaded to the actual violations generated. If there are references in deprecated_references.yml that aren't in the generated violations, then there are stale violations. The program will exits with status code 1. If there are no stale violations, program exits with status code 0.

Any issues with the code refactor? Are there stale violation cases we may have missed with this code change?

## Tophatting

My tophatting instructions use Packwerk in the shopify/shopify repo, but you can use another repo if you prefer

1) Pull this branch into local
2) Change the gemfile of local shopify/shopify repo to point to this branch
3) Modify dev.yml of shopify/shopify to include the detect-stale violations command like so:
Screen Shot 2020-11-03 at 5 46 32 PM
4) Generate a stale violation in shopify/shopify. 

- You can do this by changing package.yml of a component, enforcing dependencies on it and run Packwerk update to generate a deprecated_references.yml for that component which has boundary violations. 

- After you have a list of dependency violations for that component, you can add the component which the violation is coming from as a dependency in the  package.yml of your component

5) Run `dev packages detect-stale-violations path\to\your\component`
6) Check that it outputs stale violations were made
7) Run `$?` to check status of last command (0 if no stale violations, 1 if there were stale violations)

## Type of Change

- [X] New feature (non-breaking change which adds functionality)

### Additional Release Notes

## Checklist

- [X] It is safe to simply rollback this change.

### DIFF
--- a/lib/packwerk/cache_deprecated_references.rb
+++ b/lib/packwerk/cache_deprecated_references.rb
@@ -1,0 +1,45 @@
+# typed: true
+# frozen_string_literal: true
+
+require "sorbet-runtime"
+
+require "packwerk/deprecated_references"
+require "packwerk/reference"
+require "packwerk/reference_lister"
+require "packwerk/violation_type"
+
+module Packwerk
+  class CacheDeprecatedReferences
+    extend T::Sig
+    include ReferenceLister
+
+    def initialize(root_path, deprecated_references = {})
+      @root_path = root_path
+      @deprecated_references = deprecated_references
+    end
+
+    sig do
+      params(reference: Packwerk::Reference, violation_type: ViolationType)
+        .returns(T::Boolean)
+        .override
+    end
+    def listed?(reference, violation_type:)
+      deprecated_references = deprecated_references_for(reference.source_package)
+      deprecated_references.add_entries(reference, violation_type.serialize)
+      true
+    end
+
+    private
+
+    def deprecated_references_for(package)
+      @deprecated_references[package] ||= Packwerk::DeprecatedReferences.new(
+        package,
+        deprecated_references_file_for(package),
+      )
+    end
+
+    def deprecated_references_file_for(package)
+      File.join(@root_path, package.name, "deprecated_references.yml")
+    end
+  end
+end
--- a/lib/packwerk/cache_deprecated_references.rb
+++ b/lib/packwerk/cache_deprecated_references.rb
@@ -11,11 +11,13 @@ require "packwerk/violation_type"
 module Packwerk
   class CacheDeprecatedReferences
     extend T::Sig
+    extend T::Helpers
     include ReferenceLister
+    abstract!
 
     def initialize(root_path, deprecated_references = {})
       @root_path = root_path
-      @deprecated_references = deprecated_references
+      @deprecated_references = T.let(deprecated_references, T::Hash[String, Packwerk::DeprecatedReferences])
     end
 
     sig do
--- a/lib/packwerk/cli.rb
+++ b/lib/packwerk/cli.rb
@@ -12,10 +12,13 @@ require "packwerk/output_styles"
 require "packwerk/run_context"
 require "packwerk/updating_deprecated_references"
 require "packwerk/checking_deprecated_references"
+require "packwerk/commands/detect_stale_violations_command"
+require "packwerk/commands/offense_progress_marker"
 
 module Packwerk
   class Cli
     extend T::Sig
+    include OffenseProgressMarker
 
     def initialize(run_context: nil, configuration: nil, out: $stdout, err_out: $stderr, style: OutputStyles::Plain)
       @out = out
@@ -45,6 +48,8 @@ module Packwerk
         generate_configs
       when "check"
         check(args)
+      when "detect-stale-violations"
+        detect_stale_violations(args)
       when "update"
         update(args)
       when "update-deprecations"
@@ -142,7 +147,9 @@ module Packwerk
       all_offenses = T.let([], T.untyped)
       execution_time = Benchmark.realtime do
         all_offenses = files.flat_map do |path|
-          @run_context.process_file(file: path).tap { |offenses| mark_progress(offenses) }
+          @run_context.process_file(file: path).tap do |offenses|
+            mark_progress(offenses: offenses, progress_formatter: @progress_formatter)
+          end
         end
 
         updating_deprecated_references.dump_deprecated_references_files
@@ -165,7 +172,7 @@ module Packwerk
       execution_time = Benchmark.realtime do
         files.each do |path|
           @run_context.process_file(file: path).tap do |offenses|
-            mark_progress(offenses)
+            mark_progress(offenses: offenses, progress_formatter: @progress_formatter)
             all_offenses.concat(offenses)
           end
         end
@@ -181,19 +188,23 @@ module Packwerk
       all_offenses.empty?
     end
 
+    def detect_stale_violations(paths)
+      detect_stale_violations = DetectStaleViolationsCommand.new(
+        files: fetch_files_to_process(paths),
+        configuration: @configuration,
+        progress_formatter: @progress_formatter
+      )
+      result = detect_stale_violations.run
+      @out.puts
+      @out.puts(result.message)
+      result.status
+    end
+
     def fetch_files_to_process(paths)
       files = FilesForProcessing.fetch(paths: paths, configuration: @configuration)
       abort("No files found or given. "\
         "Specify files or check the include and exclude glob in the config file.") if files.empty?
       files
-    end
-
-    def mark_progress(offenses)
-      if offenses.empty?
-        @progress_formatter.mark_as_inspected
-      else
-        @progress_formatter.mark_as_failed
-      end
     end
 
     def validate(_paths)
--- a/lib/packwerk/deprecated_references.rb
+++ b/lib/packwerk/deprecated_references.rb
@@ -3,6 +3,7 @@
 
 require "sorbet-runtime"
 require "yaml"
+require "sorbet-runtime"
 
 require "packwerk/reference"
 require "packwerk/reference_lister"
@@ -47,6 +48,7 @@ module Packwerk
       @new_entries[reference.constant.package.name] = package_violations
     end
 
+    sig { returns(T::Boolean) }
     def stale_violations?
       prepare_entries_for_dump
       deprecated_references.any? do |package, package_violations|
--- a/lib/packwerk/deprecated_references.rb
+++ b/lib/packwerk/deprecated_references.rb
@@ -47,6 +47,23 @@ module Packwerk
       @new_entries[reference.constant.package.name] = package_violations
     end
 
+    def stale_violations?
+      prepare_entries_for_dump
+      deprecated_references.any? do |package, package_violations|
+        package_violations.any? do |constant_name, entries_for_file|
+          new_entries_violation_types = @new_entries.dig(package, constant_name, "violations")
+          return true if new_entries_violation_types.nil?
+          if entries_for_file["violations"].all? { |type| new_entries_violation_types.include?(type) }
+            stale_violations =
+              entries_for_file["files"] - Array(@new_entries.dig(package, constant_name, "files"))
+            stale_violations.present?
+          else
+            return true
+          end
+        end
+      end
+    end
+
     def dump
       if @new_entries.empty?
         File.delete(@filepath) if File.exist?(@filepath)
--- a/lib/packwerk/detect_stale_deprecated_references.rb
+++ b/lib/packwerk/detect_stale_deprecated_references.rb
@@ -1,0 +1,12 @@
+# typed: true
+# frozen_string_literal: true
+
+require "packwerk/cache_deprecated_references"
+
+module Packwerk
+  class DetectStaleDeprecatedReferences < CacheDeprecatedReferences
+    def stale_violations?
+      @deprecated_references.values.any?(&:stale_violations?)
+    end
+  end
+end
--- a/lib/packwerk/detect_stale_deprecated_references.rb
+++ b/lib/packwerk/detect_stale_deprecated_references.rb
@@ -1,10 +1,12 @@
-# typed: true
+# typed: strict
 # frozen_string_literal: true
 
 require "packwerk/cache_deprecated_references"
 
 module Packwerk
   class DetectStaleDeprecatedReferences < CacheDeprecatedReferences
+    extend T::Sig
+    sig { returns(T::Boolean) }
     def stale_violations?
       @deprecated_references.values.any?(&:stale_violations?)
     end
--- a/lib/packwerk/updating_deprecated_references.rb
+++ b/lib/packwerk/updating_deprecated_references.rb
@@ -1,51 +1,14 @@
 # typed: true
 # frozen_string_literal: true
 
-require "sorbet-runtime"
-
-require "packwerk/deprecated_references"
-require "packwerk/reference"
-require "packwerk/reference_lister"
-require "packwerk/violation_type"
+require "packwerk/cache_deprecated_references"
 
 module Packwerk
-  class UpdatingDeprecatedReferences
-    extend T::Sig
-    include ReferenceLister
-
-    def initialize(root_path, deprecated_references = {})
-      @root_path = root_path
-      @deprecated_references = deprecated_references
-    end
-
-    sig do
-      params(reference: Packwerk::Reference, violation_type: ViolationType)
-        .returns(T::Boolean)
-        .override
-    end
-    def listed?(reference, violation_type:)
-      deprecated_references = deprecated_references_for(reference.source_package)
-      deprecated_references.add_entries(reference, violation_type.serialize)
-      true
-    end
-
+  class UpdatingDeprecatedReferences < CacheDeprecatedReferences
     def dump_deprecated_references_files
       @deprecated_references.each do |_, deprecated_references_file|
         deprecated_references_file.dump
       end
-    end
-
-    private
-
-    def deprecated_references_for(package)
-      @deprecated_references[package] ||= Packwerk::DeprecatedReferences.new(
-        package,
-        deprecated_references_file_for(package),
-      )
-    end
-
-    def deprecated_references_file_for(package)
-      File.join(@root_path, package.name, "deprecated_references.yml")
     end
   end
 end
--- a/test/unit/cache_deprecated_references_test.rb
+++ b/test/unit/cache_deprecated_references_test.rb
@@ -4,12 +4,12 @@
 require "test_helper"
 
 module Packwerk
-  class UpdatingDeprecatedReferencesTest < Minitest::Test
+  class CacheDeprecatedReferencesTest < Minitest::Test
     setup do
-      @updating_deprecated_references = UpdatingDeprecatedReferences.new(".")
+      @updating_deprecated_references = CacheDeprecatedReferences.new(".")
     end
 
-    test "#listed? returns true when constant has been added to file" do
+    test "#listed? adds entry and returns true" do
       File.stubs(:open)
 
       source_package = Package.new(name: "source_package", config: {})
--- a/test/unit/deprecated_references_test.rb
+++ b/test/unit/deprecated_references_test.rb
@@ -25,6 +25,111 @@ module Packwerk
       )
     end
 
+    test "#stale_violations? returns true if deprecated references exist but no violations can be found in code" do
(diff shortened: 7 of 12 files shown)


lib/packwerk/deprecated_references.rb
```diff
@@ -47,6 +47,20 @@ def add_entries(reference, violation_type)
       @new_entries[reference.constant.package.name] = package_violations
     end
 
+    def stale_violations?
+      deprecated_references.each_key do |package|
```
tomstuart, line 51: This whole structure…

```ruby
deprecated_references.each_key do |package|
  deprecated_references[package].each_key do |constant_name|
    # …
    stale_violations = … deprecated_references.dig(package, constant_name, "files") …
    # …
    if stale_violations.present?
      return true
    end
  end
end
false
```

…could instead be written as:

```ruby
deprecated_references.any? do |package, package_violations|
  package_violations.any? do |constant_name, entries_for_file|
    # …
    stale_violations = … entries_for_file["files"] …
    # …
    stale_violations.present?
  end
end
```

The changes are:

1. use `Enumerable#any?` to check as many elements as necessary and decide whether to return `true` or `false` overall, rather than using `Hash#each_key` and breaking out of the loop explicitly ourselves;
2. use the values as well as the keys from the outer `deprecated_references` hash as we iterate over it, rather than throwing away the values and then using the keys to look them up again;
3. likewise, use the values as well as the keys from the inner `package_violations` hash as we iterate over it rather than `dig`ing into the outer `deprecated_references`; and
4. implicitly returning `stale_violations.present?` from the inner block (for use by `#any?`) rather than breaking out with an explicit `return`.

I think the second version makes it easier to see at a glance that we’re iterating over every violation in `deprecated_references` (two levels deep) and checking whether `any?` of them are stale. It also happens to be a little bit more concise which is a nice bonus. What do you think?

lib/packwerk/detect_stale_deprecated_references.rb
```diff
@@ -0,0 +1,14 @@
+# typed: true
+# frozen_string_literal: true
+
+require "packwerk/cache_deprecated_references"
+
+module Packwerk
+  class DetectStaleDeprecatedReferences < CacheDeprecatedReferences
+    def stale_violations?
+      @deprecated_references.any? do |_, deprecated_references_file|
```
tomstuart, line 9: You could do…

```suggestion
      @deprecated_references.values.any? do |deprecated_references_file|
```

…instead if you prefer that.

lib/packwerk/updating_deprecated_references.rb
```diff
@@ -5,7 +5,6 @@
 
 module Packwerk
   class UpdatingDeprecatedReferences < CacheDeprecatedReferences
-
```
tomstuart, line 31: 👀

test/unit/deprecated_references_test.rb
```diff
@@ -25,6 +25,79 @@ class DeprecatedReferencesTest < Minitest::Test
       )
     end
 
+    test "#stale_violations? returns true if deprecated references exist but no violations can be found in code" do
```
tomstuart, line 28: These four tests are (very nicely! 👏 ) unit testing just `DeprecatedReferences#stale_violations?`, which doesn’t depend on any of the other machinery being added simultaneously in `DetectStaleDeprecatedReferences` and `Cli`. Would you consider pulling this out into a separate commit, i.e. first implement & test `DeprecatedReferences#stale_violations?`, then add everything else afterwards? Having these in separate commits would make it more obvious which changes go together.

test/unit/detect_stale_deprecated_references_test.rb
```diff
@@ -0,0 +1,47 @@
+# typed: false
+# frozen_string_literal: true
+
+require "test_helper"
+require "packwerk/detect_stale_deprecated_references"
+require "byebug"
+
+module Packwerk
+  class DetectStaleDeprecatedReferencesTest < Minitest::Test
```
tomstuart, line 9: Ditto my comment about the other unit tests: after `DeprecatedReferences#stale_violations?` is added, these tests could go in their own commit along with the `DetectStaleDeprecatedReferences#stale_violations?` implementation they’re testing, and then a third commit could finally implement `Cli#detect_stale_violations`. As well as making it clearer which changes go together, this sequence of commits would make it clearer which pieces build on top of which other pieces.

lib/packwerk/cli.rb
```diff
@@ -177,6 +180,33 @@ def check(paths)
       all_offenses.empty?
     end
 
+    def detect_stale_violations(paths)
```
tomstuart, line 183: There’s no test for this. There are already other subcommands (e.g. `packwerk check`) which have no unit test, but those are covered by the integration tests. Is there an easy way to have _some_ automated test coverage of `packwerk detect-stale-violations`, whether unit or integration, so that it doesn’t get accidentally broken?

tomstuart (commented):
Thank you for doing this, Cindy!

No showstoppers here but I had a few ideas while reading through your code so I’ve added some comments for now in case you’d like to discuss them. Let me know what you think.

lib/packwerk/cache_deprecated_references.rb
```diff
@@ -0,0 +1,45 @@
+# typed: true
```
doodzik, line 1: I think it would be good that new files to this codebase are in `strict` mode. A lot of the benefits from sorbet only arise if we have enough type information available.

lib/packwerk/cli.rb
doodzik: @rochlefebvre was working on an improvement in this PR https://github.com/Shopify/packwerk/pull/59 so it could make sense to have a look at that.

We are adding more code to the cli class, which makes it harder to maintain. I think it would make sense to move the logic of this command into a separate class, which would make it easier to reason about the code and subsequently test.

lib/packwerk/detect_stale_deprecated_references.rb
```diff
@@ -0,0 +1,14 @@
+# typed: true
+# frozen_string_literal: true
+
+require "packwerk/cache_deprecated_references"
+
+module Packwerk
+  class DetectStaleDeprecatedReferences < CacheDeprecatedReferences
+    def stale_violations?
```
doodzik, line 8: Same as my previous comment. Can you add type annotations to this class?

lib/packwerk/cache_deprecated_references.rb
cindygshopify: While that would be nice, these methods are extracted from what used to be UpdatingDeprecatedReferences class and unfortunately I don't have enough context to know what the signatures for these classes would be.

lib/packwerk/detect_stale_deprecated_references.rb
cindygshopify: done

lib/packwerk/cli.rb
cindygshopify: @doodzik I'm not against moving the logic of it into another class, but I'm not really sure what the class would be. Currently I just wrote an additional test for cli_test.rb as it seemed like that was the best place to put a test for this function as many of the other command line related tests are also there. Anyway, let me know if you think there should be changes. I considered it but also didn't want to break the current pattern.

test/unit/detect_stale_deprecated_references_test.rb
cindygshopify: I went ahead and did this. let me know if you have any concerns

lib/packwerk/cli.rb
```diff
@@ -70,8 +73,6 @@ def execute_command(args)
       end
     end
 
-    private
```
wildmaples, line 73: Why was this removed?

lib/packwerk/deprecated_references.rb
```diff
@@ -7,6 +7,7 @@
 require "packwerk/reference"
 require "packwerk/reference_lister"
 require "packwerk/violation_type"
+require "byebug"
```
wildmaples, line 10: Probably didn't mean to include this line

lib/packwerk/cli.rb
cindygshopify: Testing purposes. The private keyword is added back later on in the file for methods that don't get called from `execute_command`

wildmaples (commented):
Yay this is great work!

test/unit/cache_deprecated_references_test.rb
```diff
@@ -4,12 +4,12 @@
 require "test_helper"
 
 module Packwerk
-  class UpdatingDeprecatedReferencesTest < Minitest::Test
```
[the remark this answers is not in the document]
cindygshopify: well I simply since we extracted CacheDeprecatedReferences from UpdatingDeprecatedReferences, the tests that were in UpdatingDeprecatedReferences were extracted to CacheDeprecatedReferences as the methods that are being tested now belong to CacheDeprecatedReferences. Now the UpdatingDeprecatedReferencesclass  just has one method that is specific to its' own class, but because no test was ever written for that method, UpdatingDeprecatedReferences now doesn't have a test class.

lib/packwerk/cache_deprecated_references.rb
doodzik: I think it is important that we continue adopting sorbet in this codebase because it will improve the maintainability of the code over time. It is a bit inconvenient to look up the signatures but it is beneficial in the long term.

You can add type signatures to this class pretty easily and change to strict:

```ruby
   
    sig { params(root_path: String, deprecated_references: T::Hash[String, Packwerk::DeprecatedReferences]) }
    def initialize(root_path, deprecated_references = {})
      @root_path = root_path
      @deprecated_references = T.let(deprecated_references, T::Hash[String, Packwerk::DeprecatedReferences])
    end

    sig { params(package: Package).returns(Packwerk::DeprecatedReferences) } 
    def deprecated_references_for(package)
      @deprecated_references[package] ||= Packwerk::DeprecatedReferences.new(
        package,
        deprecated_references_file_for(package),
      )
    end

    sig { params(package: Package).returns(String) } 
    def deprecated_references_file_for(package)
      File.join(@root_path, package.name, "deprecated_references.yml")
    end
```

lib/packwerk/detect_stale_deprecated_references.rb
doodzik: 🙌

lib/packwerk/cli.rb
```diff
@@ -181,6 +182,35 @@ def check(paths)
       all_offenses.empty?
     end
 
+    def detect_stale_violations(paths,
```
doodzik, line 185: This class is getting too big and complicated to maintain. I think it would be good to apply the [scout rule](https://www.oreilly.com/library/view/97-things-every/9780596809515/ch08.html) here, leaving a class cleaner than you found it. 

I would suggest moving your changes to a separate class so we can stop the rot. This should also be pretty straightforward since it doesn't require a big refactor.

lib/packwerk/cli.rb
```diff
@@ -181,6 +182,35 @@ def check(paths)
...
+      reference_lister: ::Packwerk::DetectStaleDeprecatedReferences.new(@configuration.root_path))
```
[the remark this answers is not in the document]
cindygshopify: I don't think that link is working for me. Which line were you referring to?

lib/packwerk/cli.rb
```diff
@@ -181,6 +182,35 @@ def check(paths)
...
+
+      all_offenses = T.let([], T.untyped)
+      execution_time = Benchmark.realtime do
+        all_offenses = files.flat_map do |path|
+          @run_context.file_processor.call(path).tap { |offenses| mark_progress(offenses) }
+        end
+      end
+
+      status = !reference_lister.stale_violations?
+      msg = status ? "No stale violations detected" : "There were stale violations produced, please run packwerk update"
```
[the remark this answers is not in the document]
cindygshopify: Sorry, what do you mean by the Packwerk update command? Did you just want backticks for the output message about stale violations?

lib/packwerk/cli.rb
exterm: I think it makes sense for the `CLI` class to have one method per packwerk subcommand, however we can consider reducing the size of this method considerably by moving most of its implementation somewhere else.

IMO, `CLI` should just be a thin wrapper that does the CLI specific stuff (like managing the output stream and parsing parameters) and delegates to other classes for everything else.

lib/packwerk/cli.rb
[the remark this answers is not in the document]
exterm: I think this is what Maple means:

```suggestion
      msg = status ? "No stale violations detected" : "There were stale violations found, please run `packwerk update`"
```

lib/packwerk/cache_deprecated_references.rb
```diff
@@ -0,0 +1,45 @@
...
+
+require "packwerk/deprecated_references"
+require "packwerk/reference"
+require "packwerk/reference_lister"
+require "packwerk/violation_type"
+
+module Packwerk
+  class CacheDeprecatedReferences
+    extend T::Sig
+    include ReferenceLister
```
exterm, line 14: I think we can actually make this class abstract - it should never be instantiated, just its child classes, right?

We could to that by using sorbet's `abstract!`, which [can be used to prevent a class from being instantiated.](https://sorbet.org/docs/abstract).

lib/packwerk/cli.rb
tomstuart: I agree that `CLI` is a mess and should be reorganised but am more ambivalent about whether it necessarily must happen in this PR. Could one of us open a new PR to do the cleanup, either before or after this one lands?

exterm: I agree with Tom - Actually that was what I had in mind but failed to address the point. I meant to say "what should happen to CLI is a larger refactor and should thus be done separately"

cindygshopify:
I've pushed one more commit that moves the detect_stale_violations function that was in cli.rb into it's own class so we can get some separation of concerns in the Packwerk codebase. This required some refactoring of cli.rb as well by putting some common functions into a module. I've also updated the code to reflect some of the above mentioned comments, so please take a look. In the interest of time, I'd prefer not to drag out this PR more if it is related to further refactoring. I think this is better served by creating new GitHub issues around that so we can avoid scope creep.

lib/packwerk/cache_deprecated_references.rb
cindygshopify: I attempted to do this as it seemed like it would be low hanging fruit we could ship with this PR, but these signatures generated some really weird typecheck errors that should be further investigated in another PR. For instance, in `deprecated_references_for` package param seems to be both a String and Package. It's a little weird and I really think necessitates further investigation.

cindygshopify:
I've pushed one more commit that moves the detect_stale_violations function that was in cli.rb into it's own class so we can get some separation of concerns in the Packwerk codebase. This required some refactoring of cli.rb as well by putting some common functions into a module. I've also updated the code to reflect some of the above mentioned comments, so please take a look. There are some sorbet signatures I have not added as it negatively affects the tests that I've written. The solution to that requires even more refactoring in my opinion. In the interest of time, I'd prefer not to drag out this PR more if it is related to further refactoring. I think this is better served by creating new GitHub issues around that so we can avoid scope creep.

lib/packwerk/cache_deprecated_references.rb
cindygshopify: done

tomstuart:
Thank you Cindy, the code looks good to me. My laptop is (very slowly) tophatting it now, then I’ll approve 👍

tomstuart:
`packwerk update` works fine for this branch, so that looks good.

I’m sorry but I couldn’t successfully tophat `dev packages detect-stale-violations` in Shopify/shopify. When I tried to run it I got the error:

```
lib/packwerk/cli.rb:192:in `detect_stale_violations': uninitialized constant Packwerk::Cli::DetectStaleViolationsCommand (NameError)
```

I can fix that by putting `require "packwerk/commands/detect_stale_violations_command"` at the top of `lib/packwerk/cli.rb`, but then I get a different error:

```
lib/packwerk/commands/detect_stale_violations_command.rb:33:in `block (2 levels) in run': private method `file_processor' called for #<Packwerk::RunContext> (NoMethodError)
```

I could fix _that_ by making `RunContext#file_processor` public, but rather than continue chasing these I thought I’d report back and let you know it wasn’t working right. Was it working fine before? What’s changed?

cindygshopify:
> Was it working fine before? What’s changed?

It was. I tested before all the refactoring shenanigans so that might've broken something. I figured it would have been covered by the test cases I wrote, but it seems like it wasn't. I'll take another look

cindygshopify:
There was a minor change in another file that broke the functionality of detect-stale-violations, I just pushed with an amended change. Tophatting should work now.

wildmaples:
> There was a minor change in another file that broke the functionality of detect-stale-violations, I just pushed with an amended change. Tophatting should work now.

What was the issue and how did you solve it?

lib/packwerk/cli.rb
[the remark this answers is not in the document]
wildmaples: https://github.com/Shopify/packwerk/pull/61/files/fe49aee8f6b023360c806093f6b05dc6e2ddd6b4#diff-ac5ae86965a90253f8d378798aa32f4041ec009d6b79c4420e88be6e67108e88R133 - I think this should work.

Within `#update_deprecations` method a little above in this file, it goes on to define a variable called `updating_deprecated_references`, which is a `reference_lister`. 

Reference lister is the same type as the class you are creating in this line, so the suggestion was to keep things consistent and readable in the other method by changing the variable name to reference_lister too. (It's a suggestion though, feel free to ignore!)

cindygshopify:
> > There was a minor change in another file that broke the functionality of detect-stale-violations, I just pushed with an amended change. Tophatting should work now.
> 
> What was the issue and how did you solve it?

RunContext#file_processor was a public function that was being used in DetectStaleViolationsCommand. However due to a refactor outside of this PR, it got changed to be a private function. However since this function was previously being used outside of it's own class, another function was written to replace the functionality of the function calls outside its' class. Thus I replaced RunContext#file_processor with RunContext#process_file

lib/packwerk/cli.rb
wildmaples: Can you add an issue for this refactor some of the CLI logic out of the CLI so we don't forget?

wildmaples: Never mind, I just did it! https://github.com/Shopify/packwerk/issues/70

wildmaples (approved):
🎩 Tophatted and it works!!! 🎉 

This is a huge (and much needed) change, but I would prefer to include it in release `v1.0.3` instead of `v1.0.2`. 

Reason is because we are currently fixing a change for `v1.0.2` and if this change is merged and for some unexpected reason breaks, we'd have two back-to-back broken versions.

cindygshopify:
> Reason is because we are currently fixing a change for `v1.0.2` and if this change is merged and for some unexpected reason breaks, we'd have two back-to-back broken versions.

Okay. Will we need to hold off merging this for a while then or would that not matter and just be something to do noted by the release manager when doing the releases?

lib/packwerk/cli.rb
```diff
@@ -181,21 +188,25 @@ def check(paths)
...
     end
 
-    def mark_progress(offenses)
-      if offenses.empty?
-        @progress_formatter.mark_as_inspected
-      else
-        @progress_formatter.mark_as_failed
-      end
-    end
-
```
doodzik, line 198: 🙌

doodzik (approved):
LGTM

exterm:
Is anyone following up by using this to fix the stale violations check that is already in core CI?

cindygshopify:
@exterm I'm going that

PR #63, opened 2020-11-05
add node processor factory

doodzik:

When I was reading through the code, I got a bit confused about the `RunContext` class's responsibility. The particular issue I saw was that that class is used to build a NodeProcessor and a FileProcessor. This construction is a bit backward because the FileProcessor gets initialized with a RunContext, but is used in the CLI class.

## What approach did you choose, and why?

I did an initial stab at this problem in #44  but gave the CLI class too many responsibilities.
I paired with @exterm on this problem, but we approached it from another side. We decided to call the FillProcessor#call method from the RunContext class instead of exposing the FileProcessor class to the CLI class. 

After that, I changed the code to pass a NodeProcessorFactory to the
FileProcessor instead of a RunContext instance, which reduces how complected these classes are. 

There are some other minor cleanups in the PR, but those are described in the commit messages.

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist

### DIFF
--- a/lib/packwerk/cli.rb
+++ b/lib/packwerk/cli.rb
@@ -138,7 +138,7 @@ module Packwerk
       all_offenses = T.let([], T.untyped)
       execution_time = Benchmark.realtime do
         all_offenses = files.flat_map do |path|
-          @run_context.file_processor.call(path).tap { |offenses| mark_progress(offenses) }
+          @run_context.process_file(file: path).tap { |offenses| mark_progress(offenses) }
         end
 
         updating_deprecated_references.dump_deprecated_references_files
@@ -160,7 +160,7 @@ module Packwerk
       all_offenses = T.let([], T.untyped)
       execution_time = Benchmark.realtime do
         files.each do |path|
-          @run_context.file_processor.call(path).tap do |offenses|
+          @run_context.process_file(file: path).tap do |offenses|
             mark_progress(offenses)
             all_offenses.concat(offenses)
           end
--- a/lib/packwerk/cli.rb
+++ b/lib/packwerk/cli.rb
@@ -11,6 +11,7 @@ require "packwerk/inflector"
 require "packwerk/output_styles"
 require "packwerk/run_context"
 require "packwerk/updating_deprecated_references"
+require "packwerk/checking_deprecated_references"
 
 module Packwerk
   class Cli
@@ -21,7 +22,10 @@ module Packwerk
       @err_out = err_out
       @style = style
       @configuration = configuration || Configuration.from_path
-      @run_context = run_context || Packwerk::RunContext.from_configuration(@configuration)
+      @run_context = run_context || Packwerk::RunContext.from_configuration(
+        @configuration,
+        reference_lister: ::Packwerk::CheckingDeprecatedReferences.new(@configuration.root_path),
+      )
       @progress_formatter = Formatters::ProgressFormatter.new(@out, style: style)
     end
 
--- a/lib/packwerk/file_processor.rb
+++ b/lib/packwerk/file_processor.rb
@@ -15,8 +15,8 @@ module Packwerk
       end
     end
 
-    def initialize(run_context:, parser_factory: nil)
-      @run_context = run_context
+    def initialize(node_processor_factory:, parser_factory: nil)
+      @node_processor_factory = node_processor_factory
       @parser_factory = parser_factory || Packwerk::Parsers::Factory.instance
     end
 
@@ -32,8 +32,8 @@ module Packwerk
 
       result = []
       if node
-        @node_processor = @run_context.node_processor_for(filename: file_path, ast_node: node)
-        node_visitor = Packwerk::NodeVisitor.new(node_processor: @node_processor)
+        node_processor = @node_processor_factory.for(filename: file_path, node: node)
+        node_visitor = Packwerk::NodeVisitor.new(node_processor: node_processor)
 
         node_visitor.visit(node, ancestors: [], result: result)
       end
--- a/lib/packwerk/node_processor_factory.rb
+++ b/lib/packwerk/node_processor_factory.rb
@@ -1,0 +1,40 @@
+# typed: strict
+# frozen_string_literal: true
+
+require "packwerk/constant_name_inspector"
+require "packwerk/checker"
+
+module Packwerk
+  class NodeProcessorFactory < T::Struct
+    extend T::Sig
+
+    const :root_path, String
+    const :reference_lister, ReferenceLister
+    const :node_processor_class, T.class_of(NodeProcessor)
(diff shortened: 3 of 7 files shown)


rochlefebvre (approved):
The changes make sense, according to my limited experience with Packwerk.  The new methods in RunContext clean up the initializer quite nicely. I suggest you wait for a second review from someone with more Packwerk domain knowledge.

tomstuart (commented):
I’ve run the `node_processor_factory` branch against the main Shopify monolith and confirmed that its `packwerk check` and `packwerk update` behaviour is the same as on its current base commit (dac7b105). There are merge conflicts against `main` so I haven’t yet tried this with a merged/rebased version.

lib/packwerk/run_context.rb
```diff
@@ -14,15 +14,17 @@
 require "packwerk/package_set"
 require "packwerk/privacy_checker"
 require "packwerk/reference_extractor"
+require "packwerk/node_processor_factory"
```
tomstuart, line 16: I assume this `require` belongs in a later commit, since this file doesn’t exist (yet).

lib/packwerk/file_processor.rb
```diff
@@ -32,8 +32,8 @@ def call(file_path)
 
       result = []
       if node
-        @node_processor = @run_context.node_processor_for(filename: file_path, ast_node: node)
```
tomstuart, line 35: It’s interesting that we had an unnecessary instance variable here. Looking at the history, it was originally used by a `#visit` method; that method got removed when we extracted `Packwerk::NodeVisitor` but the instance variable was left behind.

lib/packwerk/run_context.rb
```diff
@@ -36,15 +36,13 @@ class RunContext
 
     class << self
       def from_configuration(configuration, reference_lister: nil)
-        default_reference_lister = reference_lister ||
-          ::Packwerk::CheckingDeprecatedReferences.new(configuration.root_path)
```
tomstuart, line 38: Good catch!

lib/packwerk/run_context.rb
```diff
@@ -57,7 +57,7 @@ def initialize(
       node_processor_class: NodeProcessor,
       reference_lister: nil
     )
-      @root_path = File.expand_path(root_path)
+      @root_path = root_path
```
tomstuart, line 59: It took me a little while to understand why this is okay, but I think it’s because a) the only place where `RunContext` gets constructed is in `RunContext.from_configuration`, which uses `Configuration#root_path`, and b) `Configuration` already calls `File.expand_path` in its constructor so `Configuration#root_path` has always been expanded.

lib/packwerk/run_context.rb
```diff
@@ -22,8 +22,13 @@ class RunContext
...
     attr_reader(
       :root_path,
+      :load_paths,
+      :package_paths,
+      :inflector,
+      :custom_associations,
+      :checker_classes,
       :node_processor_class,
-      :reference_lister
+      :reference_lister,
```
tomstuart, line 29: 👀

lib/packwerk/run_context.rb
```diff
@@ -119,7 +124,7 @@ def checkers
     def constant_name_inspectors
       [
         ::Packwerk::ConstNodeInspector.new,
-        ::Packwerk::AssociationInspector.new(inflector: @inflector, custom_associations: @custom_associations),
+        ::Packwerk::AssociationInspector.new(inflector: inflector, custom_associations: custom_associations),
```
tomstuart, line 122: I agree we should do this, but there are still references to `@load_paths`, `@package_paths`, `@inflector` and `@checker_classes` throughout the class — should we clean those up too?

lib/packwerk/node_processor_factory.rb
```diff
@@ -10,14 +10,13 @@ class NodeProcessorFactory < T::Struct
...
     const :reference_lister, ReferenceLister
-    const :node_processor_class, T.class_of(NodeProcessor)
     const :context_provider, Packwerk::ConstantDiscovery
     const :constant_name_inspectors, T::Array[ConstantNameInspector]
     const :checkers, T::Array[Checker]
 
     sig { params(filename: String, node: AST::Node).returns(NodeProcessor) }
     def for(filename:, node:)
-      node_processor_class.new(
+      ::Packwerk::NodeProcessor.new(
```
tomstuart, line 19: This is good I think. I find it easier to understand this way, and if we’re not losing any capability then that seems like a win. As always, we can easily switch back to injecting the class if/when we ever need it.

lib/packwerk/cli.rb
```diff
@@ -21,7 +21,10 @@ def initialize(run_context: nil, configuration: nil, out: $stdout, err_out: $std
       @err_out = err_out
       @style = style
       @configuration = configuration || Configuration.from_path
-      @run_context = run_context || Packwerk::RunContext.from_configuration(@configuration)
+      @run_context = run_context || Packwerk::RunContext.from_configuration(
+        @configuration,
+        reference_lister: ::Packwerk::CheckingDeprecatedReferences.new(@configuration.root_path),
```
tomstuart, line 26: Nice! I think we should add a `require "packwerk/checking_deprecated_references"` in this file now.

lib/packwerk/run_context.rb
```diff
@@ -54,15 +54,15 @@ def initialize(
...
-      reference_lister: nil
+      reference_lister:
     )
       @root_path = root_path
       @load_paths = load_paths
       @package_paths = package_paths
       @inflector = inflector
       @custom_associations = custom_associations
       @checker_classes = checker_classes
-      @reference_lister = reference_lister || ::Packwerk::CheckingDeprecatedReferences.new(@root_path)
```
tomstuart: I think you can remove `require "packwerk/checking_deprecated_references"` now.

tomstuart (approved):
This looks good, thank you @doodzik! I left a few suggestions but didn’t see any blockers.

lib/packwerk/run_context.rb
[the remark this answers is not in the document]
doodzik: Nice catch, I probably messed up the `git commit -p`.

test/unit/cli_test.rb
```diff
@@ -52,18 +45,12 @@ class CliTest < Minitest::Test
       violation_message = "This is a violation of code health."
       offense = stub(to_s: violation_message)
 
-      file_processor = stub
-      file_processor
-        .stubs(:call)
+      run_context = stub
+      run_context.stubs(:process_file)
```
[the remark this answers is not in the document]
doodzik: I changed it to `at_least(2)` since that is more accurate. Good catch 👍

doodzik:
I fixed the merge conflict and ran the new version of packwerk on core.

PR #64, opened 2020-11-05
exclude static directory from gem

mclark:
The packwerk 1.0.1 gem file is 4,217,344 bytes in size. This seems rather large and upon further investigation appears to be due to the inclusion of a number of graphics files stored in the `static` subdirectory. Excluding these images resulted in shrinking the packwerk gem down to 634,368 bytes.

I chose to just add the `static` directory to the rejection list when computing the files for the gemspec. This was done because it is a small, safe change that is easy to read and understand.

Are these images actually needed for something in the runtime environment? Am I overlooking some longstanding gem convention of which I am unaware?

## Type of Change

- [ ] Bugfix
- [ ] New feature

## Checklist

### DIFF
--- a/packwerk.gemspec
+++ b/packwerk.gemspec
@@ -34,7 +34,7 @@ Gem::Specification.new do |spec|
   spec.executables << "packwerk"
 
   spec.files = Dir.chdir(__dir__) do
-    %x(git ls-files -z).split("\x0").reject { |f| f.match(%r{^(test|spec|features)/}) }
+    %x(git ls-files -z).split("\x0").reject { |f| f.match(%r{^(test|spec|features|static)/}) }
   end
   spec.require_paths = %w(lib)
 


mclark:
I am checking with my manager to see how to handle the CLA requirement and will update here when I know more.

mclark:
👋 I have signed an individual CLA, not sure the expected next step. Shall I just rewrite my commit and force push, or is there a manual step on Shopify's end to re-trigger the validation?

rochlefebvre:
> 👋 I have signed an individual CLA, not sure the expected next step. Shall I just rewrite my commit and force push, or is there a manual step on Shopify's end to re-trigger the validation?

Hi Matt! I re-ran the CLA check under the Checks tab, and it ✅ 'ed out!

exterm (approved):
Great catch!

wildmaples (approved):
Ah nice! The static folder contains gifs to supplement our documentation and does not need to be included with the gem.

PR #65, opened 2020-11-10
Add missing .realpath checks that was causing validation error

wildmaples:

There are two bugs in `v1.0.1` that results in `packwerk validate` failure because the paths were not real paths as it was [recently removed](https://github.com/Shopify/packwerk/pull/38/files#diff-5ca9dcd54fd0b9eeef94e00ba6adc7822dc83e5dd2bfb9e8d96fd3c356de8b8aR328). Because there was no test to catch it, this ended up in a release. 

There are two fixes in this PR. I have linked the appropriate PRs / line change that caused the bug. 

We should get this change out ASAP. 

You can help by confirming this change with the steps below: 

1. checkout a master branch of Core with a local version of Packwerk v1.0.1 linked 
2. run `bin/spring packwerk validate` - this will fail
3. checkout this branch and run validate again - this should pass

## Type of Change

- [x] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist

### DIFF
--- a/lib/packwerk/application_load_paths.rb
+++ b/lib/packwerk/application_load_paths.rb
@@ -48,7 +48,9 @@ module Packwerk
 
       sig { params(paths: T::Array[Pathname], rails_root: Pathname).returns(T::Array[String]) }
       def relative_path_strings(paths, rails_root: Rails.root)
-        paths.map { |path| path.relative_path_from(rails_root).to_s }
+        paths
+          .map { |path| path.relative_path_from(rails_root).to_s }
+          .uniq
       end
 
       sig { params(paths: T::Array[T.untyped]).void }
--- a/lib/packwerk/application_validator.rb
+++ b/lib/packwerk/application_validator.rb
@@ -329,7 +329,8 @@ module Packwerk
     end
 
     def package_manifests(glob_pattern)
-      PackageSet.package_paths(@configuration.root_path, glob_pattern).map(&:to_s)
+      PackageSet.package_paths(@configuration.root_path, glob_pattern)
+        .map { |f| File.realpath(f) }
     end
 
     def relative_paths(paths)
--- a/lib/packwerk/package_set.rb
+++ b/lib/packwerk/package_set.rb
@@ -29,7 +29,8 @@ module Packwerk
         bundle_path_match = Bundler.bundle_path.join("**").to_s
 
         Dir.glob(File.join(root_path, package_pathspec, PACKAGE_CONFIG_FILENAME))
-          .map { |path| Pathname.new(path) }.reject { |path| path.realpath.fnmatch(bundle_path_match) }
+          .map { |path| Pathname.new(path) }
+          .reject { |path| path.realpath.fnmatch(bundle_path_match) }
       end
 
       private
--- a/test/unit/application_load_paths_test.rb
+++ b/test/unit/application_load_paths_test.rb
@@ -46,5 +46,12 @@ module Packwerk
       ApplicationLoadPaths.expects(:filter_relevant_paths).once.returns([Pathname.new("/fake_path")])
       ApplicationLoadPaths.extract_relevant_paths
     end
+
+    test ".extract_relevant_paths returns unique load paths" do
+      path = Pathname.new("/application/app/models")
+      ApplicationLoadPaths.expects(:filter_relevant_paths).once.returns([path, path])
+
+      assert_equal 1, ApplicationLoadPaths.extract_relevant_paths.count
+    end
   end
 end


lib/packwerk/configuration.rb
```diff
@@ -61,10 +61,12 @@ def all_application_autoload_paths
       bundle_path_match = Bundler.bundle_path.join("**").to_s
 
       all_paths = all_paths.each_with_object([]) do |path_string, paths|
-        # ignore paths inside gems
         path = Pathname.new(path_string)
 
-        next unless path.exist?
+        # ignore paths outside of the Rails root
+        next unless path.exist? && path.realpath.fnmatch(Rails.root.join("**").to_s)
```
wildmaples, line 67: https://github.com/Shopify/packwerk/pull/21/files#diff-e43184d1a960a3c3821da21b3eb72c83379c2dcc996f6bcfac9aadd6898cd374L64

lib/packwerk/application_validator.rb
```diff
@@ -329,7 +329,8 @@ def package_glob
     end
 
     def package_manifests(glob_pattern)
-      PackageSet.package_paths(@configuration.root_path, glob_pattern).map(&:to_s)
+      PackageSet.package_paths(@configuration.root_path, glob_pattern)
+        .map { |f| File.realpath(f) }
```
wildmaples, line 333: https://github.com/Shopify/packwerk/pull/38/files#diff-5ca9dcd54fd0b9eeef94e00ba6adc7822dc83e5dd2bfb9e8d96fd3c356de8b8aL328

rochlefebvre: I verified that `File.realpath(Pathname)` returns a string, so there's no need to add `to_s` like in the old code.
```ruby
[9] pry(main)> File.realpath(Pathname.new("."))
=> "/Users/rochlefebvre/src/github.com/Shopify/parsnip"
```
👍

lib/packwerk/configuration.rb
rochlefebvre: Nitpick: it feels a bit clearer to me if these two conditionals exist on their own lines:
```ruby
next unless path.exist?
next unless path.realpath.fnmatch(Rails.root.join("**").to_s)
```
`path.exist?` is a precondition for the rest of the loop, not just for your new Rails path condition.

rochlefebvre: Can we test this new condition in configuration_test?  Maybe by appending an external path to `Rails.application.config.autoload_paths` beforehand?  Don't forget to restore the paths to their old value.

exterm (approved):
If we can add a test, we can merge this. Thank you, Maple!

wildmaples:
> If we can add a test, we can merge this. Thank you, Maple!

I know this is not optimal, but I tried testing and it is hard. Two changes in this PR: 
- For the validator: we have an [existing issue for `ApplicationValidator` unit tests](https://github.com/Shopify/packwerk/issues/17) and I believe @rochlefebvre is working on it. 
- The change within `Configuration#all_application_autoload_paths` is also quite hard to test without a huge refactor. We could move the Rails logic out of the method/class so we can make it easier to test but unfortunately that is without a larger refactor right now. 

I don't want to avoid testing. Can you see any easy test suite I can add to account for either these two cases? Otherwise I vote to ship this change/release and work on the test in another PR.

exterm:
The way I see it, `all_application_autoload_paths` does three things:

1. get all autoload paths out of rails
2. filter the paths to just the ones that are interesting to us
3. warn if no paths remain

The part that broke was `2.`. If you can extract these three steps into separate methods, you should be able to easily test just step `2`. It would just be a method that takes a list of paths as well as an application root path and a bundle path, and returns a list of paths.

rochlefebvre (approved):

lib/packwerk/configuration.rb
wildmaples: > Can we test this new condition in configuration_test? Maybe by appending an external path to Rails.application.config.autoload_paths beforehand? Don't forget to restore the paths to their old value.

Tested in https://github.com/Shopify/packwerk/pull/68 where this logic was moved into its own class.

lib/packwerk/application_load_paths.rb
```diff
@@ -48,7 +48,9 @@ def filter_relevant_paths(all_paths, bundle_path: Bundler.bundle_path, rails_roo
 
       sig { params(paths: T::Array[Pathname], rails_root: Pathname).returns(T::Array[String]) }
       def relative_path_strings(paths, rails_root: Rails.root)
-        paths.map { |path| path.relative_path_from(rails_root).to_s }
+        paths
+          .map { |path| path.relative_path_from(rails_root).to_s }
+          .uniq
```
wildmaples, line 53: We need this `.uniq` because `#filter_relevant_paths` may result in some added load paths. This was causing a failure when I tophatted on Core.

wildmaples: The `.uniq` was removed from `ApplicationValidator`, but I figure it is best to add it in this class so we can also test for it.

lib/packwerk/application_validator.rb
doodzik: \nit I'm thinking out loud here, but could it make sense to add a method to the PackageSet that returns all the realpaths for the given packages instead of ApplicationValidator being aware of retrieving that information? 🤔

doodzik (approved):

tomstuart (approved):
The code change looks correct to me. I tophatted it on core and it works. Thank you!

tonymarklove-babylon:
Apologies for the breakage, everyone! I was just looking at the latest version and spotted that I managed to break an entire release. 🤦 It worked on our codebase!

PR #68, opened 2020-11-11
Make load path extraction logic testable

wildmaples:

Pre-requisite refactor for #65

We want to move the Rails loading logic out of Configuration and into `ApplicationValidator`, which is the only place we need to load/require files from Rails. 

More specifically, we are breaking down `Configuration#all_application_autoload_paths` and moving it to `ApplicationLoadPaths`, which can be extracted within `ApplicationValidator`.  

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist

### DIFF
--- a/lib/packwerk/application_load_paths.rb
+++ b/lib/packwerk/application_load_paths.rb
@@ -1,0 +1,65 @@
+# typed: strict
+# frozen_string_literal: true
+
+require "bundler"
+
+module Packwerk
+  module ApplicationLoadPaths
+    class << self
+      extend T::Sig
+
+      sig { returns(T::Array[String]) }
+      def extract_relevant_paths
+        assert_application_booted
+        all_paths = extract_application_autoload_paths
+        relevant_paths = filter_relevant_paths(all_paths)
+        assert_load_paths_present(relevant_paths)
+        relative_path_strings(relevant_paths)
+      end
+
+      sig { void }
+      def assert_application_booted
+        raise "The application needs to be booted to extract load paths" unless defined?(::Rails)
+      end
+
+      sig { returns(T::Array[String]) }
+      def extract_application_autoload_paths
+        Rails.application.railties
+          .select { |railtie| railtie.is_a?(Rails::Engine) }
+          .push(Rails.application)
+          .flat_map do |engine|
+          (engine.config.autoload_paths + engine.config.eager_load_paths + engine.config.autoload_once_paths).uniq
+        end
+      end
+
+      sig do
+        params(all_paths: T::Array[String], bundle_path: Pathname, rails_root: Pathname)
+          .returns(T::Array[Pathname])
+      end
+      def filter_relevant_paths(all_paths, bundle_path: Bundler.bundle_path, rails_root: Rails.root)
+        bundle_path_match = bundle_path.join("**")
+        rails_root_match = rails_root.join("**")
+
+        all_paths
+          .map { |path| Pathname.new(path).expand_path }
+          .select { |path| path.fnmatch(rails_root_match.to_s) } # path needs to be in application directory
+          .reject { |path| path.fnmatch(bundle_path_match.to_s) } # reject paths from vendored gems
+      end
+
+      sig { params(paths: T::Array[Pathname], rails_root: Pathname).returns(T::Array[String]) }
+      def relative_path_strings(paths, rails_root: Rails.root)
+        paths.map { |path| path.relative_path_from(rails_root).to_s }
+      end
+
+      sig { params(paths: T::Array[T.untyped]).void }
+      def assert_load_paths_present(paths)
+        if paths.empty?
+          raise <<~EOS
+            We could not extract autoload paths from your Rails app. This is likely a configuration error.
+            Packwerk will not work correctly without any autoload paths.
+          EOS
+        end
+      end
+    end
+  end
+end
--- a/lib/packwerk/configuration.rb
+++ b/lib/packwerk/configuration.rb
@@ -42,45 +42,10 @@ module Packwerk
       @root_path = File.expand_path(root)
       @package_paths = configs["package_paths"] || "**/"
       @custom_associations = configs["custom_associations"] || []
-      @load_paths = configs["load_paths"] || all_application_autoload_paths
+      @load_paths = configs["load_paths"]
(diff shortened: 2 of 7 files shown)


exterm:
How does this make things easier to test?

exterm:
> How does this make things easier to test?

Answer: It wasn't finished.

test/unit/configuration_test.rb
```diff
@@ -2,7 +2,6 @@
 # frozen_string_literal: true
 
 require "test_helper"
-require "rails_test_helper"
```
exterm, line 5: 🎉 🎉 🎉 🎉 🎉

exterm:
I will squash the commits together before merging, but thought I'd keep them for now to ease reviews.

lib/packwerk/application_load_paths.rb
```diff
@@ -0,0 +1,65 @@
...
+      sig { returns(T::Array[String]) }
+      def extract_relevant_paths
+        assert_application_booted
+        all_paths = extract_application_autoload_paths
+        relevant_paths = filter_relevant_paths(all_paths)
+        assert_load_paths_present(relevant_paths)
+        relative_path_strings(relevant_paths)
+      end
+
+      sig { void }
```
wildmaples, line 20: Add these as private methods?

test/unit/configuration_test.rb
```diff
@@ -49,35 +48,9 @@ class ConfigurationTest < Minitest::Test
...
-
-      configuration = Configuration.from_path
-      expected_autoloads = default_autoloads - ["vendor/cache/gems/example/models"]
-
-      assert_equal expected_autoloads, configuration.load_paths.sort
-    end
-
-    private
-
-    # TODO: this isn't great, so let's come back and refactor so that we're in more control
```
wildmaples, line 71: YAY

test/unit/application_load_paths_test.rb
```diff
@@ -0,0 +1,50 @@
+# typed: false
+# frozen_string_literal: true
+
+require "test_helper"
+require "rails_test_helper"
+
+module Packwerk
+  class ApplicationLoadPathsTest < Minitest::Test
```
wildmaples, line 8: Should we add a test case for paths that do not exist? Which is the issue we are seeing in https://github.com/Shopify/packwerk/pull/65

wildmaples (commented):
I had a comment about that one test case, but aside from that this is a pleasing change. 

I can't approve my own PR, so here is a comment-approval 👍 💚

rochlefebvre (approved):
Aside from @wildmaples questions, looks good to me! This change will definitely help me simplify my test fixture stuff!

test/unit/application_load_paths_test.rb
exterm: Definitely! I had actually thought about that at some point and then forgot about it 🤦

exterm: Thank you Maple!

wildmaples:
OK -- I updated the PR to more accurately reflect this change. Once you are done with this, @exterm . I will rebase #65 and create a new release.

lib/packwerk/application_load_paths.rb
exterm: At least the filter method needs to be public so I can test it 🤷 

And I would love to list them in the order they're executed here.

Not sure making some of them private is worth it.

Or should I just circumvent privacy in the test? That doesn't seem ideal either.

test/unit/application_load_paths_test.rb
exterm: Mmmh, revisiting this - I don't think non-existent paths were the problem; the problem was that the paths were outside the project directory. Which is tested here.

exterm: I should however add a test for a more complex path, like `/application/../something/`.

exterm: I can't / don't want to use `.realpath` because that accesses the file system, which makes testing hard.

lib/packwerk/application_load_paths.rb
wildmaples: I think this is good.

test/unit/application_load_paths_test.rb
wildmaples: > Mmmh, revisiting this - I don't think non-existent paths were the problem; the problem was that the paths were outside the project directory. Which is tested here.

Yes, sorry for the confusion, that was what I meant

exterm: Alignment!

exterm:
Ready for re-review @wildmaples

tomstuart:
I’m going to try to review this, but am finding it difficult to understand now that it’s been squashed into a single commit. Why (in future) can’t we keep PRs as multiple commits?

tomstuart (approved):
I can’t fully follow what’s changing here, but the tests still pass and the new code looks better than the old code, so… 😬👍

lib/packwerk/configuration.rb
```diff
@@ -42,45 +42,10 @@ def initialize(configs = {}, config_path: nil)
...
       @package_paths = configs["package_paths"] || "**/"
       @custom_associations = configs["custom_associations"] || []
-      @load_paths = configs["load_paths"] || all_application_autoload_paths
+      @load_paths = configs["load_paths"]
       @inflections_file = File.expand_path(configs["inflections_file"] || "config/inflections.yml", @root_path)
 
       @config_path = config_path
     end
-
-    def all_application_autoload_paths
```
wildmaples, line 51: @tomstuart the gist of it is that this method is getting moved out of `Configuration` and into a new file `ApplicationLoadPaths` because it has logic to load Rails files. This method is now testable in its new home.

exterm:
Oh, I'm Sorry @tomstuart . The commits were very small so I had the choice to squash into three more meaningful ones or one, and I went for the extreme. Will err towards more commits next time.

tomstuart:
No worries, thanks @exterm!

lib/packwerk/application_load_paths.rb
```diff
@@ -0,0 +1,65 @@
+# typed: strict
```
doodzik, line 1: 🙌

PR #71, opened 2020-11-16
Support array of paths in PackageSet

rochlefebvre:
Packwerk integrators can set the scope of directories to search for package definitions by providing a `package_paths` in their packwerk.yml file.  The default value is `"**"`, which corresponds to the application's entire set of directories and subdirectories.

Although `package_paths` sounds likes one can specify more than one path (like in the `include`, `exclude`, and `load_paths` settings), this is not currently the case:
https://github.com/Shopify/packwerk/blob/6d89101f54a6d415938e852cae06f11833ccda32/lib/packwerk/package_set.rb#L28-L33
If someone were to configure `packwerk.yml` like so
```yml
# packyerk.yml
package_paths:
  - .
  - package/a/**
  - package/b
```
The they would wind up with
```ruby
# Dir.glob(File.join(root_path, package_pathspec, PACKAGE_CONFIG_FILENAME)) ==>
Dir.glob("/.../Shopify/packwerk/./package/a/**/package/b/package.yml") # oops
```
It's possible to describe the example package paths above with a single glob pattern
```yml
# packyerk.yml
package_paths: {.,package/a/**,package/b}
```
Nonetheless, an array format for listing paths is less surprising to me.

Support both array and string values in `package_paths` by using [`Array(the_value)`](https://ruby-doc.org/core-2.7.0/Kernel.html#method-i-Array), which ensures that everything going forward is an array.

All of it.

## Type of Change

- [x] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist

### DIFF
--- a/lib/packwerk/application_validator.rb
+++ b/lib/packwerk/application_validator.rb
@@ -32,7 +32,7 @@ module Packwerk
         check_acyclic_graph,
         check_package_manifest_paths,
         check_valid_package_dependencies,
-        check_root_package_exist,
+        check_root_package_exists,
       ]
 
       results.reject!(&:ok?)
@@ -297,7 +297,7 @@ module Packwerk
       end
     end
 
-    def check_root_package_exist
+    def check_root_package_exists
       root_package_path = File.join(@configuration.root_path, "package.yml")
       all_packages_manifests = package_manifests(package_glob)
 
--- a/lib/packwerk/application_validator.rb
+++ b/lib/packwerk/application_validator.rb
@@ -207,7 +207,7 @@ module Packwerk
     end
 
     def check_acyclic_graph
-      packages = Packwerk::PackageSet.load_all_from(".")
+      packages = Packwerk::PackageSet.load_all_from(@configuration.root_path)
 
       edges = packages.flat_map do |package|
         package.dependencies.map { |dependency| [package, packages.fetch(dependency)] }
--- a/lib/packwerk/package_set.rb
+++ b/lib/packwerk/package_set.rb
@@ -28,8 +28,12 @@ module Packwerk
       def package_paths(root_path, package_pathspec)
         bundle_path_match = Bundler.bundle_path.join("**").to_s
 
-        Dir.glob(File.join(root_path, package_pathspec, PACKAGE_CONFIG_FILENAME))
-          .map { |path| Pathname.new(path) }
+        glob_patterns = Array(package_pathspec).map do |pathspec|
+          File.join(root_path, pathspec, PACKAGE_CONFIG_FILENAME)
+        end
+
+        Dir.glob(glob_patterns)
+          .map { |path| Pathname.new(path).cleanpath }
           .reject { |path| path.realpath.fnmatch(bundle_path_match) }
       end
 
--- a/test/unit/package_set_test.rb
+++ b/test/unit/package_set_test.rb
@@ -32,6 +32,31 @@ module Packwerk
       assert_nil(@package_set.fetch("components/unknown"))
     end
 
+    test ".package_paths supports a path wildcard" do
+      package_paths = PackageSet.package_paths("test/fixtures/skeleton", "**")
+
+      assert_includes(package_paths, Pathname.new("test/fixtures/skeleton/components/sales/package.yml"))
+      assert_includes(package_paths, Pathname.new("test/fixtures/skeleton/package.yml"))
+    end
+
+    test ".package_paths supports a single path as a string" do
+      package_paths = PackageSet.package_paths("test/fixtures/skeleton", "components/sales")
+
+      assert_equal(package_paths, [Pathname.new("test/fixtures/skeleton/components/sales/package.yml")])
+    end
+
+    test ".package_paths supports many paths as an array" do
+      package_paths = PackageSet.package_paths("test/fixtures/skeleton", ["components/sales", "."])
+
+      assert_equal(
+        package_paths,
+        [
+          Pathname.new("test/fixtures/skeleton/components/sales/package.yml"),
+          Pathname.new("test/fixtures/skeleton/package.yml"),
+        ]
+      )
+    end
+
     test ".package_paths excludes paths inside the gem directory" do
       vendor_package_path = Pathname.new("test/fixtures/skeleton/vendor/cache/gems/example/package.yml")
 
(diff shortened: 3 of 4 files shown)


lib/packwerk/application_validator.rb
```diff
@@ -207,7 +207,7 @@ def check_inflection_file
     end
 
     def check_acyclic_graph
-      packages = Packwerk::PackageSet.load_all_from(".")
+      packages = Packwerk::PackageSet.load_all_from(@configuration.root_path)
```
rochlefebvre, line 210: We need this fix if we want to use temporary application directories in our tests. See https://github.com/Shopify/packwerk/pull/59

lib/packwerk/package_set.rb
```diff
@@ -28,8 +28,12 @@ def load_all_from(root_path, package_pathspec: nil)
...
         bundle_path_match = Bundler.bundle_path.join("**").to_s
 
-        Dir.glob(File.join(root_path, package_pathspec, PACKAGE_CONFIG_FILENAME))
-          .map { |path| Pathname.new(path) }.reject { |path| path.realpath.fnmatch(bundle_path_match) }
+        glob_patterns = Array(package_pathspec).map do |pathspec|
+          File.join(root_path, pathspec, PACKAGE_CONFIG_FILENAME)
+        end
+
+        Dir.glob(glob_patterns)
+          .map { |path| Pathname.new(path).cleanpath }.reject { |path| path.realpath.fnmatch(bundle_path_match) }
```
rochlefebvre, line 36: `.cleanpath` normalizes the glob pattern to convert things like `"packwerk_app/./app/sales"` into `"packwerk_app/app/sales"`.

lib/packwerk/application_validator.rb
doodzik: Nice! One small thing, can you move the bug fix into a separate commit on this PR? That way we can retain more context on why changes were made and it becomes easier to review the PR.

This is a good talk that explains it in more detail http://yt.dudzik.co/watch/?v=qpdYRPL3SVE

lib/packwerk/package_set.rb
```diff
@@ -28,8 +28,12 @@ def load_all_from(root_path, package_pathspec: nil)
       def package_paths(root_path, package_pathspec)
```
doodzik, line 28: I'm a bit concerned about the number of responsibilities the PackageSet class has. I think think the functionality of the `PackageSet .load_all_from(root_path, package_pathspec: nil)` class method  and by extension the `PackageSet .package_paths(root_path, package_pathspec)` class method should be extracted into a separate class. IMO the functionality should be ideally in a separate class that is responsible for loading the packages from the YAML. We can then call that class from `PackageSet .load_all_from(root_path, package_pathspec: nil)`.

I'm bringing this up because you are changing the Package loading behavior, [so it makes sense to leave it in a better state than you found it](https://www.oreilly.com/library/view/97-things-every/9780596809515/ch08.html).

lib/packwerk/package_set.rb
doodzik: \nit can we multiline the chained methods to make it easier to read?

test/unit/package_set_test.rb
```diff
@@ -32,6 +32,28 @@ class PackageSetTest < Minitest::Test
...
+
+    test ".package_paths supports many paths as an array" do
+      package_paths = PackageSet.package_paths("test/fixtures/skeleton", ["components/sales", "."])
+      assert_equal(
+        package_paths,
+        [
+          Pathname.new("test/fixtures/skeleton/components/sales/package.yml"),
+          Pathname.new("test/fixtures/skeleton/package.yml"),
+        ]
+      )
```
doodzik, line 54: I think it would be good to seperate the setup-action-assertion with whitespace to make the tests easier to read and to make diffing easier.

This approach to testing is also recommended by the ruby shopify style guide:
> Organize the setup, action, and assertion sections of the test case into paragraphs separated by empty lines.

https://github.com/Shopify/ruby-style-guide#testing

lib/packwerk/package_set.rb
rochlefebvre: I'm expanding the package loading behaviour by supporting arrays.  I'm not changing behaviour for existing inputs nor changing the reponsabilies of PackageSet.  Refactoring PackageSet is not going to make my change easier or clearer.  It's not going to affect how other bits of code call PackageSet, either.

Refactoring the class as part of this PR will distract from my small change; I certainly don't want to set this work aside, as it is already a precusory PR for a larger set of changes.

doodzik: > I'm expanding the package loading behavior by supporting arrays.

IMO this constitutes different behavior because it changes how we interact with this class. Furthermore, adding the new `glob_patterns` operation makes the code more complex, not less.

Making classes more complex just a little bit will add up over time and make the code harder to maintain. Since you are adding complexity, you are not leaving this class stagnant nor better. I understand that it might seem silly to take such a hard stance for a minuscule addition to the class, but I think it is important to take these small changes seriously because they add up over time.

> Refactoring PackageSet is not going to make my change easier or clearer. 
> It's not going to affect how other bits of code call PackageSet, either.
> Refactoring PackageSet is not going to make my change easier or clearer. It's not going to affect how other bits of code call PackageSet, either.

It is important to separate classes by making the code easier to understand; otherwise, we could put all code into one class. I already have a hard time understanding what the PackageSet class is intended for, and the proposed change will not make that easier. 

If you want to address this in a separate PR, I'm ok with that because I agree with your assertion that this change would distract from the actual PR, and I don't want to block you on your larger work.

rochlefebvre: I have thought about your arguments and I understand your point of view.  I don't think we're at a point where this incremental, backwards-compatible change warrants a refactor of PackageSet.  The class remains an enumerator for packages that also has a factory-like entrypoint.  I'm not against splitting it up at a later point if the requirements change enough.

wildmaples (approved):
This is a great addition, thanks @rochlefebvre 

Can we modify the table in [this section of the documentation](https://github.com/Shopify/packwerk/blob/main/USAGE.md#setting-up-the-configuration-file) to reflect this PR? Perhaps something like `list or patterns to find package configuration files`? There is also a missing link in `see: Defining packages` 🙈

rochlefebvre:
> this section of the documentation

Oh, good call. I will update the docs in this PR.  I'm thinking of using this description
image

_a single pattern or a list of patterns to find package configuration files_

Does this sound good?

doodzik (approved):

Issue #18, opened 2020-09-25
Change the name `packwerk update` to `packwerk update-deprecations`


wildmaples:
@exterm brought up that `packwerk update` should actually be called `packwerk update-deprecations` and I totally agree. `packwerk update` can be confusing. 

We should add a `packwerk update-deprecations` command and add a deprecation warning for the `packwerk update` command but still allow people to use it.

PR #73, opened 2020-11-17
Rename obsolete "update-deprecations" command in docs

jeromelecomte-shopify:

"update-deprecations" command is no longer supported and seems to have been replaced by "update". Keep the usage documentation up-to-date by replacing occurrences of  "update-deprecations" with "update" (thus matching the Terminal GIF example).

Confirm that "update" is indeed the new "update-deprecations".

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Non-breaking change (documentation)

## Checklist


rochlefebvre (commented):
I believe it's actually the `update` command that was deprecated in favour of `update-deprecations`:
https://github.com/Shopify/packwerk/issues/18

jeromelecomte-shopify:
'update' is the old name, 'update-deprecations' is the new one 🤦

Issue #70, opened 2020-11-13
Refactor cli.rb


wildmaples:
As @doodzik mentioned in https://github.com/Shopify/packwerk/pull/61/files#r517545831, we should look into refactoring `cli.rb` as it is growing to be a large class. We should move all of the different commands we have in to compositional objects, so we can just do `command.run`. This should also make it easier to have unit tests for each class. 

I would also be curious to see how RuboCop implements their Cli class.

doodzik:
@cindygshopify and I did some pairing on this the other day, and I thought we found a nice separation between the CLI and Command class. I think we can apply the same pattern for all the commands. Here is the implementation: https://github.com/Shopify/packwerk/pull/61/files#diff-66788d8e2cafe199807e9963fd538243bcdd6ab0ea3203f07464cb93109e78c0R12-R63

PR #74, opened 2020-11-19
Refactor CLI

santib:
Solves https://github.com/Shopify/packwerk/issues/70 by extracting out each command to a separate object.

Notes:
- I also refactored the structure/names of the files in the commands folder
- I did the changes in independent commits so it's easier to review + easy to rollback any part you don't like

The one suggested in the referenced issue above, because it was previously accepted by maintainers and to keep consistency with existing code.

Make sure I didn't break something by mistake, but apart from that it was just moving some code to other files.

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist


wildmaples:
Hey @santib - It was great chatting with you at RubyConf and I'm very happy to see you making a contribution here.

We'll get someone on our team to look at your PR as soon as possible, perhaps sometime next week. Thanks for understanding!

santib:
Hi @wildmaples it was great to meet you at the conf 😄 I hope to see you speaking at the upcoming RailsConf about packwerk 😉 

Actually I've no hurries to get this PR reviewed, so take the time you need. I had some free time and I took that simple issue to start learning about packwerk and to give back (even if it's with such a small contribution) to this amazing project.

lib/packwerk/cli.rb
```diff
@@ -134,70 +94,40 @@ def update(paths)
...
-      end
-
-      @out.puts # put a new line after the progress dots
-      show_offenses(all_offenses)
-      @progress_formatter.finished(execution_time)
-      @out.puts("✅ `deprecated_references.yml` has been updated.")
-
-      all_offenses.empty?
+      Commands::UpdateDeprecations.new(
+        out: @out,
```
doodzik, line 98: The idea of refactoring the CLI class is to create a thin wrapper around the commands and abstracting some common functionality. [If you have a look at the `DetectStaleViolation` class and you can see that we removed the dependency of `@out` and `@style`](https://github.com/Shopify/packwerk/blob/main/lib/packwerk/cli.rb#L192-L200). Instead, we opted for using a Result object that gets interpreted by the CLI class. This makes the command classes easier to test and reason because it is less complex. 

I would love for this refactor follow the same pattern so that we are consistent 🙂

doodzik (changes requested):
Thank you for looking into refactoring the CLI class. This is great and definitely a needed improvement. 🎉🎉🎉  

[I left one comment, which applies throughout the whole refactor](https://github.com/Shopify/packwerk/pull/74/files#r528821940). I would like to see a stronger boundary between the CLI and Command classes, which is why we removed the dependency of `@out` in the DetectStaleViolationsCommand.

This PR is pretty big, which makes it hard to review. You could help us review it faster by tackling one command because it would make it easier to review 🙂  I saw that you started with the init command. That command is relatively complex and needs some additional refactoring because it passes the `@out` dependency into the config generator. I would recommend starting with one of the other commands first.

PR #75, opened 2020-11-19
Introduce fixture helper module

rochlefebvre:
In [proof-of-concept PR #59](https://github.com/Shopify/packwerk/pull/59), I fleshed out a way to set up certain Packwerk tests by copying a template folder into a temporary directory, modifying those configuration fixtures, running the test, and then cleaning up the temporary artifacts.  In this PR, I will introduce the first of two helper modules to support that: ApplicationFixtureHelper.

**Note that I updated my code examples to use the new method names which came from the review process.**

Most test classes in Packwerk do not require a whole application directory to test their functionality.  There is no need for ApplicationFixtureHelper's methods to be available in every test.  Instead, the module simply needs to be included when it is needed.
```ruby
class YourThingyThatNeedsToBeRunningAgainstAnAppTest < Minitest::Test
  include ApplicationFixtureHelper
```
As mentioned in the POC PR, Packwerk uses the host application's current working directory to find its configuration artifacts and to inspect the application code.  Before the helper module copies the application directory into a temporary location, it needs to cache some state from the test process.
```ruby
  setup do
    setup_application_fixture  # caches current working directory
    # other setup logic...
  end
```
The test class needs to call the module's teardown method as well.  This method restores the old application state and deletes temporary artifacts.
```ruby
  teardown do
    # other teardown logic...
    teardown_application_fixture
  end
```
To copy an application fixture and change the current working directory, call `ApplicationFixtureHelper#use_template(:the_template)`:

image

Calling `use_template` as part of the `setup` method allows every test to have its own isolated copy of the same application fixture.
```ruby
  setup do
    setup_application_fixture  # caches current working directory
    use_template(:skeleton)
    # other setup logic...
  end

  teardown do
    # other teardown logic...
    teardown_application_fixture
  end

  test "wrecking my application's packwerk.yml causes pretty sparks to fly" do
    # ...
  end

  test "i'm with stupid :arrow_up: but I'm ok" do
    # assert things about the skeleton app
  end
```
Calling `use_template` inside a test is preferable when only some tests require a temporary application sandbox, or if the sandbox varies by test.
```ruby
  test "something using the minimal application" do
    use_template(:minimal)
    # assert things about the minimal app
  end
```

The module adds a few helper methods to work with the temporary application:
### `app_dir`
returns the root of the temporary application.  ~~It will first copy the `:minimal` template unless this has already occured for the test.~~ It will raise unless `use_template` has been called for this test.

### `config`
returns a `Configuration` object for the temporary application.

### `to_app_paths(*relative_paths)`
Converts relative paths into absolute paths rooted in the temporary application.

### `merge_into_app_yaml_file(relative_path, hash)`
Either create a new yaml file at the indicated path, or update the existing one by treating its contents as a hash and merging in the new hash.

Consider the existing array values for `b` and `c`,
```yml
# path/to/file.yml
a: some_value
b:
- 1
- 2
c:
- hi
```
Then calling 
```ruby
merge_into_app_yaml_file("path/to/file.yml", { b: [3,4], c: "bye" })
```
Will result in this yml
```yml
# path/to/file.yml
a: some_value
b:
- 1
- 2
- 3
- 4
c:
- hi
- bye
```

### `remove_app_entry(relative_path)`
Delete the file if it exists. 

I chose to implement the module and update a single existing suite to showcase its use.  I decided not to update other suites that already create temporary directories; I didn't see a clear benefit in changing them.  I also don't want to make this PR bigger than it already is.

Most of the new files make up the new `test/fixtures/minimal` application fixture, which is a slimmed down counterpart to `skeleton`, while still being a non-trivial one.  I would really like some feedback on what should go into this new, simple application.  I don't want another "everything and the kitchen sink" application like `skeleton`, but I also want `minimal` to be fleshed out enough to be useful without having to constantly add to it in most tests.

I do think it could be even slimmer than it already is, and I welcome @Shopify/packwerk veterans' suggestions for what should be left out.  After a quick settling period, `minimal` should not change very often.

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist

### DIFF
--- a/test/support/application_fixture_helper.rb
+++ b/test/support/application_fixture_helper.rb
@@ -1,0 +1,67 @@
+# typed: ignore
+# frozen_string_literal: true
+
+module ApplicationFixtureHelper
+  TEMP_FIXTURE_DIR = ROOT.join("tmp", "fixtures").to_s
+  DEFAULT_TEMPLATE = :minimal
+
+  def setup_application_fixture
+    @old_working_dir = Dir.pwd
+  end
+
+  def teardown_application_fixture
+    Dir.chdir(@old_working_dir)
+    FileUtils.remove_entry(@app_dir, true) if using_template?
+  end
+
+  def use_template(template)
+    raise "use_template may only be called once per test" if using_template?
+    copy_dir("test/fixtures/#{template}")
+    Dir.chdir(app_dir)
+  end
+
+  def app_dir
+    unless using_template?
+      raise "You need to set up an application fixture by calling `use_template(:the_template)`."
+    end
+
+    @app_dir
+  end
+
+  def config
+    @config ||= Packwerk::Configuration.from_path(app_dir)
+  end
+
+  def to_app_path(relative_path)
+    File.join(app_dir, relative_path)
+  end
+
+  def to_app_paths(*relative_paths)
+    relative_paths.map { |path| to_app_path(path) }
+  end
+
+  def merge_into_app_yaml_file(relative_path, hash)
+    path = to_app_path(relative_path)
+    YamlFile.new(path).merge(hash)
+  end
+
+  def remove_app_entry(relative_path)
+    FileUtils.remove_entry(to_app_path(relative_path))
+  end
+
+  private
+
+  def using_template?
+    defined? @app_dir
+  end
+
+  def copy_dir(path)
+    root = FileUtils.mkdir_p(fixture_path).last
+    FileUtils.cp_r("#{path}/.", root)
+    @app_dir = root
+  end
+
+  def fixture_path
+    File.join(TEMP_FIXTURE_DIR, "#{name}-#{Time.now.strftime('%Y_%m_%d_%H_%M_%S')}")
+  end
+end
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -4,17 +4,18 @@
 ENV["RAILS_ENV"] = "test"
 
 $LOAD_PATH.unshift(File.expand_path("../lib", __dir__))
+ROOT = Pathname.new(__dir__).join("..").expand_path
+
 require "constant_resolver"
 require "packwerk"
 
 require "minitest/autorun"
 require "minitest/focus"
+require "mocha/minitest"
+require "support/application_fixture_helper"
 require "support/test_macro"
 require "support/test_assertions"
-
-require "mocha/minitest"
-
-ROOT = Pathname.new(__dir__).join("..").expand_path
+require "support/yaml_file"
 
 Minitest::Test.extend(TestMacro)
 Minitest::Test.include(TestAssertions)
--- a/test/unit/configuration_test.rb
+++ b/test/unit/configuration_test.rb
@@ -5,6 +5,16 @@ require "test_helper"
 
 module Packwerk
   class ConfigurationTest < Minitest::Test
+    include ApplicationFixtureHelper
+
+    setup do
+      setup_application_fixture
+    end
+
+    teardown do
+      teardown_application_fixture
+    end
+
     test ".from_path raises ArgumentError if path does not exist" do
       File.expects(:exist?).with("foo").returns(false)
       error = assert_raises ArgumentError do
@@ -15,8 +25,8 @@ module Packwerk
     end
 
     test ".from_path uses packwerk config when it exists" do
-      File.expects(:exist?).with(".").returns(true)
-      File.expects(:file?).with("./packwerk.yml").returns(true)
+      use_template(:minimal)
+      remove_app_entry("packwerk.yml")
 
       configuration_hash = {
         "include" => ["xyz/*.rb"],
@@ -26,32 +36,32 @@ module Packwerk
         "custom_associations" => ["custom_association"],
         "inflections_file" => "custom_inflections.yml",
       }
-      YAML.expects(:load_file).with("./packwerk.yml").returns(configuration_hash)
+      merge_into_app_yaml_file("packwerk.yml", configuration_hash)
 
-      configuration = Configuration.from_path(".")
+      configuration = Configuration.from_path(app_dir)
 
       assert_equal ["xyz/*.rb"], configuration.include
       assert_equal ["{exclude_dir,bin,tmp}/**/*"], configuration.exclude
-      assert_equal File.expand_path("."), configuration.root_path
+      assert_equal app_dir, configuration.root_path
       assert_equal ["app/models"], configuration.load_paths
       assert_equal "**/*/", configuration.package_paths
       assert_equal ["custom_association"], configuration.custom_associations
-      assert_equal File.expand_path("custom_inflections.yml"), configuration.inflections_file
+      assert_equal to_app_path("custom_inflections.yml"), configuration.inflections_file
     end
 
     test ".from_path falls back to some default config when no existing config exists" do
-      File.expects(:exist?).with(Dir.pwd).returns(true)
-      File.expects(:file?).with(File.join(Dir.pwd, "packwerk.yml")).returns(false)
+      use_template(:minimal)
+      remove_app_entry("packwerk.yml")
 
       configuration = Configuration.from_path
 
       assert_equal ["**/*.{rb,rake,erb}"], configuration.include
       assert_equal ["{bin,node_modules,script,tmp,vendor}/**/*"], configuration.exclude
-      assert_equal File.expand_path("."), configuration.root_path
+      assert_equal app_dir, configuration.root_path
       assert_empty configuration.load_paths
       assert_equal "**/", configuration.package_paths
       assert_empty configuration.custom_associations
-      assert_equal File.expand_path("config/inflections.yml"), configuration.inflections_file
+      assert_equal to_app_path("config/inflections.yml"), configuration.inflections_file
     end
   end
 end
--- a/test/fixtures/minimal/components/sales/app/models/order.rb
+++ b/test/fixtures/minimal/components/sales/app/models/order.rb
@@ -1,0 +1,5 @@
+# typed: ignore
+# frozen_string_literal: true
+
+class Order
+end
--- a/test/fixtures/minimal/components/sales/app/models/sales/order.rb
+++ b/test/fixtures/minimal/components/sales/app/models/sales/order.rb
@@ -1,0 +1,7 @@
+# typed: ignore
+# frozen_string_literal: true
+
+module Sales
+  class Order
+  end
+end
--- a/test/fixtures/minimal/components/sales/app/public/sales/record_new_order.rb
+++ b/test/fixtures/minimal/components/sales/app/public/sales/record_new_order.rb
@@ -1,0 +1,7 @@
+# typed: ignore
+# frozen_string_literal: true
+
+module Sales
+  module RecordNewOrder
+  end
+end
--- a/test/fixtures/minimal/components/sales/package.yml
+++ b/test/fixtures/minimal/components/sales/package.yml
@@ -1,0 +1,8 @@
+---
+enforce_privacy: true
+
+metadata:
+  stewards:
+  - "@Shopify/sales"
+  slack_channels:
+  - "#sales"
--- a/test/fixtures/minimal/packwerk.yml
+++ b/test/fixtures/minimal/packwerk.yml
@@ -1,0 +1,4 @@
+include:
+  - "**/*.rb"
+load_paths:
+  - components/sales/app/models
--- a/test/support/yaml_file.rb
+++ b/test/support/yaml_file.rb
@@ -1,0 +1,39 @@
+# typed: ignore
+# frozen_string_literal: true
+
+class YamlFile
+  def initialize(path)
+    @path = path
+  end
+
+  def merge(hash)
+    merged_data = recursive_merge(read_or_create, hash)
+    write(merged_data)
+  end
+
+  private
+
+  attr_reader :path
+
+  def read_or_create
+    FileUtils.mkpath(File.dirname(path))
+    FileUtils.touch(path)
+    YAML.load_file(path) || {}
+  end
+
+  def write(data)
+    File.open(path, "w") { |f| YAML.dump(data, f) }
+  end
+
+  def recursive_merge(hash, other_hash)
+    hash.merge(other_hash) do |_, old_value, new_value|
+      if old_value.is_a?(Hash) && new_value.is_a?(Hash)
+        recursive_merge(old_value, new_value)
+      elsif old_value.is_a?(Array)
+        old_value + Array(new_value)
+      else
+        new_value
+      end
+    end
+  end
+end


test/support/application_fixture_helper.rb
```diff
@@ -0,0 +1,75 @@
+# typed: ignore
+# frozen_string_literal: true
+
+module ApplicationFixtureHelper
+  TEMP_FIXTURE_DIR = ROOT.join("tmp", "fixtures").to_s
```
rochlefebvre, line 5: I'm writing the temporary data in packwerk's own /tmp directory, instead of the system's temporary directory. While I was debugging tests, I found it quite convenient to see the temporary files right in the project, using my code editor. It's not so fun to figure out where the fixtures go in the system temp directory.

Also, if ever there is a leak of sorts with temporary fixture directories, it's easier to spot, being right in the project workspace.

test/support/application_fixture_helper.rb
```diff
@@ -0,0 +1,75 @@
...
+    FileUtils.remove_entry(@app_dir, true) if defined? @app_dir
+  end
+
+  def copy_template(template)
+    copy_dir("test/fixtures/#{template}")
+    Dir.chdir(app_dir)
+  end
+
+  def app_dir
+    copy_template(DEFAULT_TEMPLATE) unless defined? @app_dir
```
rochlefebvre, line 23: This line creates a temporary directory and copies over the `minimal` template unless `copy_template(x)` has already been called.  This effectively makes calls to `copy_template(:minimal)` optional, and allows ApplicationFixtureHelper's other methods to work without preconditions. Examples:

```ruby
test "the minimal app is valid" do
  validator = Packwerk::Packwerk::ApplicationValidator(
    config_file_path: config.config_path,  # the `config` method calls `app_dir`
    configuration: config
  )
  assert validator.check_all.ok?
end

test "I can just add a new components/timeline/package.yml file and test the new Package" do
  merge_into_yaml_file("components/timeline/package.yml", { "enforce_privacy" => true })
  package_set = PackageSet.load_all_from(app_dir)
  assert package_set.fetch("components/timeline").enforce_privacy
end
```
I'm not completely sold on the benefits, especially since this behaviour is quite magical.  I can leave it as is, or replace this line with 
```ruby
raise "You need to set up an application fixture by calling `copy_template(:the_template)`." unless defined? @app_dir
```
which makes `copy_template` a required setup step.

Thoughts?

test/fixtures/minimal/components/sales/app/public/sales/errors.rb
```diff
@@ -0,0 +1,8 @@
+# typed: ignore
+# frozen_string_literal: true
+
+module Sales
+  module Errors
+    SomethingWentWrong = Class.new(RuntimeError)
+  end
```
doodzik, line 7: > I do think it could be even slimmer than it already is, and I welcome @Shopify/packwerk veterans' suggestions for what should be left out. After a quick settling period, minimal should not change very often.

Since this is the minimal skeleton do we need to have so many modules in it? Shouldn't it be enough if we keep it to `Orders`, `Sales::Orderes`, `RecordNewOrdere` to be able to build on top of it? WDYT?

test/support/application_fixture_helper.rb
doodzik: I that case can you add it to the .gitignore file so we don't commit it accidentally?

test/support/application_fixture_helper.rb
doodzik: I would prefer to be explicit here to avoid a mystery guest from occurring. https://thoughtbot.com/blog/mystery-guest

test/support/application_fixture_helper.rb
```diff
@@ -0,0 +1,75 @@
...
+    FileUtils.cp_r("#{path}/.", root)
+    @app_dir = root
+  end
+
+  def fixture_path
+    File.join(TEMP_FIXTURE_DIR, "#{name}-#{Time.now.strftime("%Y%m%d")}")
+  end
+
+  def recursive_merge!(hash, other_hash)
+    hash.merge!(other_hash) do |_, old_value, new_value|
```
doodzik, line 65: I think it would be better to return a new value instead of mutating the data. That way it makes it easier to reason what a method does.

test/support/application_fixture_helper.rb
```diff
@@ -0,0 +1,75 @@
...
+  def app_dir
+    copy_template(DEFAULT_TEMPLATE) unless defined? @app_dir
+    @app_dir
+  end
+
+  def config
+    @config ||= Packwerk::Configuration.from_path(app_dir)
+  end
+
+  def path_to(relative_path)
```
doodzik, line 31: I don't see this method being used other than in this module. I think it would make sense to make it private.

test/support/application_fixture_helper.rb
```diff
@@ -0,0 +1,75 @@
...
+    relative_paths.map { |path| path_to(path) }
+  end
+
+  def merge_into_yaml_file(relative_path, hash)
+    path = path_to(relative_path)
+    FileUtils.mkpath(File.dirname(path))
+    FileUtils.touch(path)
+    data = YAML.load_file(path) || {}
+    recursive_merge!(data, hash)
+    File.open(path, "w") { |f| YAML.dump(data, f) }
```
doodzik, line 45: \nit I think it would be nice to abstract the method into multiple smaller private methods, which makes it easier to read. Something like:

```ruby
def merge_into_yaml_file(relative_path, hash)
    data = read_yaml(path)
    merged_data = recursive_merge(data, hash)
    write_yaml(path)
end

def read_yaml(relative_path)
    path = path_to(relative_path)
    FileUtils.mkpath(File.dirname(path))
    FileUtils.touch(path)
    YAML.load_file(path) || {}
end

def write_yaml(path)
    File.open(path, "w") { |f| YAML.dump(data, f) }
end
```

Thinking of it a bit more, we could potentially move that functionality into a separate YamlFile class since that is the data we are operating on. I think it would make the module a bit easier to understand.

```ruby
class YamlFile
  attr_reader :path

  def initialize(path)
    @path = path
  end

  def merge!(hash)
    merged_data = recursive_merge(read, hash)
    write(merged_data)
  end

  private
  
  def read
     path = path_to(relative_path)
     FileUtils.mkpath(File.dirname(path))
     FileUtils.touch(path)
     YAML.load_file(path) || {}
  end

  def write(data)
    File.open(path, "w") { |f| YAML.dump(data, f) }
  end
```

test/test_helper.rb
```diff
@@ -4,17 +4,17 @@
...
 require "minitest/autorun"
 require "minitest/focus"
+require "mocha/minitest"
 require "support/test_macro"
 require "support/test_assertions"
-
-require "mocha/minitest"
-
-ROOT = Pathname.new(__dir__).join("..").expand_path
+require "support/application_fixture_helper"
```
doodzik, line 17: Is the order of those require statements important or can we sort them alphabetically? If it is maybe it makes sense to add a comment?

test/unit/configuration_test.rb
```diff
@@ -26,31 +36,31 @@ class ConfigurationTest < Minitest::Test
         "custom_associations" => ["custom_association"],
         "inflections_file" => "custom_inflections.yml",
       }
-      YAML.expects(:load_file).with("./packwerk.yml").returns(configuration_hash)
+      merge_into_yaml_file("packwerk.yml", configuration_hash)
```
doodzik, line 39: Hmmm, reading the test case I find the naming of the helper methods a bit confusing. Maybe we should follow a naming standard so it is clear that we are working with the fixture helpers. Maybe `app_fixture_*` -> 
```ruby
app_fixture_use_template(:minimal)
app_fixture_merge_yaml_file(...) # or app_fixture_override_package_config(...)
app_fixture_remove_file("packwerk.yml") # instead of remove_app_entry()
```


What do you think?

test/support/application_fixture_helper.rb
doodzik: Just saw that it is used in the next commit.

test/unit/configuration_test.rb
```diff
@@ -26,31 +36,31 @@ class ConfigurationTest < Minitest::Test
...
       configuration = Configuration.from_path
 
       assert_equal ["**/*.{rb,rake,erb}"], configuration.include
       assert_equal ["{bin,node_modules,script,tmp,vendor}/**/*"], configuration.exclude
-      assert_equal File.expand_path("."), configuration.root_path
+      assert_equal app_dir, configuration.root_path
       assert_equal "**/", configuration.package_paths
       assert_empty configuration.custom_associations
-      assert_equal File.expand_path("config/inflections.yml"), configuration.inflections_file
+      assert_equal path_to("config/inflections.yml"), configuration.inflections_file
```
doodzik, line 63: Ditto https://github.com/Shopify/packwerk/pull/75/commits/89aeef3e7ef99086d0862538c9d029e9e31bd7ef#r528890926

test/support/application_fixture_helper.rb
```diff
@@ -58,7 +58,7 @@ def copy_dir(path)
   end
 
   def fixture_path
-    File.join(TEMP_FIXTURE_DIR, "#{name}-#{Time.now.strftime("%Y%m%d")}")
+    File.join(TEMP_FIXTURE_DIR, "#{name}-#{Time.now.strftime('%Y%m%d')}")
```
doodzik, line 61: Can you update the first commit to fix this rubocop failure. I think this is appropriate since you are introducing the failure in that commit and it is not yet on main.

doodzik (commented):
🎉 I have a couple of minimal suggestions. All in all, I think this is a great addition 🙂

test/support/application_fixture_helper.rb
rochlefebvre: If `/tmp/` is already on that list, I believe that's already covered.

test/support/application_fixture_helper.rb
rochlefebvre: Yeah, I agree. I tried the convenience on for size, but the explicit route is better here.

test/fixtures/minimal/components/sales/app/public/sales/errors.rb
rochlefebvre: Those three make sense, and we can revisit once tests start using the fixture.

test/test_helper.rb
rochlefebvre: I think `require` statements are sensitive to ordering in general, since `require` executes the file.  The new `application_fixture_helper` entry is not dependent on the other `support` entries.  I can move it up to follow alphabetical ordering, at least for now.

test/support/application_fixture_helper.rb
rochlefebvre: If my mental compilation is correct, then I think I can achieve this by simply switching to `hash.merge` on line 65, and then rename my method to `recursive_merge` to indicate it does not affect the parameter values.

Works for me!

test/support/application_fixture_helper.rb
rochlefebvre: I like your approach with a separate class.  I'll try that out.

test/support/application_fixture_helper.rb
rochlefebvre: New behaviour:
image
image

test/unit/configuration_test.rb
rochlefebvre: I don't think the suggested prefix adds much clarity, nor does it follow a common naming pattern.  If someone is confused, they'll find the helper module soon enough.

I find that `merge_into_yaml_file`, although not perfect, is truer to the actual behaviour than `override_package_config`.

I went with `remove_app_entry` since it models https://ruby-doc.org/stdlib-2.7.0/libdoc/fileutils/rdoc/FileUtils.html#method-c-remove_entry in its behaviour (works on directories as well, is idempotent, works on missing entries).

I will go with what I have for now and we can revisit later.

test/support/application_fixture_helper.rb
rochlefebvre: Sure! I'll roll that in when I update the commits with final reviewer suggestions.

test/unit/configuration_test.rb
doodzik: My issue with the current names is that they suggest behavior that they don't have.
For example, if I read `merge_into_yaml_file`, I would think of a method that takes a path to a YAML file and merges a hash with it. That is not what it does, though. It assumes a path for the fixture directory and merges a hash with a file in that directory.

Thinking about it, it seems like we have two functionalities. One is for setting up a fixture helper, and another for methods for manipulating IO. It would become clearer if we separate the two responsibilities, and the current names would make more sense. So, for example:

```ruby 
module FileHelper
  # If you look at those two methods, you can see that they are quite separate from the fixture functionality.
  def merge_into_yaml_file(path, hash)
    YamlFile.new(path).merge(hash)
  end

  def remove_file(path)
    FileUtils.remove_entry(path)
  end
```

And then we can call it like this:

```ruby
remove_app_entry(path_to("packwerk.yml"))
merge_into_yaml_file(path_to("packwerk.yml"), configuration_hash)
```

doodzik: If we separate the responsibilities, we can also make the other names a bit clearer. 

I would suggest to rename `copy_template(:minimal)` to `use_template(:minimal)`. Copy, in this case, describes how the fixture helper is setting up the template. Since a user can effectively use only one template, I would go with `use_template(:minimal)` here. It abstracts how the feature is implemented from the intended outcome. 

We could also clear up `path_to` because it is not clear where we are creating the path. I want to avoid someone needing to hunt down the implementation to understand what is happening in a test case. What do you think of something like `expand_app_fixture_path`, `expand_app_path` or `to_app_path`?

doodzik:
>  I decided not to update other suites that already create temporary directories; I didn't see a clear benefit in changing them. I also don't want to make this PR bigger than it already is.

Are you planning on changing the other test suites in a subsequent PR? I think it is important to change the other test suites to use the new pattern because otherwise, we will have two ways of using app fixtures. This allows room for cargo culting and makes the test suite less consistent.

test/unit/configuration_test.rb
rochlefebvre: You listed a lot of good alternatives to what I currently have.  Comparing this first approach
```ruby
remove_entry(path_to("packwerk.yml")) 
merge_into_yaml_file(path_to("components/sales/package.yml"), { enforce_privacy: false })
```
To this second one
```ruby
remove_app_entry("packwerk.yml")
merge_into_app_yaml_file("components/sales/package.yml", { enforce_privacy: false })
```
I like the second one better.  Adding `app` everywhere should clear things up.

I will rename `path_to` as well
```ruby
assert_equal to_app_paths(["components/sales"]), config.load_paths
```

rochlefebvre:
> > I decided not to update other suites that already create temporary directories; I didn't see a clear benefit in changing them. I also don't want to make this PR bigger than it already is.
> 
> Are you planning on changing the other test suites in a subsequent PR? I think it is important to change the other test suites to use the new pattern because otherwise, we will have two ways of using app fixtures. This allows room for cargo culting and makes the test suite less consistent.

Recall that I have yet to introduce the RailsApplicationFixtureHelper, which takes care of Rails-adjacent suites like ApplicationValidatorTest.  So I'm not done porting existing suites to application fixture helpers.

But there are many other suites that create temporary directories during setup.  A follow up PR will entertain the whole consistency vs. applicability debate.

doodzik (approved):
🚀

PR #77, opened 2020-11-20
Fix CLI init by defining default load path value

### LINKED ISSUE #76: [Bug Report] `packwerk init` failing due to TypeError
**Description**
When running `bundle exec packwerk init` on a fresh install, a Sorbet TypeError was thrown when generating the configuration file. This is due to a missing `load_paths` value, for which Sorbet expects to receive an Array. 

**To Reproduce**
1. Run `bundle exec packwerk init` in a project that doesn't yet have a packwerk config file

**Expected Behaviour**
All `init` generators should run without failure.

**Version Information**
 - Packwerk: v1.0.2
 - Ruby: v2.7.1

**Additional Context**
This problem seems to have been introduced in #68, specifically at https://github.com/Shopify/packwerk/pull/68/files#diff-e43184d1a960a3c3821da21b3eb72c83379c2dcc996f6bcfac9aadd6898cd374L45.

Error details:
```
Parameter 'load_paths': Expected type T::Array[String], got type NilClass (TypeError)
Caller: /Users/mikelkew/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/sorbet-runtime-0.5.6084/lib/types/private/methods/call_validation.rb:78
Definition: /Users/mikelkew/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/packwerk-1.0.2/lib/packwerk/generators/configuration_file.rb:21
```

mikelkew:
Resolves #76.

`ConfigurationFile#initialize` is expecting an Array value for `load_paths`, yet when called via the `packwerk init` CLI command, it was being passed as `nil` and thus throwing a Sorbet type violation.

I've just provided an empty array as a fallback if no `load_paths` value is defined, which is the case when running the config generator. Seemed like the most logical solution and matches the approach used for `custom_associations`.

I assume an empty array is the desired fallback here, although I guess `ApplicationLoadPaths.extract_relevant_paths` could also be an option.

## Type of Change

- [x] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist

### DIFF
--- a/lib/packwerk/cli.rb
+++ b/lib/packwerk/cli.rb
@@ -3,6 +3,7 @@
 require "benchmark"
 require "sorbet-runtime"
 
+require "packwerk/application_load_paths"
 require "packwerk/application_validator"
 require "packwerk/configuration"
 require "packwerk/files_for_processing"
@@ -101,7 +102,7 @@ module Packwerk
 
     def generate_configs
       configuration_file = Packwerk::Generators::ConfigurationFile.generate(
-        load_paths: @configuration.load_paths,
+        load_paths: Packwerk::ApplicationLoadPaths.extract_relevant_paths,
         root: @configuration.root_path,
         out: @out
       )
--- a/lib/packwerk/configuration.rb
+++ b/lib/packwerk/configuration.rb
@@ -42,7 +42,7 @@ module Packwerk
       @root_path = File.expand_path(root)
       @package_paths = configs["package_paths"] || "**/"
       @custom_associations = configs["custom_associations"] || []
-      @load_paths = configs["load_paths"]
+      @load_paths = configs["load_paths"] || []
       @inflections_file = File.expand_path(configs["inflections_file"] || "config/inflections.yml", @root_path)
 
       @config_path = config_path
--- a/test/unit/configuration_test.rb
+++ b/test/unit/configuration_test.rb
@@ -48,6 +48,7 @@ module Packwerk
       assert_equal ["**/*.{rb,rake,erb}"], configuration.include
       assert_equal ["{bin,node_modules,script,tmp,vendor}/**/*"], configuration.exclude
       assert_equal File.expand_path("."), configuration.root_path
+      assert_empty configuration.load_paths
       assert_equal "**/", configuration.package_paths
       assert_empty configuration.custom_associations
       assert_equal File.expand_path("config/inflections.yml"), configuration.inflections_file


mikelkew:
Note, CLA has been signed,. Tried reopening the PR to trigger a re-check of CLA status, but no luck..

doodzik (approved):

doodzik:
This change looks good to me 👍 Thank you for fixing the issue.
I would like to get @wildmaples or @rochlefebvre to have a look at this PR before we go ahead and merge it because I'm unsure about the default value.

rochlefebvre:
The Configuration class recently stopped inferring a set of load paths unless packwerk.yml file specified one:
https://github.com/Shopify/packwerk/pull/68/files#r524332054 .  It now expects the calling code to provide a list of paths.  Evidently, not all callers provide one.

For non-Rails applications, your change reinstates the previous behaviour of setting `load_paths = []` by default:
image

As for Rails applications, I think only the ApplicationValidator class will notice that Configuration#load_paths is inconsistent with Rails' own load paths:
https://github.com/Shopify/packwerk/blob/88526ccc3a9b5e4b85fa0bc5385da4c35bc59350/lib/packwerk/application_validator.rb#L50-L63

I think your change is a step in the right direction, @mikelkew .  Do you have anything to add, @wildmaples ?

wildmaples (changes requested):
I don't think the proposed solution in this PR is correct. By adding this change, we are not generating any load_paths for `packwerk init`. Instead it will just return an empty list. The correct behaviour is to get a list of load paths through Rails, and populate the configuration file. 

We'll have to change the load path argument on this line: 
https://github.com/Shopify/packwerk/blob/14890862e61869091813775d00d811213b09977a/lib/packwerk/cli.rb#L92 to use the new `ApplicationLoadPaths#extract_relevant_paths` method.

mikelkew:
Thanks @wildmaples, I had a suspicion that might be the desired behaviour. I've just pushed b9844f7 to address it (keeping the PR as a single commit as preferred).

I've intentionally kept the empty array as a fallback for `load_paths`  when loading a configuration file. This seems to be [expected to be an array elsewhere](https://github.com/Shopify/packwerk/blob/main/lib/packwerk/application_validator.rb#L52), so this covers us if a config file doesn't have it defined.

rochlefebvre:
Calling `ApplicationLoadPaths.extract_relevant_paths` in a non-Rails app will assert, So your solution is halfway there:
https://github.com/Shopify/packwerk/blob/88526ccc3a9b5e4b85fa0bc5385da4c35bc59350/lib/packwerk/application_load_paths.rb#L12-L23

_Somewhere_ in the generate configs logic, we're missing this pseudocode

```ruby
generated_load_paths = rails_app? ? ApplicationLoadPaths.extract_relevant_paths : []
```

### Solution 1
The `ApplicationLoadPaths` class is already aware of the application's potential for not being a Rails app, and raises when that's the case.  Having `extract_relevant_paths` return early with `[]` instead of asserting could work, and that's all we'd need to fix your PR.  `ApplicationLoadPaths` remains the source of truth for load path determination, even if it's to say ¯\_(ツ)_/¯.

### Solution 2
The CLI should not be in the business of implementing the decision modeled in the pseudocode.  But there is precedent for calling a generator with a `:for_rails_app` hint:

https://github.com/Shopify/packwerk/blob/88526ccc3a9b5e4b85fa0bc5385da4c35bc59350/lib/packwerk/cli.rb#L80-L87

Say we passed the same hint into `Packwerk::Generators::ConfigurationFile.generate` and pushed the default load paths decision into that generator:

```ruby
# lib/packwerk/cli.rb
Packwerk::Generators::ConfigurationFile.generate(
  for_rails_app: rails_app?,
  root: @configuration.root_path,
  out: @out
)

# lib/packwerk/generators/configuration_file.rb
class << self
  def generate(for_rails_app:, root:, out:)
    new(for_rails_app: for_rails_app, root: root, out: out).generate
  end
end

def initialize(for_rails_app:, root:, out: $stdout)
  @for_rails_app = for_rails_app
  @root = root
  @out = out

  set_template_variables
end

def load_paths
  @load_paths ||= @for_rails_app ? Packwerk::ApplicationLoadPaths.extract_relevant_paths : []
end

def set_template_variables
  @load_paths_formatted = if load_paths.empty?
    "# load_paths:\n# - 'app/models'\n"
  else
    load_paths.map { |path| "- #{path}\n" }.join
  end

  @load_paths_comment = unless load_paths.empty?
    "# These load paths were auto generated by Packwerk.\nload_paths:\n"
  end
end
```

I have a preference for the first solution.

mikelkew:
Thanks for catching that @rochlefebvre. I've gone with your first suggestion, both for simplicity, and because it matches the behaviour for non-rails apps that was in place before #68 was merged.

lib/packwerk/application_load_paths.rb
```diff
@@ -10,18 +10,14 @@ class << self
 
       sig { returns(T::Array[String]) }
       def extract_relevant_paths
-        assert_application_booted
+        return [] unless defined?(::Rails)
```
wildmaples, line 13: Is there a way to return `[]` at the end of this method instead of in an early return? 

`defined?(::Rails)` is really misleading because it (almost) always this will be true, as Packwerk itself has Rails and will fall back to that.

In the case where `defined?(::Rails)` works the way we intend it too though:

In the context of `ApplicationLoadPaths`, we would still want to raise if Rails is not defined because the point of the class is to get the Rails load paths. Returning an empty array would be misleading users to the actual functionality API of this class. 

TL;DR: We should not return empty array here but instead outside of this class, as it would be misleading to the functionality of this class.

wildmaples:
I had a chat with @rafaelfranca 

There are two bugs here: the first is solved by [this line of change](https://github.com/Shopify/packwerk/pull/77/files#diff-ac5ae86965a90253f8d378798aa32f4041ec009d6b79c4420e88be6e67108e88R105). 

The second issue is that `bundle exec packwerk` / `packwerk` and `bin/packwerk packwerk` produces different behaviour in Packwerk. `bin/packwerk` loads Rails files, while `bundle exec packwerk` does not. 

We previously had `bin/packwerk` because we wanted to keep the Rails part of the code separate from Packwerk as it may be used in a Ruby project. Now that Packwerk presumably only works with Rails (sorry, team Ruby), the solution is to load the Rails app within the `ApplicationLoadPaths` class. 

I will add an issue for this sometime tomorrow. In the meantime, I think the appropriate change is the line I linked above.

lib/packwerk/application_load_paths.rb
wildmaples: As per [my latest comment](https://github.com/Shopify/packwerk/pull/77#issuecomment-733284451), we can just leave this class as is and come back to add the Rails loading logic in another PR. When that is added, we won't have to raise when Rails is not defined, and just return an empty array as you proposed here. 

We could make the change now and return an empty array, but it would be hiding the fact that Rails _should_ be defined/loaded but is not. 

Unless someone has strong opinions, I say we keep the raise until we can make that change.

mikelkew: Ok, thanks @wildmaples, implemented in fe1697e

If we want to avoid raising for non-rails apps in the meantime, then I could implement [the second suggestion](https://github.com/Shopify/packwerk/pull/77#issuecomment-733002795) from @rochlefebvre instead. But if the PR for the Rails logic is going to be worked on sooner rather than later, perhaps the raise isn't too much of a concern for now.

rochlefebvre (approved):
This PR addresses the Sorbet violation when initializing a ConfigurationFile, at least within a Rails app.  It breaks `packwerk init` and `packwerk generate_configs` for non-Rails applications, now that `Packwerk::Generators::ConfigurationFile.generate_configs` relies on `ApplicationLoadPaths`.  We intend to track and fix that in a separate issue/PR.

Having Configuration return `[]` unless provided some load_paths prevents a NoMethodError in `ApplicationValidator#check_autoload_path_cache`, even if the check itself will fail, as designed.

apeacock1991:
Just stumbled upon this as I've just tried to install on a brand new Rails app and got the same error, is this going to be merged soon / tagged to fix the issue? As currently it's impossible (I think) to install Packwerk on any Rails app 😬

wildmaples (approved):
👍 Thanks, @mikelkew

PR #78, opened 2020-11-23
Refactor CLI: update-deprecations

santib:
Partially solves issue #70 by extracting `update-deprecations` command.

Notes:
- First commit: Extracts `update-deprecations` command logic. I introduced an `OffensesFormatter` so I'm able to print offenses somewhere within the command without the need to pass `out` and `style` (following what was done with `ProgressFormatter`). Otherwise we'll need to return something more complex as part of the `Result` of the command: offenses + execution time + message, which doesn't look good to me.
- Second commit: Adds a namespace to an existing command. I usually like it this way and I saw that for example the `Formatters` was a namespace so I think adding `Commands` would make the code more consistent. I can revert the commit if you don't like it.

The one suggested in the referenced issue above, because it was previously accepted by maintainers and to keep consistency with existing code.

Making sure they are good with the boundaries set and that I didn't break the command by mistake.

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist

### DIFF
--- a/lib/packwerk/commands/detect_stale_violations_command.rb
+++ b/lib/packwerk/commands/detect_stale_violations_command.rb
@@ -5,14 +5,13 @@ require "benchmark"
 require "packwerk/run_context"
 require "packwerk/detect_stale_deprecated_references"
 require "packwerk/commands/offense_progress_marker"
+require "packwerk/commands/result"
 
 module Packwerk
   module Commands
     class DetectStaleViolationsCommand
       extend T::Sig
       include OffenseProgressMarker
-      Result = Struct.new(:message, :status)
-
       def initialize(files:, configuration:, run_context: nil, progress_formatter: nil, reference_lister: nil)
         @configuration = configuration
         @run_context = run_context
@@ -54,7 +53,7 @@ module Packwerk
         if result_status
           message = "No stale violations detected"
         end
-        Result.new(message, result_status)
+        Result.new(message: message, status: result_status)
       end
     end
   end
--- a/lib/packwerk/commands/detect_stale_violations_command.rb
+++ b/lib/packwerk/commands/detect_stale_violations_command.rb
@@ -7,53 +7,55 @@ require "packwerk/detect_stale_deprecated_references"
 require "packwerk/commands/offense_progress_marker"
 
 module Packwerk
-  class DetectStaleViolationsCommand
-    extend T::Sig
-    include OffenseProgressMarker
-    Result = Struct.new(:message, :status)
+  module Commands
+    class DetectStaleViolationsCommand
+      extend T::Sig
+      include OffenseProgressMarker
+      Result = Struct.new(:message, :status)
 
-    def initialize(files:, configuration:, run_context: nil, progress_formatter: nil, reference_lister: nil)
-      @configuration = configuration
-      @run_context = run_context
-      @reference_lister = reference_lister
-      @progress_formatter = progress_formatter
-      @files = files
-    end
+      def initialize(files:, configuration:, run_context: nil, progress_formatter: nil, reference_lister: nil)
+        @configuration = configuration
+        @run_context = run_context
+        @reference_lister = reference_lister
+        @progress_formatter = progress_formatter
+        @files = files
+      end
 
-    sig { returns(Result) }
-    def run
-      @progress_formatter.started(@files)
+      sig { returns(Result) }
+      def run
+        @progress_formatter.started(@files)
 
-      execution_time = Benchmark.realtime do
-        @files.flat_map do |path|
-          run_context.process_file(file: path).tap do |offenses|
-            mark_progress(offenses: offenses, progress_formatter: @progress_formatter)
+        execution_time = Benchmark.realtime do
+          @files.flat_map do |path|
+            run_context.process_file(file: path).tap do |offenses|
+              mark_progress(offenses: offenses, progress_formatter: @progress_formatter)
+            end
           end
         end
+
+        @progress_formatter.finished(execution_time)
+        calculate_result
       end
 
-      @progress_formatter.finished(execution_time)
-      calculate_result
-    end
+      private
 
-    private
-
-    def run_context
-      @run_context ||= Packwerk::RunContext.from_configuration(@configuration, reference_lister: reference_lister)
-    end
-
-    def reference_lister
-      @reference_lister ||= ::Packwerk::DetectStaleDeprecatedReferences.new(@configuration.root_path)
-    end
-
-    sig { returns Result }
-    def calculate_result
-      result_status = !reference_lister.stale_violations?
-      message = "There were stale violations found, please run `packwerk update`"
-      if result_status
-        message = "No stale violations detected"
+      def run_context
+        @run_context ||= Packwerk::RunContext.from_configuration(@configuration, reference_lister: reference_lister)
       end
-      Result.new(message, result_status)
+
+      def reference_lister
+        @reference_lister ||= ::Packwerk::DetectStaleDeprecatedReferences.new(@configuration.root_path)
+      end
+
+      sig { returns(Result) }
+      def calculate_result
+        result_status = !reference_lister.stale_violations?
+        message = "There were stale violations found, please run `packwerk update`"
+        if result_status
+          message = "No stale violations detected"
+        end
+        Result.new(message, result_status)
+      end
     end
   end
 end
--- a/lib/packwerk/commands/update_deprecations_command.rb
+++ b/lib/packwerk/commands/update_deprecations_command.rb
@@ -1,10 +1,11 @@
-# typed: true
+# typed: strict
 # frozen_string_literal: true
 
 require "sorbet-runtime"
 require "benchmark"
 
 require "packwerk/commands/offense_progress_marker"
+require "packwerk/commands/result"
 require "packwerk/run_context"
 require "packwerk/updating_deprecated_references"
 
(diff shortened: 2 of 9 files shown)


santib:
Hi @doodzik thanks for your review on #74 I agree that it grew out of control 😆 so now I created this separate PR to tackle a single command, I chose to go with `update-deprecations` first. I get your idea about `out` and `styles` not being part of the commands. Because of that I needed to introduce the `OffensesFormatter`, but if you had some other solution in mind just let me know 😄

lib/packwerk/commands/detect_stale_violations_command.rb
```diff
@@ -10,54 +10,56 @@
...
+        calculate_result
+      end
 
-    private
+      private
 
-    def run_context
-      @run_context ||= Packwerk::RunContext.from_configuration(@configuration, reference_lister: reference_lister)
-    end
+      def run_context
```
doodzik, line 46: I think the command class doesn't need to know how the run_context is defined. I think it would be better to keep it in the cli class.

lib/packwerk/commands/detect_stale_violations_command.rb
```diff
@@ -10,54 +10,56 @@
...
-    Result = Struct.new(:message, :status)
-
-    def initialize(files:, configuration:, run_context: nil, progress_formatter: nil, reference_lister: nil)
-      @configuration = configuration
-      @run_context = run_context
-      @reference_lister = reference_lister
-      @progress_formatter = progress_formatter
-      @files = files
-    end
+  module Commands
```
doodzik, line 13: I like that you moved the class into its own namespace, which makes it cleaner :)

lib/packwerk/commands/update_deprecations_command.rb
```diff
@@ -0,0 +1,67 @@
...
+        end
+
+        @offenses_formatter.show_offenses(all_offenses)
+        @progress_formatter.finished(execution_time)
+        calculate_result(all_offenses)
+      end
+
+      private
+
+      def run_context
```
doodzik, line 47: Ditto https://github.com/Shopify/packwerk/pull/78/files#r530053258

lib/packwerk/commands/update_deprecations_command.rb
```diff
@@ -0,0 +1,67 @@
...
+          all_offenses = @files.flat_map do |path|
+            run_context.process_file(file: path).tap do |offenses|
+              mark_progress(offenses: offenses, progress_formatter: @progress_formatter)
+            end
+          end
+
+          updating_deprecated_references.dump_deprecated_references_files
+        end
+
+        @offenses_formatter.show_offenses(all_offenses)
```
doodzik, line 40: I think it is an excellent idea to move the offense formatter logic outside of this class. 👏 

If you think about the behavior of the offenses_formatter, it behaves like @out but takes a list of offenses and prints them. This essentially means that we have two ways of how a result message gets emitted from the command class, making it harder to reason about and test.

I would suggest changing the `OffenseFormatter` to return a string rather than printing to the stdout. The string would then be returned as the Result message. Something like this:

```ruby
class OffensesFormatter
      extend T::Sig

      def initialize(style: OutputStyles::Plain)
        @style = style
      end

      sig { params(offenses: T::Array[T.nilable(Offense)]).void }
      def format_offenses(offenses)
        return "No offenses detected 🎉"  if offenses.empty?
       
         <<~CONTENT
           #{offenses_list(offenses)}
           #{offenses_summary(offenses)}
         CONTENT
      end
      
      private
      
      def offenses_list(offenses)
        offenses
          .compact
          .map { |offense| offense.to_s(@style)) }
          .join("\n")
      end

      def offenses_summary(offenses)
        offenses_string = Inflector.default.pluralize("offense", offenses.length)
         "#{offenses.length} #{offenses_string} detected"
      end
  end
```

You would have to change the order when the offenses get printed concerning `@progress_formatter.finished(execution_time)`, but I think that is an acceptable change for what we are getting. Also, keep in mind that my change removes the pretending of an empty new line. The OffensesFormatter shouldn't be concerned about how its result is represented next to other content because that is not its responsibility.

lib/packwerk/commands/update_deprecations_command.rb
```diff
@@ -0,0 +1,67 @@
+# typed: true
```
doodzik, line 1: If we are introducing new files in the codebase I think it makes sense to make them strict by default unless it requires a lot of refactoring or is a test case.

doodzik (commented):
This PR starts to look really good. Once you address my comments I will test it in core to make sure that it works as expected 🙂 

I'm really happy about your contribution because your PR is improving the quality of the codebase! Thank you 🙇‍♂️

lib/packwerk/commands/detect_stale_violations_command.rb
santib: I'm not following how moving the `run_context` outside the command class will also remove the `reference_lister` dependency. It's being used to calculate the result here:
https://github.com/Shopify/packwerk/pull/78/files#diff-66788d8e2cafe199807e9963fd538243bcdd6ab0ea3203f07464cb93109e78c0R56
and

The options I see are to either pass also the `reference_lister` as a param, or to do `run_context.reference_lister. stale_violations?` to calculate the result but none of them convince me.. thoughts?

lib/packwerk/commands/update_deprecations_command.rb
santib: https://github.com/Shopify/packwerk/pull/78/commits/1b2b7f8ee983bdf18a6f3fad510b4cd7276b5582#diff-e2c1abfdf20fdf0a35e15eefaf378b134a2274fd170ae63865b041bfec4ed1b0R56

This change was to prevent sorbet complaining about memoization, but now when running tests I get

`lib/packwerk/commands/update_deprecations_command.rb:55: warning: instance variable @run_context not initialized`

any ideas?

test/integration/custom_executable_test.rb
```diff
@@ -54,10 +54,18 @@ class CustomExecutableTest < Minitest::Test
         assert_successful_run("update-deprecations")
 
         deprecated_reference_content_after_update = read_deprecated_references
+        expected_output = <<~EOS
```
santib, line 57: I did this to have more confidence on the refactor

lib/packwerk/commands/update_deprecations_command.rb
```diff
@@ -0,0 +1,82 @@
...
+        @run_context = T.let(@run_context, T.nilable(RunContext))
+        @run_context ||= RunContext.from_configuration(
+          @configuration,
+          reference_lister: updating_deprecated_references
+        )
+      end
+
+      sig { returns UpdatingDeprecatedReferences }
+      def updating_deprecated_references
+        @updating_deprecated_references = T.let(
```
santib, line 64: This change was to prevent sorbet complaining about memoization, but now when running tests I get

`lib/packwerk/commands/update_deprecations_command.rb:65: warning: instance variable @updating_deprecated_references not initialized`

any ideas?

lib/packwerk/commands/detect_stale_violations_command.rb
doodzik: Hmmm, I think the issue is that the `RunContext` class isn't a great abstraction for what we use it. But that is irrelevant to this PR. Let's keep the code you wrote 👍

lib/packwerk/commands/update_deprecations_command.rb
```diff
@@ -0,0 +1,82 @@
...
+require "packwerk/commands/offense_progress_marker"
+require "packwerk/commands/result"
+require "packwerk/run_context"
+require "packwerk/updating_deprecated_references"
+
+module Packwerk
+  module Commands
+    class UpdateDeprecationsCommand
+      extend T::Sig
+      include OffenseProgressMarker
```
doodzik, line 16: newline

lib/packwerk/commands/update_deprecations_command.rb
```diff
@@ -0,0 +1,82 @@
...
+      sig { returns RunContext }
+      def run_context
+        @run_context = T.let(@run_context, T.nilable(RunContext))
+        @run_context ||= RunContext.from_configuration(
+          @configuration,
+          reference_lister: updating_deprecated_references
+        )
+      end
+
+      sig { returns UpdatingDeprecatedReferences }
```
doodzik, line 62:
```suggestion
      sig { returns(UpdatingDeprecatedReferences) }
```

lib/packwerk/commands/update_deprecations_command.rb
doodzik: Unfortunately I have no idea. Have you tried to define the instance variable in the initializer?

doodzik (approved):
🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 
Really great work @santib 
🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉

I have to small comments but other than that I think it is good to go. I will check if the refactoring works in core and then we can merge it 🙂

santib:
Thanks for taking the time to review the PR thoroughly 😄 I just pushed the tweaks for the latest comments.

lib/packwerk/commands/detect_stale_violations_command.rb
santib: Once we have all the commands in their own classes it might be a good idea to try to refactor the RunContext 👌

doodzik:
Just ran it on core and it works like a charm 👌

PR #81, opened 2020-11-27
Introduce rails app fixture helper module

rochlefebvre:
Follow up to #75. This Rails-aware helper is built on top of the ApplicationFixtureHelper from that PR.

Tests that are sensitive to Rails load paths and the Rails root cannot work with the parent helper's copy of the application files into a temporary directory.

This PR introduces the RailsApplicationFixtureHelper module from POC #59.  The helper caches the Rails application singleton's root & load paths before each test, then restores them during teardown.  When `use_template(:the_template)` is called, the helper reconfigures Rails to use the temporary application fixture paths.

This PR introduces the helper and reworks the ApplicationValidatorTest suite's existing tests. In an upcoming PR, we will add more tests to the suite and likely use RailsApplicationFixtureHelper in more suites.

Consider reading through POC #59 if you wish to have more context.  The first commit reworks ApplicationValidatorTest in a few minor ways.  The other two commits introduce the helper module and then update ApplicationValidatorTest to use it.

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist


test/unit/application_validator_test.rb
```diff
@@ -67,7 +69,6 @@ class ApplicationValidatorTest < Minitest::Test
 
       config_path = "test/fixtures/skeleton/packwerk.yml"
       configs = YAML.load_file(config_path)
-      configs["inflections_file"] = "custom_inflections.yml"
```
rochlefebvre, line 70: This is already defined in fixtures/skeleton/packwerk.yml

test/support/rails_application_fixture_helper.rb
```diff
@@ -0,0 +1,75 @@
+# typed: ignore
+# frozen_string_literal: true
+
+require "rails_test_helper"
+
+module RailsApplicationFixtureHelper
+  include ApplicationFixtureHelper
+
+  def setup_application_fixture
+    super()
```
rochlefebvre, line 10: Have ApplicationFixtureHelper set up and tear down the temporary application fixture, then manage the Rails state.

test/support/rails_application_fixture_helper.rb
```diff
@@ -0,0 +1,75 @@
...
+    super(template)
+
+    case template
+    when :minimal
+      rails_autoload_paths(paths_to("/components/sales/app/models"))
+      rails_eager_load_paths([])
+      rails_autoload_once_paths([])
+      rails_root_path(app_dir)
+    when :skeleton
+      rails_autoload_paths(paths_to("components/sales/app/models"))
```
rochlefebvre, line 29: From https://github.com/Shopify/packwerk/blob/main/test/rails_test_helper.rb

test/unit/application_validator_test.rb
```diff
@@ -14,72 +12,66 @@
...
     end
 
     test "check_inflection_file returns error for mismatched inflections.yml file" do
-      config_path = "test/fixtures/skeleton/packwerk.yml"
-      configs = YAML.load_file(config_path)
-      configs["inflections_file"] = "different_inflections.yml"
-
-      configuration = Packwerk::Configuration.new(configs, config_path: config_path)
+      use_template(:skeleton)
+      merge_into_app_yaml_file("different_inflections.yml", { "acronym" => %w(TLA WTF LOL) })
```
rochlefebvre, line 44: We can simply generate this yaml file for the one test that requires it.  No need to commit it to source control.

test/support/rails_application_fixture_helper.rb
doodzik: Should we maybe rename the method to `setup_rails_application_fixture` to be more explicit about what is happening? There are benefits of allowing them to have the same name and using separate method names. I don't have a strong preference, but I would like to know how you think about it.

test/support/rails_application_fixture_helper.rb
```diff
@@ -0,0 +1,75 @@
...
+      )
+      rails_root_path(app_dir)
+    else
+      raise "Unknown fixture template #{template}"
+    end
+  end
+
+  private
+
+  def rails_autoload_paths(paths)
```
doodzik, line 46: \nit I see a lot of private methods that just deal with rails paths. I think it would be cleaner if those methods would live in a separate class such as RailsPaths. Something like this:

```ruby
class RailsPaths
  class << self
    def autoload(paths:)
      Rails.application.config.autoload_paths = paths
    end

    def eager_load(paths:)
      Rails.application.config.eager_load_paths = paths
    end

    def autoload_once(paths:)
      Rails.application.config.autoload_once_paths = paths
    end

    def rails_root(path:)
      Rails.application.config.root = path
    end

    def cache_rails_paths
      @autoload_paths= Rails.application.config.autoload_paths
      @eager_load_paths = Rails.application.config.eager_load_paths
      @autoload_once_paths = Rails.application.config.autoload_once_paths
      @root_path = Rails.application.config.root
    end

    def restore
      autoload(paths: @autoload_paths)
      eager_load(paths: @eager_load_paths)
      autoload_once(paths: @autoload_once_paths)
      root(path: @root_path)
    end
  end
end
```

test/support/rails_application_fixture_helper.rb
doodzik, line 29: Can we move the skeleton and minimal setup code into dedicated private methods? That will clear the method up because there is less happening in the copy_template class.

test/support/rails_application_fixture_helper.rb
```diff
@@ -0,0 +1,75 @@
...
+    super()
+    cache_rails_paths
+  end
+
+  def teardown_application_fixture
+    restore_rails_paths
+    super()
+  end
+
+  def copy_template(template)
```
doodzik, line 19: did you mean `use_template`?

test/support/rails_application_fixture_helper.rb
```diff
@@ -0,0 +1,75 @@
...
+      rails_autoload_paths(paths_to("components/sales/app/models"))
+      rails_eager_load_paths(paths_to("components/platform/app/models"))
+      rails_autoload_once_paths(
+        paths_to(
+          "components/timeline/app/models",
+          "components/timeline/app/models/concerns",
+          "vendor/cache/gems/example/models",
+        )
+      )
+      rails_root_path(app_dir)
```
doodzik, line 38: Shouldn't `rails_root_path(app_dir)` apply to any template we use?

doodzik (commented):
Looks pretty good 👏  I left a bunch of comments that should be pretty straight forward to address.

test/support/rails_application_fixture_helper.rb
rochlefebvre: I also prefer to have ApplicationFixtureHelper and RailsApplicationFixtureHelper present the same interface to test authors.

test/support/rails_application_fixture_helper.rb
rochlefebvre: Why yes, I did! The tests happen to pass because rails_test_helper.rb configures the Dummy application with Rails paths pointing to test/fixtures/skeleton/**, which happens to contain the same files as the copy directory.  And virtually all of the tests in ApplicationValidatorTest refer to skeleton, not minimal.

Good catch!

test/support/rails_application_fixture_helper.rb
rochlefebvre: It does.  I chose to repeat this line in both `when` blocks rather than extract it at the bottom.  If I shorten the method by extracting the path setting logic into separate methods, then I can have a single `rails_root_path(app_dir)` outside of the `case` statement.

test/support/rails_application_fixture_helper.rb
rochlefebvre: When I wrote this, I didn't see enough benefits to extracting the path setting and caching into a separate class.  Since then, I've thought of a few upcoming tests where I'd call something like `RailsPaths.autoload([])`.

test/support/rails_application_fixture_helper.rb
doodzik: > I also prefer to have ApplicationFixtureHelper and RailsApplicationFixtureHelper present the same interface to test authors.

I figured because you implemented it this way. What I want to understand is why you think this is the correct way and when you think it makes sense to have two distinct names?

doodzik (approved):
🚢 🍏

test/support/rails_application_fixture_helper.rb
rochlefebvre: The setup and teardown methods are part of this helper class' public interface.  Public methods names should aim to be declarative and short.  They should not explain _how_ something will be accomplished (something that is permissible in private helper methods), nor list all of the side effects, nor indicate where the method is defined.

~~`ApplicationFixtureHelper#copy_temp_directory_and_set_cwd`~~ --> `ApplicationFixtureHelper#setup_application_fixure`

~~`RailsApplicationFixtureHelper#copy_temp_directory_and_set_cwd_and_cache_rails_paths`~~ --> `RailsApplicationFixtureHelper#setup_application_fixure`

I also chose to override `#setup_application_fixture` so testers would not have access to both a `#setup_application_fixture` and a ``#setup_rails_application_fixture` at the same time.  Having both definitions provides no additional functionality, and only gives test implementers more ways to do something bad.

rochlefebvre:
The static type checking failures can be fixed with https://github.com/Shopify/packwerk/pull/83.  I will update the files I'm introducing in this PR as well.

PR #83, opened 2020-11-30
Set typed: false on all test helper classes, SanityChecker, and SpringCommand

rochlefebvre:
Fixes Sorbet type checking failures in test helper modules, such as [this one] in #81(https://github.com/Shopify/packwerk/pull/81/checks?check_run_id=1474632609)
image

From Sorbet documentation on [file-level strictness levels](https://sorbet.org/docs/static#file-level-granularity-strictness-levels)
>Each strictness level reports all errors at lower levels, plus new errors:
>
>At # typed: ignore, the file is not even read by Sorbet, and so no errors at all are reported in that file. Note: ignoring a file can cause errors to appear in other files, because that other file references something defined in an ignored file. We recommend pushing the entire project to out of ignore (at Stripe, 100% of non-test files are not ignored.)

So this change gets packwerk closer to removing `typed: ignore` completely.

I enabled basic type checking only on support classes & files, such as `test/test_helper.rb` and `test/support/*.rb`, but not on test suites or fixture files.   I also found that SpringCommand and SanityChecker were not being type checked, and enabled those two as well.

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist

### DIFF
--- a/lib/packwerk/sanity_checker.rb
+++ b/lib/packwerk/sanity_checker.rb
@@ -1,4 +1,4 @@
-# typed: ignore
+# typed: false
 # frozen_string_literal: true
 
 require "packwerk/application_validator"
--- a/lib/packwerk/spring_command.rb
+++ b/lib/packwerk/spring_command.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
-# typed: ignore
+# typed: false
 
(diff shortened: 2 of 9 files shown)


exterm (approved):
🍨 💪

PR #85, opened 2020-12-01
Convert more tests to using application fixture helpers

rochlefebvre:
Rework some existing test suites to use the new application fixture helpers introduced in #75 and #81.  As mentioned in [this comment thread](https://github.com/Shopify/packwerk/pull/75#issuecomment-734330902), we would like to have consistent test setup/teardown patterns where possible.

Have tests call `use_template(:skeleton)` and read/write into the temporary copy instead of:
- reading from `test/fixtures/skeleton/*` directly;
- creating temporary application files (e.g. yml or rb) into a separate system temp dir.

Although many suites were working just fine with the existing setup/teardown code, I chose to replace those with equivalent aplication fixture setup/teardown, which does not always offer a clear benefit.  I'm doing so for consistency.

Are there any suites that should not be adapted? Are there any suites that **should** be adapted?

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist

### DIFF
--- a/test/integration/custom_executable_test.rb
+++ b/test/integration/custom_executable_test.rb
@@ -7,14 +7,18 @@ require "rails_test_helper"
 module Packwerk
   module Integration
     class CustomExecutableTest < Minitest::Test
+      include ApplicationFixtureHelper
+
+      TIMELINE_PATH = File.join("components", "timeline")
+
       setup do
         @out = StringIO.new
-        @previous_cwd = Dir.pwd
-        Dir.chdir(skeleton_root)
+        setup_application_fixture
+        use_template(:skeleton)
       end
 
       teardown do
-        Dir.chdir(@previous_cwd)
+        teardown_application_fixture
       end
 
       test "'packwerk check' with no violations succeeds" do
@@ -24,7 +28,7 @@ module Packwerk
       end
 
       test "'packwerk check' with violations fails and displays violations" do
-        Tempfile.create(["timeline_comment", ".rb"], timeline_path("app", "models")) do |file|
+        open_app_file(TIMELINE_PATH, "app", "models", "timeline_comment.rb") do |file|
           file.write("class TimelineComment; belongs_to :order, class_name: '::Order'; end")
           file.flush
 
@@ -36,7 +40,7 @@ module Packwerk
       end
 
       test "'packwerk check' with violations accounts for custom inflections, fails and displays violations" do
-        Tempfile.create(["timeline_comment", ".rb"], timeline_path("app", "models")) do |file|
+        open_app_file(TIMELINE_PATH, "app", "models", "timeline_comment.rb") do |file|
           # payment_details is an uncountable inflection
           file.write("class TimelineComment; has_one :payment_details; end")
           file.flush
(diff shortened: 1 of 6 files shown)


test/integration/custom_executable_test.rb
```diff
@@ -86,8 +88,6 @@ class CustomExecutableTest < Minitest::Test
             "expected no updates to any deprecated references files besides timeline/deprecated_references.yml")
 
           assert_match(/`deprecated_references.yml` has been updated./, captured_output)
-        ensure
```
rochlefebvre, line 89: The entire application fixture directory is deleted during teardown.  There is no need to create Tempfiles nor ensure cleanup.

test/unit/package_set_test.rb
```diff
@@ -5,8 +5,16 @@
 
 module Packwerk
   class PackageSetTest < Minitest::Test
+    include RailsApplicationFixtureHelper
+
     setup do
-      @package_set = PackageSet.load_all_from("test/fixtures/skeleton/")
```
rochlefebvre, line 9: I would like packwerk tests to treat `test/fixtures/skeleton` and `test/fixtures/minimal` as templates to copy into a temporary artifact instead of reading directly from them.

test/unit/package_set_test.rb
```diff
@@ -33,38 +41,38 @@ class PackageSetTest < Minitest::Test
     end
 
     test ".package_paths supports a path wildcard" do
-      package_paths = PackageSet.package_paths("test/fixtures/skeleton", "**")
+      package_paths = PackageSet.package_paths(".", "**")
```
rochlefebvre, line 44: While `use_template(:x)` is in effect, the current working directory changes to `tmp/fixtures/the_unique_fixture_copy_for_this_test`, which is the root of the application.

test/unit/package_set_test.rb
```diff
@@ -33,38 +41,38 @@ class PackageSetTest < Minitest::Test
     end
 
     test ".package_paths supports a path wildcard" do
-      package_paths = PackageSet.package_paths("test/fixtures/skeleton", "**")
```
rochlefebvre, line 36: The current working directory used to be the packwerk project directory itself.

test/integration/custom_executable_test.rb
```diff
@@ -24,7 +26,7 @@ class CustomExecutableTest < Minitest::Test
       end
 
       test "'packwerk check' with violations fails and displays violations" do
-        Tempfile.create(["timeline_comment", ".rb"], timeline_path("app", "models")) do |file|
+        File.open(timeline_path("app", "models", "timeline_comment.rb"), "w") do |file|
```
doodzik, line 29: Should we maybe create a helper for this type of operation to simplify it? 

```ruby

# application helper module
  def open_app_file(*path)
      expanded_path = to_app_path(File.join(*path))
      File.open(expanded_path, "w") { |file| yield file }
  end

# test 

open_app_file(timeline_path("app", "models", "timeline_comment.rb")) do |file|
  # ...
end
```

test/unit/package_set_test.rb
doodzik: Hmmm, I'm not sure if this would make things harder to understand, because we are mixing absolute and relative references to the application skeleon in the test. Can't we pass the app_dir to PackageSet?

doodzik (commented):
Looks pretty good. Two small comments than it should be good to go I think.

test/unit/package_set_test.rb
rochlefebvre: I just discovered that real calls to `PackageSet.load_all_from` pass in an absolute path, whereas this test suite uses a relative one.  We should test it the way it's used, and pass in an absolute path.  I'll switch over to `app_dir` and see what happens.

test/integration/custom_executable_test.rb
rochlefebvre: I like it.

doodzik (approved):
🎉

exterm (approved):
❤️

PR #87, opened 2020-12-03
Ensure parsed configuration is a Hash

alexblackie:
Setting up packwerk caused a bug during the initial `packwerk check`:

```
bundler: failed to load command: packwerk (/Users/alexblackie/.gem/ruby/2.6.6/bin/packwerk)
NoMethodError: undefined method `[]' for false:FalseClass
  /Users/alexblackie/.gem/ruby/2.6.6/gems/packwerk-1.1.0/lib/packwerk/configuration.rb:39:in `initialize'
  /Users/alexblackie/.gem/ruby/2.6.6/gems/packwerk-1.1.0/lib/packwerk/configuration.rb:25:in `new'
```

This appears to have been because the YAML configuration file that is
generated is empty (comments only) by default, causing `load_file` to
return `false` because it isn't yet a valid YAML document.

To ensure we comply with the API of `Configuration`, we should ensure
any falsey values get turned into an empty hash.

After this change we get the error we expected, instead of the one above:

```
 → be packwerk check
📦 Packwerk is inspecting 465 files
bundler: failed to load command: packwerk (/Users/alexblackie/.gem/ruby/2.6.6/bin/packwerk)
ConstantResolver::Error: could not find any files
 Make sure autoload paths are added to the config file.
```

Co-authored-by: Jacob Erickson <jacob.erickson@shopify.com>

## Type of Change

- [x] Bugfix
- [ ] New feature

## Checklist

### DIFF
--- a/lib/packwerk/configuration.rb
+++ b/lib/packwerk/configuration.rb
@@ -22,7 +22,10 @@ module Packwerk
       private
 
       def from_packwerk_config(path)
-        new(YAML.load_file(path), config_path: path)
+        new(
+          YAML.load_file(path) || {},
+          config_path: path
+        )
       end
     end
 
--- a/test/unit/configuration_test.rb
+++ b/test/unit/configuration_test.rb
@@ -63,5 +63,11 @@ module Packwerk
       assert_empty configuration.custom_associations
       assert_equal to_app_path("config/inflections.yml"), configuration.inflections_file
     end
+
+    test ".from_path falls back to empty config when existing config is an empty document" do
+      use_template(:blank)
+      Configuration.expects(:new).with({}, config_path: to_app_path("packwerk.yml"))
+      Configuration.from_path
+    end
   end
 end


wildmaples (approved):
Thanks @alexblackie and @jaruserickson !

lib/packwerk/configuration.rb
```diff
@@ -22,7 +22,10 @@ def from_path(path = Dir.pwd)
       private
 
       def from_packwerk_config(path)
-        new(YAML.load_file(path), config_path: path)
+        new(
+          YAML.load_file(path) || {},
```
doodzik, line 26: Can you move this line of code into a variable dedicated variable? That way the instantiation becomes a bit easier to read.

doodzik (approved):

lib/packwerk/configuration.rb
alexblackie: Ah, I actually purposefully avoided it to minimize allocations, but I can certainly add one. :+1:

alexblackie: Personally, I find it more difficult to read with a variable in there:

```ruby
yaml_config = YAML.load_file(path) || {}
new(yaml_config, config_path: path)
```

PR #89, opened 2020-12-08
give more helpful error message if privatized constant can not be resolved

exterm:

We recently had someone stumbling over the error message that appears when a constant that Packwerk can't resolve is declared as private. This is due to `constant_resolver` not auto-vivifying namespaces.

I added some detail to the error message. And while I was at it, I also did two other things:

- no longer generally require a constant to be defined in its own file. Packwerk should now be able to always correctly resolve something like `Nested::Constant` to `nested.rb` if `nested/constant.rb` doesn't exist, and to `nested/constant.rb` otherwise. This check was a leftover from before we supported custom inflections.
- added a check that makes sure that a constant can only be declared private on the package that defines it.

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist

### DIFF
--- a/lib/packwerk/application_validator.rb
+++ b/lib/packwerk/application_validator.rb
@@ -35,16 +35,7 @@ module Packwerk
         check_root_package_exists,
       ]
 
-      results.reject!(&:ok?)
-
-      if results.empty?
-        Result.new(true)
-      else
-        Result.new(
-          false,
-          results.map(&:error_value).join("\n===\n")
-        )
-      end
+      merge_results(results)
     end
 
     def check_autoload_path_cache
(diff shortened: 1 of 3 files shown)


exterm:
@graces354li reported this problem, thank you Grace!

lib/packwerk/application_validator.rb
```diff
@@ -65,37 +65,42 @@ def check_autoload_path_cache
...
 
+        this_package = package_set.package_from_path(relative_path(filepath))
+
         setting.each do |constant|
           # make sure the constant can be loaded
           constant.constantize # rubocop:disable Sorbet/ConstantsFromStrings
-          context = resolver.resolve(constant)
 
-          unless context
-            errors << "#{constant}, listed in #{filepath.inspect}, could not be resolved"
```
exterm, line 86: Not sure how I ever thought this would be a helpful error message 🤷

lib/packwerk/application_validator.rb
```diff
@@ -65,31 +65,42 @@ def check_autoload_path_cache
...
         setting.each do |constant|
           # make sure the constant can be loaded
           constant.constantize # rubocop:disable Sorbet/ConstantsFromStrings
 
+          # make sure the constant can be resolved
           constant_context = resolver.resolve(constant)
-          next if constant_context
 
-          explicit_filepath = (constant.start_with?("::") ? constant[2..-1] : constant).underscore + ".rb"
+          if constant_context
```
rochlefebvre, line 89: `check_package_manifests_for_privacy` was already quite long, and now it has two more variables and an additional branch of logic.  How about we move the `if` block into its own method?  That allows us to move both `package_set` and `this_package` along with it.

The `else` block could move into a method as well.

exterm: Good point.

exterm: `package_set` I wouldn't want to recompute for each package, but I guess we can memoize it

rochlefebvre: I think `package_set` is only used in the new code for detecting cross-package privatization™. If you're moving all of that code inside a new method, you can load all package sets within it.

rochlefebvre:
```ruby
def verify_private_constant_origin(filepath, constant_context)
  package_set = Packwerk::PackageSet.load_all_from(@configuration.root_path, 
package_pathspec: package_glob)
  this_package = package_set.package_from_path(relative_path(filepath))
  constant_package = package_set.package_from_path(constant_context.location)

  unless constant_package == this_package
    errors << "'#{constant}' is declared as private in the '#{this_package}' package but appears to be "\
        "defined\nin the '#{constant_package}' package. Packwerk resolved it to #{constant_context.location}."
  end
end
```

exterm: that would be called in the loop iterating over all packages, and re-initialize the package set for each iteration.

exterm: I did a refactor. Had to make quite a few modifications to make it work.

lib/packwerk/application_validator.rb
```diff
@@ -333,7 +296,7 @@ def package_glob
       @configuration.package_paths || "**"
     end
 
-    def package_manifests(glob_pattern)
+    def package_manifests(glob_pattern = package_glob)
```
rochlefebvre, line 299: I like this change.

rochlefebvre (approved):
You made some nice quality-of-life changes to ApplicationValidator.  The error messages are more helpful as well.  The PR covers several different types of changes at once, but it's definitely manageable.

Thanks!

Issue #90, opened 2020-12-09
[Bug Report] bundle exec packwerk init is throwing a Sorbet Type Error


steve-low:
**Description**
Running `bundle exec packwerk init` is failing while creating application_load_paths due to Sorbet error: `Return value: Expected type T::Array[String], got T::Array[T.any(Pathname, String)]`

**To Reproduce**
This may be repo dependent, but running `bundle exec packwerk init` failed on the Shopify Flow repo.

**Expected Behaviour**
No error.

**Screenshots**
![image](https://user-images.githubusercontent.com/5199046/101661054-d360e900-3a15-11eb-94b7-95b553d14d92.png)

**Version Information**
 - Packwerk: 1.0.0 & 1.0.1
 - Ruby 2.6.5

**Additional Context**
I found a workaround to get it working and was able to initialize, validate, and execute packwerk checks. 
I changed https://github.com/Shopify/packwerk/blob/499750f665d862cde4d55cf04122fabffccf039e/lib/packwerk/application_load_paths.rb#L31
from:
```
(engine.config.autoload_paths + engine.config.eager_load_paths + engine.config.autoload_once_paths).uniq
```
to:
```
(engine.config.autoload_paths + engine.config.eager_load_paths + engine.config.autoload_once_paths).map(&:to_s).uniq
```  
I can submit a PR with this change if those with more packwerk context think this is the correct change to make.

exterm:
Thank you for reporting this!

🤔 

Yeah, the cast to string seems like a workable solution to make the code satisfy the type signature.

I wonder if this changed in Rails, and why this didn't happen in CI 🤔 

It would be great if you could look at adding a test too.

steve-low:
So it looks like it's not something that changed in Rails, it's a Pathname that's added in the Flow repo. It adds a Pathname to the `eager_load_path` [here](https://github.com/Shopify/flow/blob/master/config/application.rb#L52) and that's what causes the exception. I'll update the Flow repo to add this Pathname as a String. 

I will still open the PR I have ready for packwerk in case it's something that would still be useful to protect against this error. If it's seen as not necessary we can close.

exterm:
fixed

PR #91, opened 2020-12-10
Convert autoload paths to strings to prevent sorbet error related to Pathname

steve-low:
Fix a Sorbet error that occurs when running `bundle exec packwerk init`,  due to autoload paths potentially bing a pathname. Detailed in the issue here - https://github.com/Shopify/packwerk/issues/90.

Since making this PR I've discovered the Pathname causing the exception in the Repo mentioned in the issue was a custom `eager_load_path` added by the repo. I am going to update the Repo in question to provide the `eager_load_path` as a String, which will fix the issue, but the question I ask of the packwerk team is - Are Pathname's valid `eager_load_paths` that should be handled? If the answer is `yes`, then I think this PR would be useful, otherwise we can close it. I don't have enough context in Rails engines to answer this, but the repo from the issue seems to have been working correctly, perhaps we should handle this.

Map the autoload paths to a String so that any Pathname returned for a Rails engine autoload path is converted. This fixes the Sorbet type check and the Pathname value is included in the autoload paths. 
An alternative would have been to make the type returned from the `extract_application_autoload_paths` method a `T.any(Pathname, String)` but I chose to keep the typing simple rather than make the caller of the method have to handle 2 types.

- Do we need to do this String conversion? Should we handle Pathnames as `eager_load_paths`.
- Is there another approach that could be better?
- Is the testing sufficient?

## Type of Change

- [X] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist

### DIFF
--- a/lib/packwerk/application_load_paths.rb
+++ b/lib/packwerk/application_load_paths.rb
@@ -28,8 +28,9 @@ module Packwerk
           .select { |railtie| railtie.is_a?(Rails::Engine) }
           .push(Rails.application)
           .flat_map do |engine|
-          (engine.config.autoload_paths + engine.config.eager_load_paths + engine.config.autoload_once_paths).uniq
-        end
+            paths = (engine.config.autoload_paths + engine.config.eager_load_paths + engine.config.autoload_once_paths)
+            paths.map(&:to_s).uniq
+          end
       end
 
       sig do
--- a/test/unit/application_load_paths_test.rb
+++ b/test/unit/application_load_paths_test.rb
@@ -53,5 +53,21 @@ module Packwerk
 
       assert_equal 1, ApplicationLoadPaths.extract_relevant_paths.count
     end
+
+    test ".extract_application_autoload_paths returns unique autoload paths" do
+      path = Pathname.new("/application/app/models")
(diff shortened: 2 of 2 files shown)


rochlefebvre (approved):
Given that paths are external inputs into packwerk, and that `to_s` is a fairly safe bit of defensive code, I'm OK with this change.

exterm (approved):

lib/packwerk/application_load_paths.rb
```diff
@@ -28,7 +28,9 @@ def extract_application_autoload_paths
           .select { |railtie| railtie.is_a?(Rails::Engine) }
           .push(Rails.application)
           .flat_map do |engine|
-          (engine.config.autoload_paths + engine.config.eager_load_paths + engine.config.autoload_once_paths).uniq
+          (engine.config.autoload_paths + engine.config.eager_load_paths + engine.config.autoload_once_paths)
```
doodzik, line 31: I find this a bit hard to read. Can you move the contamination into a separate descriptive variable?

doodzik (approved):

lib/packwerk/application_load_paths.rb
steve-low: I've updated to be more readable.

PR #92, opened 2020-12-10
Improve types on interface implementations 

samuelgiles:

To improve the types on implementations of `Packwerk::Checker`, `Packwerk::ConstantNameInspector` and `Packwerk::ReferenceLister` so that where possible the strictness level can be upped to `strict` to enable the interfaces these classes implement be better enforced by Sorbet.

- Added types on the implementations of abstract methods within mentioned classes whilst leaving the sigil as `typed: true`
- Later upgrading to `typed: strict` via the addition of types on all other methods

- I note this is the first usage of `T.type_alias` in the project, are we happy with the usage of this or has it been actively avoided?

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist

### DIFF
--- a/lib/packwerk/const_node_inspector.rb
+++ b/lib/packwerk/const_node_inspector.rb
@@ -6,8 +6,14 @@ require "packwerk/constant_name_inspector"
 module Packwerk
   # Extracts a constant name from an AST node of type :const
   class ConstNodeInspector
+    extend T::Sig
     include ConstantNameInspector
 
+    sig do
+      override
+        .params(node: AST::Node, ancestors: T::Array[AST::Node])
+        .returns(T.nilable(String))
+    end
     def constant_name_from_node(node, ancestors:)
       return nil unless Node.constant?(node)
       parent = ancestors.first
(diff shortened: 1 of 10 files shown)


lib/packwerk/const_node_inspector.rb
```diff
@@ -28,23 +34,27 @@ def constant_name_from_node(node, ancestors:)
...
       parent_name && parent_name == Node.constant_name(node)
     end
 
+    sig { params(node: AST::Node, ancestors: T::Array[AST::Node]).returns(String) }
     def fully_qualify_constant(node, ancestors:)
       # We're defining a class with this name, in which case the constant is implicitly fully qualified by its
       # enclosing namespace
       name = Node.parent_module_name(ancestors: ancestors)
-      name ||= generate_qualified_constant(node, ancestors)
+      name ||= generate_qualified_constant(node, ancestors: ancestors)
```
exterm, line 53: this is strange. How did this even work before?

exterm: Is this line covered by tests?

exterm:
Pretty cool, thank you @samuelgiles ! I think @doodzik will be interested in this

lib/packwerk/const_node_inspector.rb
samuelgiles: I know right? It can't have been covered by tests I guess 🤷‍♂️  - I was quite surprised Sorbet didn't catch it just based on the arguments

exterm: yeah it's sometimes surprising what sorbet misses.

doodzik: I guess we are never running into a situation where we need to generate a qualified constant because Node.parent_module_name is not returning nil.  Core is also not running into this issue so maybe we can get rid of the method and see what happens?

doodzik:
This is so great!!! Thanks, @samuelgiles for adding more type information to Packwerk!

After you resolve @exterm comment we can merge this PR.

lib/packwerk/const_node_inspector.rb
samuelgiles: I chased this today:
- `name = Node.parent_module_name(ancestors: ancestors)` is never empty, it will always return `"Object"` or the module name
- This means we never run the `name ||=` line since `name` is never `nil`

I added a sig to `Node.parent_module_name` in https://github.com/Shopify/packwerk/pull/92/commits/d1a5b7d35fcfd432b2bd41278d6e604ac6b5dacf and at this point Sorbet was able to confirm this line is unreachable:
image

doodzik: @tomstuart @wildmaples  You might have a bit more context why we were using generate_qualified_constant just to make sure that we don't have a regression. Given that we aren't reaching this core method and don't have any regressions, we can probably get rid of it. WDYT?

tomstuart: I don’t, sorry. This line of code was present when Packwerk was [first extracted](https://github.com/shopify/packwerk-old/commit/7112ab0813f7c4cad859c0519c4baee24894041e#diff-b9878467a38ac6cfae24fc57d7ef5e853dfeb5f5b6873147c648201725ce6c01R37) from Shopify’s core monolith and I haven’t traced it any further back than that. The most likely explanation is that the author was unaware that `RuboCop::AST::Node#parent_module_name` is [incapable of returning `nil`](https://github.com/rubocop-hq/rubocop-ast/blob/5cd306f40e5d5ba4dacf78c698354747cdac7825/lib/rubocop/ast/node.rb#L312-L320) and was trying to be defensive. I agree we should remove it.

exterm (approved):
Fantastic, thank you Samuel!

doodzik (approved):

PR #93, opened 2020-12-11
Improve types surrounding Packwerk::OutputStyles

samuelgiles:

Making it safer to work with `Packwerk::OutputStyles` and easier to add additional output styles in future.

- ~Consolidated all `T.any(T.class_of(OutputStyles::Plain...` references into a `OutputStyles::Any` type alias~
  - Edit: Introduced a `Packwerk::OutputStyle` interface and switched the existing singleton based output style classes to instances that implement said interface to avoid using to use `T.any`
- Wrapped all other uses of output styles with `sig` to reveal type

- ~Are we happy with the `.stubs(:is_a?).with(X)` usage to keep mocks working alongside Sorbet? For classes like `Packwerk::Configuration` it might require a lot of stubbing to keep them working if we tried to use actual instances~
- ~I note the use of singleton methods here prevents us from having an `OutputStyle` interface, a `T::Enum` with methods using `case` or just instance methods (which would mean `OutputStyles::Plain.new`) might be better alternatives to a `T.any`~

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist


lib/packwerk/output_styles.rb
```diff
@@ -1,18 +1,25 @@
-# typed: true
+# typed: strict
 # frozen_string_literal: true
 
 module Packwerk
   module OutputStyles
+    Any = T.type_alias { T.any(T.class_of(Plain), T.class_of(Coloured)) }
```
doodzik, line 6: ❤️

test/unit/cli_test.rb
```diff
@@ -21,8 +21,10 @@ class CliTest < Minitest::Test
     test "#execute_command with the subcommand check starts processing files" do
       violation_message = "This is a violation of code health."
       offense = stub(error?: true, to_s: violation_message)
+      offense.stubs(:is_a?).with(Packwerk::Offense).returns(true)
```
doodzik, line 24: > Are we happy with the .stubs(:is_a?).with(X) usage to keep mocks working alongside Sorbet? For classes like Packwerk::Configuration it might require a lot of stubbing to keep them working if we tried to use actual instances

Instead of stubbing `Packwerk::Offense` here, we could use a Fake Test Double, which we can pass and sorbet would be happy. That way we could get rid of this particular stub. The same should hold true for the RunContext and Configuration.

doodzik:
You created a bunch of great PRs! Thanks for contributing. I left two small comments, but other than this, it looks good.

cindygshopify:
is the `.stubs(:is_a?).with(X)` just to pass sorbet type checking?

test/unit/cli_test.rb
samuelgiles: Yeah good idea 👍

rochlefebvre (approved):
Excellent.

Issue #80, opened 2020-11-25
[Bug Report]


wildmaples:
**Description**
`bundle exec packwerk` and `bin/packwerk` produces different result. `bin/packwerk` loads files while `bundle exec` does not. 

**To Reproduce**
Try `bundle exec packwerk validate` and `bin/packwerk validate` 
`packwerk validate` requires files to be loaded, so it will fail on the `bundle exec` command. 

**Expected Behaviour**
`packwerk validate` requires files to be loaded, so it will fail on the `bundle exec` command. 

**Screenshots**

**Version Information**
 - Packwerk: v1.1.0

**Additional Context**
The solution is to move the file loading logic from `bin/packwerk` script into `ApplicationLoadPath`. `ApplicationLoadPath` is only used in `packwerk init` and `packwerk validate` commands, the two commands that require Rails to be booted. 

For additional context, see: https://github.com/Shopify/packwerk/pull/77#issuecomment-733284451

PR #94, opened 2020-12-11
Add logic to load application during load path extraction

samuelgiles:

To fix https://github.com/Shopify/packwerk/issues/80, an issue that prevents `bundle exec packwerk init` and `bundle exec packwerk validate` from working in the same way as `bin/packwerk init` & `bin/packwerk validate`.

It was suggested over in #80 that bringing the Rails loading logic in from `bin/packwerk` to `ApplicationLoadPath` would address this issue, I followed along these lines:

> The solution is to move the file loading logic from bin/packwerk script into ApplicationLoadPath. ApplicationLoadPath is only used in packwerk init and packwerk validate commands, the two commands that require Rails to be booted.

- `bin/packwerk` loads Rails by doing a fairly standard `require` for `config/environment`:
https://github.com/Shopify/packwerk/blob/373993a437aeca2223186617a95f09f2e57cee29/lib%2Fpackwerk%2Fgenerators%2Ftemplates%2Fpackwerk#L18
- `bundle exec packwerk init|validate` both end up calling `Packwerk::ApplicationLoadPaths.extract_relevant_paths` which checks if Rails is loaded, which it isn't because `bundle exec packwerk` is running `exe/packwerk`, this allows us to reach this assertion where we explode :
https://github.com/Shopify/packwerk/blob/19d40eadd70758eff6e9fbbab8d3ec69c84a03a4/lib/packwerk/application_load_paths.rb#L22
- We can swap out the assertion to instead load Rails; we can use `#root_path` from `Packwerk::Configuration` to get the path to `config/environment`.

- I guess there's maybe a question around the fact that Packwerk's CLI already `warn`ed people using `bundle exec packwerk` if `Rails` wasn't loaded that they should use `bin/packwerk` which did load `Rails`. So is running Packwerk via `bundle exec` with `exe/packwerk` something the project should continue to support? Or should this behaviour be deprecated and instead it should just point users at `bin/packwerk`?

## Type of Change

- [x] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist


test/unit/application_validator_test.rb
```diff
@@ -32,7 +32,10 @@ class ApplicationValidatorTest < Minitest::Test
 
     test "check_autoload_path_cache fails on extraneous config load paths" do
       use_template(:minimal)
-      ApplicationLoadPaths.expects(:extract_relevant_paths).returns([])
+      ApplicationLoadPaths
+        .expects(:extract_relevant_paths)
+        .with { |root_path| root_path.match?(%r{tmp\/fixtures}) }
```
rochlefebvre, line 37: Consider using [`ApplicationFixtureHelper.app_dir`](https://github.com/Shopify/packwerk/blob/19d40eadd70758eff6e9fbbab8d3ec69c84a03a4/test/support/application_fixture_helper.rb#L23) to get the absolute directory of your temporary app:
```ruby
.with { |root_path| root_path.match?(%r{#{app_dir}}) } # I have not tried this
```

samuelgiles: Ah cool, thanks @rochlefebvre 👍  Have moved it to using `.with(app_dir)`

lib/packwerk/application_load_paths.rb
```diff
@@ -8,18 +8,25 @@ module ApplicationLoadPaths
...
         assert_load_paths_present(relevant_paths)
         relative_path_strings(relevant_paths)
       end
 
-      sig { void }
-      def assert_application_booted
-        raise "The application needs to be booted to extract load paths" unless defined?(::Rails)
+      sig { params(root_path: String).void }
+      def load_application(root_path)
+        require File.expand_path(File.join(root_path, "config/environment"))
```
rochlefebvre, line 22: @wildmaples Now that ApplicationLoadPath loads the Rails app, does `bin/packwerk` still need the following line?
https://github.com/Shopify/packwerk/blob/373993a437aeca2223186617a95f09f2e57cee29/lib/packwerk/generators/templates/packwerk#L18

samuelgiles: Good idea - I've just checked against one of our applications and it can be removed from the `bin/packwerk` file 👍  I'll update it in the template now

samuelgiles: This has the nice effect of meaning `bin/packwerk check` no longer loads Rails

rochlefebvre, line 22: We're missing a second check for `Rails` after a successful `require`. At this point, we still don't know whether this is actually a Rails application.  How about?
```ruby
require File.expand_path(File.join(root_path, "config/environment"))
raise "Packwerk must be running in a Rails application" unless application_loaded?
rescue LoadError
  # ...
```

rochlefebvre (commented):
Apologies for the late follow up.  I believe we're missing one last check in `load_application`.  What do you think?

exterm:
🤔 

We need the application loaded for two things:
- the _actual_ application load paths (vs our cache on disk)
- the _actual_ application inflections

In terms of commands, this comes down to `init` (for only the load paths) and `check` (for both).

If we move the logic to load the application, does that mean that extracting the inflections in `check` works accidentally? It seems it would still work only if we extract the load paths first.

exterm:
Sorry, this is me essentially saying the solution proposed in the issue is wrong.

But it looks like a better solution would be to have a third thing that knows how to make sure the application is loaded, and call it everywhere where we need it.

That would be 

1. in ApplicationLoadPath
2. just before we call `ActiveSupport::Inflector.inflections` (currently in `ApplicationValidator`)

PR #99, opened 2021-01-11
bump packwerk version for new release v1.1.2

rochlefebvre:
Bump packwerk to v1.1.2 - See [Draft Release](https://github.com/Shopify/packwerk/releases) for the list of relevant changes

Tested on core using `dev packages check`, `dev packages update` (packwerk 1.1.1 and 1.1.2 report the same deprecated references), and `bin/spring packwerk validate`.
Tested on payment-service using `packwerk check`, `packwerk update` and `bin/packwerk validate`.

Are the release notes ok?  Running `bundle update` updates many gems in Gemfile.lock, so I did not do that. I manually incremented the version for `packwerk` instead, as it appears we've done in previous releases.

## Type of Change

- [ ] Bugfix
- [ ] New feature

## Checklist

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -85,7 +85,7 @@ GIT
 PATH
   remote: .
   specs:
-    packwerk (1.1.1)
+    packwerk (1.1.2)
       activesupport (>= 5.2)
       ast
       better_html
--- a/lib/packwerk/version.rb
+++ b/lib/packwerk/version.rb
@@ -2,5 +2,5 @@
 # frozen_string_literal: true
 
 module Packwerk
-  VERSION = "1.1.1"
+  VERSION = "1.1.2"
 end


doodzik:
The release notes look good to me. 

I would add that if `bundle update` is bumping many of our dependencies then maybe it is time to just bump them for this release. What do you think?

rochlefebvre:
> The release notes look good to me.
> 
> I would add that if `bundle update` is bumping many of our dependencies then maybe it is time to just bump them for this release. What do you think?

I hesistate to publish a new build right after updating so many dependencies.  How about we release this new version and _then_ I update the dependencies? That will give the new lineup a few weeks of soak time before the next one.

doodzik:
Sounds good to me 👍

doodzik (approved):

gmcgibbon (approved):
`bundle update` isn't important for gem releases. The `Gemfile.lock` isn't used when installing the gem. What's important is the constraints you specify in the gem's gemspec. It is up to the consuming app to update its own dependencies.

exterm (approved):

PR #100, opened 2021-01-15
add violation_count to deprecated_references.yml

mclark:

In order to improve our metrics around reference violations, we would like to be able to view the actual violation count for each constant, rather than just the number of files that contain violations. A single file could conceivably contain a hundred reference violations, but that impossible to determine without parsing the file either via script or visually.

Adding a violation count per constant is a relatively unobtrusive way to get this count that should not interfere or overly complicate the core functions of packwerk. With this number included in `deprecated_references.yml`, the granularity of the monitoring for referential integrity of each package is improved and progress toward detangling dependencies can be better tracked.

I chose to just add another key/value pair to the package violations hash. Now, in addition to `"files"` and `"violations"`, we also have `"violation_count"`. This is an additive change to the existing structure of the `deprecated_dependencies.yml` file that is unlikely to affect consumers' expectations for the file's structure.

On [another branch](https://github.com/Shopify/packwerk/compare/main...mclark:ref-violation-counts-per-file) I also wrote a _per file_ violation count which **does** alter the structure of the `deprecated_dependencies.yml` but I thought I'd hold off on that for now to see if we truly need it at GitHub or if there is interest elsewhere in the community.

There's just one line of production code change, so I guess just focus on ensuring this is an accurate reflection of the violation count for the constant. No new tests were added but existing tests have been adapted in a way I believe sufficiently covers the change. Maybe there are other scenarios that should be tested which I did not consider?

## Type of Change

- [ ] Bugfix
- [x] New feature

## Checklist

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -275,6 +275,7 @@ components/merchant:
     - dependency
     files:
     - components/merchant/app/public/merchant/generate_order.rb
+    violation_count: 3
 ```
 
 Above is an example of a constant violation entry in `deprecated_references.yml`.
@@ -283,5 +284,6 @@ Above is an example of a constant violation entry in `deprecated_references.yml`
 * `::Checkouts::Core::CheckoutId` - violated constant in question
 * `dependency` - type of violation, either dependency or privacy
 * `components/merchant/app/public/merchant/generate_order.rb` - path to the file containing the violated constant
+* `violation_count` - the number of distinct reference violations for this constant, accounting for the fact a single file can have multiple violations for the same constant
 
 Violations exist within the package that makes a violating reference. This means privacy violations of your package can be found listed in `deprecated_references.yml` files in the packages with the reference to a private constant.
--- a/lib/packwerk/deprecated_references.rb
+++ b/lib/packwerk/deprecated_references.rb
@@ -95,6 +95,7 @@ module Packwerk
       @new_entries.each do |package_name, package_violations|
         package_violations.each do |_, entries_for_file|
           entries_for_file["violations"].sort!.uniq!
+          entries_for_file["violation_count"] = entries_for_file["files"].count
           entries_for_file["files"].sort!.uniq!
         end
         @new_entries[package_name] = package_violations.sort.to_h
--- a/test/unit/deprecated_references_test.rb
+++ b/test/unit/deprecated_references_test.rb
@@ -173,7 +173,11 @@ module Packwerk
     test "#add_entries and #dump adds constant violation to file in the appropriate format" do
       expected_output = {
         "buyers" => {
-          "::Checkout::Wallet" => { "violations" => ["privacy"], "files" => ["some/violated/path.rb"] },
+          "::Checkout::Wallet" => {
+            "violations" => ["privacy"],
+            "files" => ["some/violated/path.rb"],
+            "violation_count" => 1,
+          },
         },
       }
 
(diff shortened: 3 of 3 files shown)


exterm (commented):
This is an interesting idea with a neatly simple implementation.

For context: In the early days of packwerk we listed each violation separately, with the line it occurred on in the code. However, that is very brittle, so we reduced granularity to _per file_. We had not considered the possibility of persisting the count.

I'm leaning towards merging this, however...

My one concern is that this may make packwerk more confusing for people who are not intimately familiar with it. the current design is based on the assumption that there is high cohesion within one file, so that it makes sense that access to a constant is either allowed or disallowed for the whole file. Someone modifying a file may not immediately understand why an existing reference may be OK, but the one that they just added is not.

I'm interested in other people's (especially user's) opinions on this.

wildmaples:
Yes I am also concerned that this complicates the `deprecated_references.yml` file as it is confusing enough for people who sees the file as a blocker to their work. 

>I'm interested in other people's (especially user's) opinions on this.

I am not a user of Packwerk but on the Modular Monolith team (rip) here at Shopify, we were definitely interested in the number of violation count as a metric. We used a script that parses on all the deprecated references files to attain that number for the packages. So I think having an easy way to access some metrics via Packwerk directly would be useful. 

However, I think only the right people will see a benefit in having some metrics, but not all (i.e. not all developers have to care about this metric and will be confused when they see it). 

To avoid confusion and potentially over complicating the deprecation file, I am suggesting that we perhaps have this feature into a command line output? Perhaps something like `packwerk generate metrics` that outputs a metrics to a yaml file.

mclark:
👋 @wildmaples, @exterm thanks for taking a look at this PR!

Yeah I share your concerns that this is not strictly a necessary bit of information and could potentially further confuse users. @wildmaples I _really_ like your idea of adding new command line option to generate metrics. A tool like packwerk is typically going to be used by organizations that do care about metrics, but these numbers are not necessary to fulfill the core functions of dependency violation tracking. Having it as an additional option should work well with both those constraints.

My only concern remaining is that it will require an additional run of packwerk to calculate all the metrics, but I guess that's just unavoidable. I suppose if we really wanted to, we could use the existing `deprecated_references.yml` files as a cache to limit the set of files that get parsed, but I don't know the source code well enough to comment on the feasibility of that idea.

wildmaples:
> My only concern remaining is that it will require an additional run of packwerk to calculate all the metrics, but I guess that's just unavoidable. I suppose if we really wanted to, we could use the existing deprecated_references.yml files as a cache to limit the set of files that get parsed, but I don't know the source code well enough to comment on the feasibility of that idea.

Apologies for the late response, @mclark I no longer work on the Packwerk team.  

You make a good observation and I can see two use cases here:

1) A new codebase, with all boundaries enforced to begin with. This means all `deprecated_references.yml` files are up-to-date. In this case, we should be able to treat the `deprecated_references.yml` files as a cache as you mentioned and get metrics on the first run. 

2) A legacy codebase, where not all packages have enforced boundaries. In this case, I'd foresee that folks who care about modularity metrics would care about two types of metrics: 

    * First, the **true state** of the codebase, meaning what would the metrics be if _all boundaries_ between packages were enforced. In this first case, we would need to run Packwerk on the code base twice: once to get the updated list of deprecated references, and then to parse that information. 
    * Second, the **current state** of the codebase. In this case, we would just parse over the existing deprecated references file, similar to point 1 above. 

I think it is likely the majority of people fall into the second category of use case, such as Shopify Core. 

As this is just a proposal for a new feature and I no longer work directly on Packwerk, I don't have any strong convictions and I'll wait for other Packwerk users to chime in.

PR #109, opened 2021-03-31
Parallel packwerk [discussion only - DO NOT MERGE]

mclark:
This PR makes Packwerk about 3x faster by introducing parallelism.

Packwerk currently iterates through all the matching files in sequence, mutating a `ReferenceLister` instance to track relevant violations. This PR introduces a dependency on the [Parallel gem](https://github.com/grosser/parallel) to allow Packwerk to process these files in parallel. As it fans out to multiple subprocesses in order to accomplish this, we cannot mutate a shared variable, thus necessitating the extensive refactoring seen in this PR.

I ended up making the file processing logic more functional in nature. The reference lister is removed entirely and instead each reference in each file is just evaluated according to each of the configured checkers. If there is a violation, we now create a new `ReferenceViolation`, a subclass of `Violation` which contains a `Reference` and `ViolationType`. These violations are returned to the parent process and added to a list of `all_violations`.
Now that the new reference violations have the additional data we have enough information to filter them at a later stage so we can exclude violations that are already cached or ones that aren't reference type.

I believe this more functional approach will be simultaneously easier to understand as well as parallel ready.

This is a proof of concept PR, not something we should ever consider merging. This demonstrates the broad strokes of how things could be structured and doesn't try too hard on naming or avoiding technical debt. Those discussions can wait for the real PRs that this one will be broken down into. Right now, we just need to discuss whether there is an appetite for a significant refactor such as this with the end goal of introducing parallel execution out of the box.

## Type of Change

- [ ] Bugfix
- [ ] New feature

## Checklist

cc @exterm for 👀


exterm:
Matt and I talked through this yesterday, and I think the refactor makes sense.

Specifically, removing the `ReferenceLister` interface is a good thing; it was left over from the days when `packwerk` was still based on `rubocop`. It was an abstraction that never worked really well.

So, I am confident it makes sense to merge the refactor.

As for the behavior change, to make packwerk execute in parallel - I think the functionality is valuable.

Could we run into any problems by making parallel execution the default?

By the way - once the tests are green on this I'd be super curious to run it on shopify's core monolith.

exterm:
you should be able to cherry-pick this https://github.com/Shopify/packwerk/pull/111

exterm:
Just ran `packwerk update-deprecations` on our core monolith using this branch.

Good news: It works! There are some ordering issues (I wonder if there's a `sort` missing somewhere) but the results are as expected.

It is a great heat source too, using my MacBook to its full capacity.

A full run was reported by packwerk as having taken 91 seconds vs 450 seconds with the latest release. I'm running on a 16 inch MacBook with Ruby 2.7.2p137.

exterm:
> Do we know what is the bottleneck for parallelization? Do we spend time waiting IO or CPU?

From earlier exploration we know that packwerk spends >80% of run time in parsing - Tom's incomplete prototype using `RubyVM::AbstractSyntaxTree` as an alternative parser yielded a 5x speedup. I assume if switching out the parser library gives that kind of speedup the bottleneck is CPU. A different parser is unlikely to do less IO. I don't know for sure though.

exterm:
> I think we can do well with DRB and forking ourselves.

I talked through the added `parallel` dependency briefly with Matt and he said he gets an advantage from not having to handle any inter-process communication, thanks to the parallel `flat_map` implementation. It makes it really easy to do the "reduce" step of this, essentially map/reduce like pattern. I like that, but I'm also not quite sure what an implementation without the gem would look like.

We don't technically _need_ communication between processes, as the existing parallelization through a wrapper script in shopify's CI shows. However, for the `update-deprecations` command specifically the complexity of the implementation would increase a lot; we'd have to parallelize per package instead of per file, and since packages can be nested (and are by default), that's a bit of a gnarly problem.

mclark:
Hi @rafaelfranca! Thanks for taking a look at this PR. I think @exterm has already addressed your concerns much as I would, thanks for that @exterm!

The work is definitely CPU constrained as my profiling observed similar numbers as those described above. I am struggling to conceive of a feasible implementation for `update-deprecations` without the Parallel gem. We need to do the map-reduce fanout somehow which necessitates the serialization/marshalling process. Rolling our own implementation is a possibility but it feels like a fair bit of logic that has already been implemented in other gems. If we could run all this logic without the map/reduce then it might be feasible but would introduce its own set of new complications and risks. It all comes down to how much we want to avoid this dependency.

If it helps, this same dependency is used by [Rubocop](https://github.com/rubocop/rubocop/blob/6de9f1ee51c058c04e5883cef6c1071c36a01e93/rubocop.gemspec#L33) for a similar purpose, which was the inspiration for this approach.

Do these comments address your concerns @rafaelfranca?

mclark:
https://github.com/Shopify/packwerk/pull/127 has been merged and since it covers the rest of what was in this PR I think we can close this one. 🎉

PR #110, opened 2021-03-31
update dependencies

exterm:
- use Rails 7
- use Ruby 3.0 as default
- test against `ruby: [2.6, 2.7, 3.0]` and `rails: [6.0, main]`

Packwerk is affected by the `mimemagic` dilemma like so many other projects, since the mimemagic version we were using was pulled. I used this as a trigger to update all our dependencies.

### Followup

- [ ] change branch protection rules to require new CI steps


test/unit/file_processor_test.rb
```diff
@@ -46,20 +46,20 @@ class FileProcessorTest < Minitest::Test
...
+      @node_processor.expects(:call).with do |node, ancestors|
         parent = ancestors.first # class Hello; world; end
         Node.constant?(node) && # Hello
           Node.constant_name(node) == "Hello" &&
           ancestors.length == 1 &&
           Node.class?(parent) &&
           Node.class_or_module_name(parent) == "Hello"
       end
-      @node_processor.expects(:call).with do |node, ancestors:|
+      @node_processor.expects(:call).with do |node, ancestors|
```
exterm, line 62: This threw a `ArgumentError: missing keyword: :ancestors` with Ruby 3.0. I assume there's a bug with mocha and Ruby 3, but I really don't want to go down that rabbit hole right now so I just changed the method to not use keyword arguments. I don't think it significantly degrades the design.

I added a sorbet signature to make up for it 😉

test/unit/cli_test.rb
```diff
@@ -18,34 +18,26 @@ class CliTest < Minitest::Test
...
-    end
-
-    class FakeRunContext < Packwerk::RunContext
-      def initialize; end
-    end
-
-    class FakeConfiguration < Packwerk::Configuration
-      def initialize; end
-    end
-
```
exterm, line 32: new style guide rules rightly rejected this; it doesn't initialize the objects correctly. The fix was a little more involved but I discovered that actually non of these fake classes are necessary.

exterm:
This turned out to be quite the rabbit hole. I'd advise reviewers to go through commit by commit.

exterm:
I am thinking about deconstructing this into multiple PRs.

exterm:
I investigated but...

- Rails 7 needs at least Ruby 2.7
- I want to test against the two most recent major Ruby versions, so it is 3.0 and 2.7 then
- lots of gems are incompatible with Ruby 3.0 in their old versions, so the y need to be updated

everything else follows from that. So I think we should merge this PR as is.

exterm:
minimal fix for just mimemagic lives in https://github.com/Shopify/packwerk/pull/111

.github/workflows/ci.yml
```diff
@@ -8,8 +8,8 @@ jobs:
     strategy:
       matrix:
         ruby:
-          - 2.6
```
rafaelfranca, line 11: We need to change the gemspec to require Ruby >= 2.7, but Ruby 2.6 is still supported at Shopify so I think we should continue to support here.

.github/workflows/ci.yml
```diff
@@ -29,7 +29,7 @@ jobs:
       - name: Set up Ruby 2.6
```
rafaelfranca, line 29: This is now wrong

exterm: ugh right - good catch. Not sure why the version needs to be in the name anyway

.github/workflows/ci.yml
exterm: That would mean that I need to test against two different Rails versions, as Rails 7 doesn't support Ruby 2.6.

exterm:
Fixed everything @rafaelfranca @gmcgibbon 

The only thing I am not happy about is the size of the changeset, but I haven't managed to significantly reduce it for the reasons lined out in https://github.com/Shopify/packwerk/pull/110#issuecomment-811255851

.github/workflows/ci.yml
exterm: Fixed this by setting up a proper build matrix and multiple Gemfiles.

PR #111, opened 2021-03-31
use a mimemagic version that wasn't pulled

exterm:
Builds are currently broken on master because the mimemagic version we're using has been pulled.

Update mimemagic - this is the simplest possible fix.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -133,7 +133,9 @@ GEM
     marcel (0.3.3)
       mimemagic (~> 0.3.2)
     method_source (1.0.0)
-    mimemagic (0.3.5)
+    mimemagic (0.3.10)
+      nokogiri (~> 1)
+      rake
     mini_mime (1.0.2)
     mini_portile2 (2.4.0)
     minitest (5.14.0)


gmcgibbon (changes requested):
We can't use mimemagic anymore because it is GPL. Can you update rails instead? A release was pushed 4 days ago to amend this.

exterm:
> We can't use mimemagic anymore because it is GPL. 

I thought that just applied to the pulled releases?

😢 

I have a Rails update in https://github.com/Shopify/packwerk/pull/110, but it is a much larger changeset.

exterm:
current `mimemagic` has MIT license. https://github.com/mimemagicrb/mimemagic/blob/master/LICENSE

exterm:
(as does the version used in this PR, `0.3.10`) https://github.com/mimemagicrb/mimemagic/blob/v0.3.10/LICENSE

gmcgibbon:
The package still depends on `shared-mime-info` which is GPL IIRC... I'm not sure what that means 🤔

gmcgibbon:
It seems easier to just bump rails if we're going to have to do that anyway.

exterm:
Bumping Rails is Definitely not as easy as I thought 😞

exterm:
> The package still depends on `shared-mime-info`

Does it? That dependency doesn't show either in my lock file or in https://github.com/mimemagicrb/mimemagic/blob/v0.3.10/mimemagic.gemspec

PR #116, opened 2021-04-28
Normalize parse commands into single class

mclark:
In https://github.com/Shopify/packwerk/pull/109 there is quite a bit of repeated logic between the three commands that parse the subject files for reference violations (`update_deprecations`, `check`, and `stale_violations`). Normalizing all of this logic now reduces the possibility of bugs and makes the code base smaller and more testable. Also, for future refactoring code related to parallelization, this change should make those changes more readable in PR form.

I brought the three common parsing commands into a single `ParseRun` class. As each command essentially differs only in how the result is interpreted, it made sense to just extract that part from each legacy command.

* The `check` command was never properly extracted and has special handling for interrupts. As this is now handled in the same code that handles the other two commands we need to ensure no new bugs are introduced at that level.
* This change is purely restructuring and adds no new tests. Existing tests have been interpreted and written to conform to the new code interface.

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist


mclark:
I have verified this produces identical output to the `main` branch against the github monolith for both `check` and `update-deprecations`.

PR #122, opened 2021-05-25
Use autoload instead of explicit requires

rafaelfranca:
This has two goals:

1. Remove the need to use `require: false` in the Gemfile given all the code will be lazily evaluated.
2. Will avoid the mistake of requiring a file in the tests but not in the code, since all autoload are setup in the same place.

The only disadvantage is that is a file is removed but the autoload is not nothing fails.

### DIFF
--- a/lib/packwerk.rb
+++ b/lib/packwerk.rb
@@ -3,7 +3,6 @@
 
 require "sorbet-runtime"
 require "active_support"
-require "constant_resolver"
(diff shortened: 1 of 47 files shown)


mclark (approved):
This is a great idea. Since Packwerk is already reliant on rails anyway, why not leverage that dependency to make it a bit more resilient? LGTM.

PR #123, opened 2021-05-27
normalize Reference creation in tests

mclark:
Creating references in tests is a fairly common pattern and includes a fair bit of boilerplate. This PR introduces a new module with a method to quickly build these references with support for a few common overrides. Normalizing this logic into a helper method reduces the surface area for upcoming changes for [parallelization](https://github.com/Shopify/packwerk/pull/109). Also, by only needing to specify key parts of the Reference data, it should draw attention to the code that matters most for the given test.

A helper method with sensible defaults is a common pattern that is easy to understand and maintain.

I guess the biggest risk is that the meaning of a test might have been altered. Pay particular attention to any tests that changed anything other than `Reference` initialization.

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist

### DIFF
--- a/test/unit/dependency_checker_test.rb
+++ b/test/unit/dependency_checker_test.rb
@@ -19,7 +19,6 @@ module Packwerk
     setup do
       setup_application_fixture
       use_template(:skeleton)
-      @destination_package = Package.new(name: "destination_package", config: {})
       @reference_lister = CheckingDeprecatedReferences.new(app_dir)
     end
 


test/unit/dependency_checker_test.rb
```diff
@@ -29,78 +30,34 @@ def listed?(_references, violation_type:)
...
       checker = dependency_checker
-
-      reference =
-        Reference.new(
-          source_package,
-          "some/path.rb",
-          ConstantDiscovery::ConstantContext.new(
-            "::SomeName",
-            "some/location.rb",
-            @destination_package,
```
rafaelfranca, line 40: Now `@destination_package` is not needed in this test anymore.

mclark: Wow, impressive spotting! I've removed the `@destination_package` ivar and things seem fine 👍

PR #124, opened 2021-05-31
Functional codebase parsing

mclark:
This is the "meatiest" of the changes needed to enable parallelization of Packwerk. Currently, packwerk mutates a `ReferenceLister` object to add the violations that it encounters while parsing the codebase. This object is passed all the way down to the `NodeProcessor` level. There are a couple problems with this:

1. `ReferenceLister#listed?` is overloaded in a surprising way and actually mutates data in the `CacheDeprecatedReferences` implementation.
2. These mutated instances are lost when the subprocesses exit in a parallel world.

This pull request instead makes the code parsing process functional (aside from progress output) and moves the filtering for new violations happen after the processing is done.

Having this code be more functional should help with future improvements as there is fewer dependencies involved in the parsing stage. This path was chosen as it is necessary to make packwerk parsing run in parallel as described in https://github.com/Shopify/packwerk/pull/109.

I tried to make this change as small as possible as the changes are subtle and I fear the regression potential. Each commit in this PR should be able to be evaluated in isolation to help improve the narrative for reviewers. Reviewers should definitely give this one a run on their own monolith code base to ensure there are no unexpected behaviours. This pull request will expose gaps in the tests if they are there.

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist


lib/packwerk/parse_run.rb
```diff
@@ -76,5 +84,20 @@ def run(reference_lister)
...
+      if failed
+        @progress_formatter.mark_as_failed
+      else
+        @progress_formatter.mark_as_inspected
+      end
+    end
+
+    def includes_unlisted_offense?(offenses, reference_lister)
+      offenses.any? do |offense|
+        next true unless offense.is_a?(ReferenceOffense)
```
rafaelfranca, line 98: This same block is now duplicated inside the same called `filtered_offenses`. Should not it be possible to execute this block inside `find_offenses` so we don't need to call it twice?

mclark: Want me to remove the second filtering stage for now and reintroduce it in the next PR when parallel is brought in?

lib/packwerk/parse_run.rb
```diff
@@ -49,23 +49,30 @@ def update_deprecations
...
       @progress_formatter.started(@files)
 
-      run_context = Packwerk::RunContext.from_configuration(@configuration, reference_lister: reference_lister)
+      run_context = Packwerk::RunContext.from_configuration(@configuration)
       all_offenses = T.let([], T.untyped)
       execution_time = Benchmark.realtime do
         @files.each do |path|
           run_context.process_file(file: path).tap do |offenses|
-            mark_progress(offenses: offenses, progress_formatter: @progress_formatter)
+            includes_unlisted = show_errors && offenses.any? { |offense| unlisted?(offense, reference_lister) }
```
rafaelfranca, line 74: This is still iterating over the same list twice. Can we filter here and remove the `select` from `filtered_offenses?`. Something like:

```ruby
offenses = offenses.select { |offense|  unlisted?(offense, reference_lister) }
includes_unlisted = show_errors && !offenses.empty?
update_progress(failed: includes_unlisted)
all_offenses.concat(offenses)
```

I don't think `find_offenses` is called in a context where we we want the ones that are listed.

mclark: Yeah you are right in the context of this PR, did you see my [comments above](https://github.com/Shopify/packwerk/pull/124#discussion_r642648570)? Basically I can pull the extra iteration now but it'll be coming back pretty soon once parallel is in.

rafaelfranca (approved):

PR #125, opened 2021-06-08
introduce OffenseCollection

mclark:
finding the offenses for a `ParseRun` now returns an `OffenseCollection` object.
This refactor is made possible by amalgamating the `CacheDeprecatedReferences`
class and all its subclasses into a single `OffenseCollection` class.

`OffenseCollection`s know about the different between errors and reference
violations which is something that was handled implicitly before. We now
no longer need to mutate the new entries in a method called `listed?`
just to satisfy an interface.

This change makes it much more clear what each method is really doing
and simplifies the code for future readers.

More consolidation and simplification to continue making the parsing process behave in a more functional manner so that the parallel implementation is clear and concise.

I decided to consolidate this class hierarchy into a single class because each sub class only ever added a single method. This was previously necessary because of the necessary adherence to (and abuse of) the `ReferenceLister` interface. Now that adding files to an `OffenseCollection` and asking if an offense is listed are different methods everything can live in one class.

Now that `OffenseCollection` exists, it can be the object returned from parsing all the files instead of it being passed in as an argument and mutated.

I struggled with naming but in the end I think `OffenseCollection` captures the intent fairly well. I also struggled with calling non-reference offenses "errors" but that is generally what they are. I may follow that up with an `ErrorOffense` at some point and make `Offense` an interface. That can wait though.

Please test this change on your own monolith. I didn't see any changes vs `main` on the `github/github` monolith but it's good to be certain.

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist


rafaelfranca:
Yeah. I agree we can change that later. I don't have any suggestion on how to fix this problem right now though. There are two comments in the test to use `assert_predicate`, otherwise, looks good to me.

mclark:
Thanks for the super speedy reviews @rafaelfranca! Suggestions have been merged in 👍

rafaelfranca (approved):

PR #126, opened 2021-06-09
Hash id for packages

mclark:

If we add two violations for the same _logical_ package, then they should be added to the same `DeprecatedReference` instance. In a single threaded environment, it's always the same package instance, so we don't have these hash misses, but with multiple processes marshalling data, we have a very high likelihood of these misses and thus multiple deprecated reference instances for the same logical package. The consequence of this is incorrect data being written when update-deprecations is run.

I implemented the missing methods that Ruby needs to hash objects accurately.

This is pretty basic stuff, so test coverage and acceptability of making some private methods public I guess.

## Type of Change

- [x] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist


rafaelfranca:
I'm not sure about the two first commits. Are we able to test the hash id without exposing those methods only for tests purpose?

mclark:
> Are we able to test the hash id without exposing those methods only for tests purpose?

Yes, but it requires a full integration test and writing to disk unfortunately. I don't like them either but I also don't want full integration tests. I need to give this a bit more thought, maybe hold off on the merge a bit and I will see if there is any other reasonable path forward.

tomstuart:
I took a look at this but couldn’t see a straightforward way for us to write the test without touching the filesystem, mainly because `Packwerk::DeprecatedReferences` doesn’t have a relevant public API beyond “dump the new entries to disk”. Unless I’m missing something obvious, I think our immediate options are to either:

* bite the bullet and write the acceptance test (with mocked-out IO if that’s the main fear); or
* refactor `Packwerk::DeprecatedReferences` to have fewer responsibilities so that it can collaborate with a separate object whose job is to write out the deprecations file, and unit test this PR at the resulting API boundary.

The latter is more work than is perhaps warranted for such a simple fix, although improving the internal design would contribute to the ease of long-term maintenance.

tomstuart:
What I’m implicitly saying above is that I agree we shouldn’t increase the public API surface area for the sole purpose of supporting tests. But if we can actively improve the API design to achieve the same goal then that’s a different situation.

mclark:
Thanks for your thoughts @tomstuart, they are appreciated 🙇 

I'm going to spell out what I _think_ is an exhaustive list of our options here along with my preference at this point in the story:

* Refactor `Packwerk::DeprecatedReferences` so that it provides the "data to write" to a caller which would handle writing to disk.
* Take on some serious tech debt and use `instance_variable_get` with wild abandon.
* Grow the public interface as is happening in this PR.
* Write a full integration test.
* Just test that `Package` hashes properly and leave the behaviour of `OffenseCollection` as-is for now.

My preference is to not go too deep on this. Unless there is any objection, I'll just go with the last two options above and abandon the growth of the public api. The big refactor would be ideal in the long term, but I'd like to not grow these changes even more, especially as we're so close to having the feature change merged.

rafaelfranca (approved):

tomstuart (approved):
Thanks @mclark. The test is useful and I agree this is an appropriate solution to get a fix in now.

PR #127, opened 2021-06-10
Parallel Packwerk execution

mclark:
As described in the original [spike](https://github.com/Shopify/packwerk/pull/109), this lets packwerk processing run in parallel, using the [Parallel gem](https://github.com/grosser/parallel). 

The `parallel` gem was chosen over other options as discussed over in https://github.com/Shopify/packwerk/pull/109.

* I left `parallel` to choose the best number of processes based on the machine it is running but this could reasonably also be a configuration parameter instead.
* There is only one test that explicitly runs in parallel although many others execute the parallel code implicitly as it is the default. The single parallel test just verifies that the result output from the execution is as expected. It can't verify the output sent to the `StringIO` as that is handled in a subprocess so the result was deemed sufficient.
* When running in parallel we can no longer interrupt the process and get any meaningful output as we are using a `flat_map` instead of gradually appending to a list. I can't think of a way to support interrupt and still get the partial data so I view this as a necessary sacrifice for the faster execution unless someone else has any ideas. I've never used this particular feature but maybe others are concerned?
* Parallelism is enabled by default so that new users get it without needing to "discover" it. We could conceivably set this to false by default initially while we shake out any potential bugs.

## Type of Change

- [ ] Bugfix
- [x] New feature

### Additional Release Notes

## Checklist


rafaelfranca:
Just tested in our application.

### Before

```
📦 Finished in 553.96 seconds

No offenses detected 🎉
bin/spring packwerk check  1.40s user 0.31s system 0% cpu 9:23.80 total
```

### After

```
📦 Finished in 197.96 seconds

No offenses detected 🎉
bin/spring packwerk check  1.49s user 0.27s system 0% cpu 3:50.73 total
```

rafaelfranca:
Could be Ruby 3.0.0 as well, let me upgrade to 3.0.1.

rafaelfranca:
I'm going to merge this so I can release so more people can try it.

rafaelfranca:
1.2.0 released with this on.

PR #128, opened 2021-06-11
Remove the need for a custom binstub or test

rafaelfranca:

The first thing done in this PR was to allow packwerk to boot the Rails application only when it is needed.

This made possible to move the concern of how to boot the application from the binstub to inside Packwerk.

This will also allow applications to not need to be booted to run the all packwerk commands, removing the need to use Spring.

The breaking change is that non-rails applications that run `packwerk init` or `packwerk create_configs` will see a failure, but in reality that never worked anyway, the error was just different. 

## Type of Change

- [ ] Bugfix
- [x] New feature

### Additional Release Notes

## Checklist

- [ x I have added tests to cover my changes.


doodzik (approved):
LGTM. Thanks for splitting the changes up into so many smaller commitse It made it really straight forward to review.

lib/packwerk/application_load_paths.rb
```diff
@@ -54,6 +49,19 @@ def relative_path_strings(paths, rails_root: Rails.root)
           .uniq
       end
 
+      private
+
+      sig { params(root: String).void }
+      def assert_application_booted(root)
```
mclark, line 55: The name and implementation of this method seem a bit mismatched now. What do you think of something like this?

```ruby
  def require_application(root)
```

rafaelfranca: Done

rafaelfranca:
Now it is a matter to decide which version I release. I don't want to release a 2.0 since it is too small of a change, but it is somewhat breaking if users were using the spring command, but I'm also not sure how many users used that command in the first place. Everything else is the same. I guess I'll ask some people at Gusto to see if they use the spring command.

lib/packwerk/application_load_paths.rb
```diff
@@ -8,20 +8,15 @@ module ApplicationLoadPaths
     class << self
       extend T::Sig
 
-      sig { returns(T::Array[String]) }
-      def extract_relevant_paths
-        assert_application_booted
+      sig { params(root: String, environment: String).returns(T::Array[String]) }
+      def extract_relevant_paths(root, environment)
+        require_application(root, environment)
```
exterm, line 13: I think the application technically also needs to be booted when extracting / comparing inflections ( `InflectionsFile` generator and `check_inflection_file` in `ApplicationValidator`).

I assume this works because things happen to be executed in the right order.

PR #130, opened 2021-06-18
Fix support for constants under the public path of the root package

tjwp:

Constants in the public folder of the root package are not treated as public. This is due to their location not being recognized as public.

The constant location that is used to check if the constant is public is relative to the root of the project. That path is checked against the configured public path for a package to determine if the constant is public.

For all packages, the public path that is used for this check is prefixed by the package name. For the root package that is ".".

The constant location is relative and not prefixed by "./" and therefore never matches the root package's public path used for the check (default for the check would be "./app/public/").

There are multiple ways of approaching this, but I chose to use the public path without the package name prefix when testing if a path is a public path for the root project. This kept the change small and avoids modifying the `#public_path` method or the `#package_path?` check which is working.

There is added test coverage for the root package to make the behavior more obvious.

The change is pretty small, but is there another way we'd like to implement it?

Are constants in the public folder of the root package expected to be supported?

## Type of Change

- [x] Bugfix

### Additional Release Notes

- [x] Bugfix: fix support for constants under the public path of the root package

## Checklist

### DIFF
--- a/lib/packwerk/package.rb
+++ b/lib/packwerk/package.rb
@@ -71,11 +71,5 @@ module Packwerk
     def root?
       @name == ROOT_PACKAGE_NAME
     end
-
-    private
-
-    def unprefixed_public_path
-      @unprefixed_public_path ||= user_defined_public_path || "app/public/"
-    end
   end
 end
--- a/lib/packwerk/package.rb
+++ b/lib/packwerk/package.rb
@@ -33,11 +33,15 @@ module Packwerk
     end
 
     def public_path
-      @public_path ||= File.join(@name, user_defined_public_path || "app/public/")
+      @public_path ||= File.join(@name, unprefixed_public_path)
     end
 
     def public_path?(path)
-      path.start_with?(public_path)
+      if root?
+        path.start_with?(unprefixed_public_path)
+      else
+        path.start_with?(public_path)
+      end
     end
 
     def user_defined_public_path
@@ -66,6 +70,12 @@ module Packwerk
 
     def root?
       @name == ROOT_PACKAGE_NAME
+    end
+
+    private
+
+    def unprefixed_public_path
+      @unprefixed_public_path ||= user_defined_public_path || "app/public/"
     end
   end
 end
--- a/lib/packwerk/package.rb
+++ b/lib/packwerk/package.rb
@@ -33,7 +33,15 @@ module Packwerk
     end
 
     def public_path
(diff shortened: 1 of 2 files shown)


lib/packwerk/package.rb
```diff
@@ -33,11 +33,15 @@ def package_path?(path)
     end
 
     def public_path
-      @public_path ||= File.join(@name, user_defined_public_path || "app/public/")
+      @public_path ||= File.join(@name, unprefixed_public_path)
```
rafaelfranca, line 36:
```suggestion
      @public_path ||= begin
        unprefixed_public_path  = user_defined_public_path || "app/public/"

        if root?
          unprefixed_public_path
        else
          File.join(@name, unprefixed_public_path)
        end
      end
```

lib/packwerk/package.rb
```diff
@@ -33,11 +33,15 @@ def package_path?(path)
...
     end
 
     def public_path?(path)
-      path.start_with?(public_path)
+      if root?
+        path.start_with?(unprefixed_public_path)
+      else
+        path.start_with?(public_path)
+      end
```
rafaelfranca, line 44:
```suggestion
      path.start_with?(public_path)
```

lib/packwerk/package.rb
```diff
@@ -67,5 +71,11 @@ def to_s
     def root?
       @name == ROOT_PACKAGE_NAME
     end
+
+    private
+
+    def unprefixed_public_path
+      @unprefixed_public_path ||= user_defined_public_path || "app/public/"
+    end
```
rafaelfranca, line 79:
```suggestion
```

test/unit/package_test.rb
```diff
@@ -36,6 +36,17 @@ class PackageTest < Minitest::Test
       assert_equal("components/timeline/my/path/", package.public_path)
     end
 
+    test "#public_path returns expected path when using the default public path in root package" do
+      package = Package.new(name: ".", config: {})
+      assert_equal("./app/public/", package.public_path)
```
rafaelfranca, line 41: I don't think the public path starting with `.` is expect or even correct.

rafaelfranca (changes requested):

test/unit/package_test.rb
tjwp: I've fixed up the test coverage that I added and nothing else appears to break with the suggested changes.

rafaelfranca:
Great finding. We don't have any root package in the main app so we never found this issue.

Issue #129, opened 2021-06-15
Introduce command that combines `check` and `detect-stale-violations`


mclark:
Currently in our monolith we are running `packwerk check` in our CI. This is great and helps us manage our technical in a controlled manner. However, if a team removes some deprecated references and forgets to run `packwerk update-deprecations` there is no CI failure and the associated `deprecated_references.yml` file just gets a bit out of date with reality since we don't run `packwerk detect-stale-violations` in CI.

This only becomes a problem when other people start running `packwerk update-deprecations` and suddenly multiple PRs have the same line removals in them.

I propose one of the following ideas and I'm looking for thoughts and feedback about which approach might be most palatable:

* Update `packwerk check` to also detect stale violations
* Add `packwerk check --include-stale` to optionally change its return code based on stale violations
* Add a new command `packwerk full-check` or something similar that checks both

What do folks think? Is this something that others would find helpful?

mclark:
Yes, we can definitely run the other check as a separate CI but it just seems wasteful to me. At scale, that's a lot of processing that doesn't really need to happen, we can probably make a worthwhile positive environmental impact by combining these two.

rafaelfranca:
I agree, but at the same time if we combine both commands in our CI it would take more time than running all tests of the app, I'm surprised this is not the case in your app as well.

I'm ok with the inclusion of a new command though, but I'd prefer to change check to do both. Can you see how much the total time would increase if we change `check` to do both? If it is a lot we will need to create another command.

PR #131, opened 2021-06-23
Combined check and detect-stale-violations

mclark:
Currently, `packwerk check` just checks for new violations in the codebase. If we want to also check for stale violations we need to do that as an additional parse of the entire codebase which is quite expensive. This PR makes `packwerk check` check for _new_ **and** _stale_ violations.

I opted to change the behaviour of `packwerk check` after consultation with @rafaelfranca in https://github.com/Shopify/packwerk/issues/129.

* The old "No offenses detected 🎉 " message has been made sadder and less celebratory in order to conform with the more austere language of the "no stale violations" message. If anyone is matching on that, then it could cause problems. That being said, no one should be matching on that.
* This breaks `packwerk check`. Previously passing checks could easily start failing.

## Type of Change

- [ ] Bugfix
- [x] New feature

### Additional Release Notes

`packwerk check` will now fail if stale violations are present in the codebase.

## Checklist

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -230,7 +230,7 @@ After enforcing the boundary checks for a package, you may execute:
 
     bundle exec packwerk check
 
-Packwerk will check the entire codebase for any violations.
+Packwerk will check the entire codebase for any new or stale violations.
 
 You can also specify folders or packages for a shorter run time:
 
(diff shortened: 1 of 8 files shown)


mclark:
@rafaelfranca performance of this roughly matches the performance of the existing `check` command, around 2 minutes for the github monolith on my laptop. If it is similar for you on Shopify's monolith we should be all good with this hopefully.

rafaelfranca:
Same for us. I'll merge then.

PR #134, opened 2021-07-29
Add option to ignore nested packages

shageman:
This PR adds an option to ignore nested packages when running packwerk. This is helpful in cases where you want to parallelize checks (on CI) for packwerk packages when, for example, the root package is included. Without this option checking the root package always means checking everything (i.e., CI runs can't really parallelize packwerk check).

If unused, this doesn't change performance or behavior of what is currently there.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -231,9 +231,13 @@ After enforcing the boundary checks for a package, you may execute:
 
 Packwerk will check the entire codebase for any new or stale violations.
 
-You can also specify folders for a shorter run time:
+You can also specify folders for a shorter run time. When checking against folders all subfolders will be analyzed, irrespective of nested package boundaries.
 
     packwerk check components/your_package
+
+You can also specify packages for a shorter run time. When checking against packages any packages nested underneath the specified packages will not be checked. This can be helpful to test packages like the root package, which can have many nested packages.
+
+    packwerk check --packages=components/your_package,components/your_other_package
 
 ![](static/packwerk_check.gif)
 
--- a/lib/packwerk/cli.rb
+++ b/lib/packwerk/cli.rb
@@ -1,6 +1,8 @@
 # typed: true
 # frozen_string_literal: true
 
+require "optparse"
+
 module Packwerk
   # A command-line interface to Packwerk.
   class Cli
@@ -124,8 +126,12 @@ module Packwerk
       result.status
     end
 
-    def fetch_files_to_process(paths)
-      files = FilesForProcessing.fetch(paths: paths, configuration: @configuration)
+    def fetch_files_to_process(paths, ignore_nested_packages)
+      files = FilesForProcessing.fetch(
+        paths: paths,
+        ignore_nested_packages: ignore_nested_packages,
+        configuration: @configuration
+      )
       abort("No files found or given. "\
         "Specify files or check the include and exclude glob in the config file.") if files.empty?
       files
@@ -156,9 +162,24 @@ module Packwerk
       end
     end
 
-    def parse_run(paths)
+    def parse_run(params)
+      paths = T.let([], T::Array[String])
+      ignore_nested_packages = nil
+
+      if params.any? { |p| p.include?("--packages") }
+        OptionParser.new do |parser|
+          parser.on("--packages=PACKAGESLIST", Array, "package names, comma separated") do |p|
+            paths = p
+          end
+        end.parse!(params)
+        ignore_nested_packages = true
+      else
+        paths = params
+        ignore_nested_packages = false
+      end
+
       ParseRun.new(
-        files: fetch_files_to_process(paths),
+        files: fetch_files_to_process(paths, ignore_nested_packages),
         configuration: @configuration,
         progress_formatter: @progress_formatter,
         offenses_formatter: @offenses_formatter
--- a/test/unit/files_for_processing_test.rb
+++ b/test/unit/files_for_processing_test.rb
@@ -43,6 +43,39 @@ module Packwerk
       assert_equal files, files.uniq
     end
 
+    test "fetch with custom paths without ignoring nested packages includes only include glob in custom paths including nested package files" do
+      files = ::Packwerk::FilesForProcessing.fetch(
+        paths: ["."],
+        configuration: @configuration,
+        ignore_nested_packages: false
+      )
+      included_file_patterns = @configuration.include.map { |pattern| File.join(@configuration.root_path, pattern) }
+
+      assert_all_match(files, included_file_patterns)
+    end
+
+    test "fetch with no custom paths ignoring nested packages includes only include glob across codebase" do
+      files = ::Packwerk::FilesForProcessing.fetch(
+        paths: [],
+        configuration: @configuration,
+        ignore_nested_packages: true
+      )
+      included_file_patterns = @configuration.include.map { |pattern| File.join(@configuration.root_path, pattern) }
+
+      assert_all_match(files, included_file_patterns)
+    end
+
+    test "fetch with custom paths ignoring nested packages includes only include glob in custom paths without nested package files" do
+      files = ::Packwerk::FilesForProcessing.fetch(
+        paths: ["."],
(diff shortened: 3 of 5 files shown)


exterm:
Do you still need this option for parallelization on a process level now that packwerk has builtin multithread parallelization?

shageman:
> Do you still need this option for parallelization on a process level now that packwerk has builtin multithread parallelization?

The parallelization helped a ton. But even with this, packwerk is the slowest step in one of our pipelines and is thus the top priority to reduce time for if we want to reduce overall build times (we have an active project for this).

> This would be nice to have, but in general I'd like to keep the overall number of supported features small, and therefore thoroughly interrogate the reasons for adding anything. Packwerk now has builtin parallelization, which may or may not already solve your problem (I'm curious to hear more about your use case).

I totally get it. If there is another way in which I could split up work without this change, I would love to test that out!

> On a semantic level, it seems that with this change we're still passing paths to the command, but then asking it to honor package boundaries. 

Great point! I admit that I was initially surprised when I learned that `packwerk check` and `packwerk check .` behave the same ... which makes perfect sense when you know the param is paths and not packages! For context, we have used that param exclusively with package names because it feels like the natural "level of analysis" (why would I define packages one way and then test in some way _not_ along package lines - we don't have a use-case for that).

All that said, I left the API completely backwards compatible because in a chat with some folks on your team I heard that compatibility as a primary concern. 

> Would it make more sense to specify the packages we want to check and specify whether or not to include nested ones?

Maybe? From [USAGE](https://github.com/Shopify/packwerk/blob/main/USAGE.md#checking-for-violations) "You can also specify folders or packages for a shorter run time". From that point of view we're half covered here. I am unsure that restricting the parameter to only allow for packages is very important.

exterm:
> why would I define packages one way and then test in some way not along package lines - we don't have a use-case for that

Checking a single file comes to mind.

> You can also specify folders or packages for a shorter run time

I think that's a bug in the readme :D because as you said, specifying a package would imply "honoring package boundaries".

🤔 

What if we specify a path that contains multiple top level packages? like `packwerk check components`, and the packages in there are `components/merchandising`, `components/sales`, `components/timeline` (completely made up examples 😉 ) and maybe some of them have nested packages.

`honor_package_boundaries` doesn't make sense to me in that case.

From a product perspective (I actually haven't looked at your implementation in detail yet), more semantically consistent alternatives that I can think of are

- `packwerk check components --ignore-nested-packages` (only scans toplevel packages in the path provided)
- `packwerk check --packages="components/merchandising,components/sales"` (providing an alternative option where instead of paths we're specifying packages). Maybe we'd even support package name globbing like `--packages=components/*`.

Both of these would be backwards compatible.

Ignore the concrete syntax, the examples are just to demonstrate the approaches.

shageman:
I can see both of those syntaxes working. Interpreting the first one naively, I am not sure how that one could be done as a non-breaking change (i.e., what if there are non-packaged files underneath components - are they included?). The second one would work wonderfully for the use case I am working on.

Your second proposal without globbing is a straightforward change from what I have so far. Let me know if that seems acceptable and I will send a new version.

shageman:
Thanks, @rafaelfranca, I will see how I might change our setup to work similarly.

I realize I have only been describing the CI use-case, there is a development one too. We're in the process of moving all code out of the root package, but we are not done yet. It falls largely on one team to analyze and improve (read: remove stuff from) the root package. This team would like to analyze the root package only. The difference between interpreting the parameter to check as a package (and honor boundaries) vs treating it as a path is 1:4 in terms of number of files and 1:5 in clock time to perform the task. These numbers will only get starker as we move more code out of root.

More generally, the use-case is for teams owning a package to be able to package-interact with just that package (even if it is oddly shaped).

exterm:
It seems this PR is kind of stuck right now - 

@rafaelfranca after hearing about the local development use case, what do you think about implementing this with the `packwerk check --packages="components/merchandising,components/sales"` command syntax?

(I am in favour)

shageman:
@rafaelfranca any more thoughts on this?

exterm:
Just talked to Rafael, he is on board.

shageman:
@rafaelfranca @exterm - finally got around to switching the API to `--packages="..."` instead of `--honor-package-boundaries`

USAGE.md
```diff
@@ -231,10 +231,14 @@ After enforcing the boundary checks for a package, you may execute:
 
 Packwerk will check the entire codebase for any new or stale violations.
 
-You can also specify folders for a shorter run time:
+You can also specify folders for a shorter run time. When checking against folders all subfolders will be analyzed, irrespective of nested package boundaries.
 
     packwerk check components/your_package
 
+You can also specify packages for a shorter run time. When checking against packages all subfolders will be analyzed and nested packaged boundaries will be honored.
```
exterm, line 238: I would be more explicit and say that if you specify a _package_ to check, packages nested inside of it will not be checked.

exterm: I don't think it's immediately apparent what "honoring nested package boundaries" means in this context.

lib/packwerk/cli.rb
```diff
@@ -156,12 +157,24 @@ def list_validation_errors(result)
     end
 
     def parse_run(paths)
+      packages_modifier = paths.find { |el| el.start_with?("--packages") }
+      paths.reject! { |el| el.start_with?("--packages") }
+      abort("Provide either paths or packages but not both.") if packages_modifier && !paths.empty?
```
exterm, line 162: 👍🏼

lib/packwerk/cli.rb
```diff
@@ -156,12 +157,24 @@ def list_validation_errors(result)
     end
 
     def parse_run(paths)
```
exterm, line 159: I wonder if we should rename "paths" to "params" or something like that?

shageman: Yep!

USAGE.md
shageman: Makes sense - I will push a fix

lib/packwerk/cli.rb
```diff
@@ -123,8 +123,9 @@ def output_result(result)
       result.status
     end
 
-    def fetch_files_to_process(paths)
-      files = FilesForProcessing.fetch(paths: paths, configuration: @configuration)
+    def fetch_files_to_process(paths, honor_package_boundaries)
+      files = FilesForProcessing.fetch(paths: paths, honor_package_boundaries: honor_package_boundaries,
+configuration: @configuration)
```
rafaelfranca, line 128:
```suggestion
        configuration: @configuration)
```

lib/packwerk/cli.rb
```diff
@@ -155,13 +156,25 @@ def list_validation_errors(result)
       end
     end
 
-    def parse_run(paths)
+    def parse_run(params)
+      packages_modifier = params.find { |el| el.start_with?("--packages") }
```
rafaelfranca, line 160: Should we be using an option parser here?

test/unit/files_for_processing_test.rb
```diff
@@ -43,6 +43,30 @@ def setup
       assert_equal files, files.uniq
     end
 
+    test "fetch with custom paths without honoring package boundaries includes only include glob in custom paths including nested package files" do
+      files = ::Packwerk::FilesForProcessing.fetch(paths: ["."], configuration: @configuration,
+honor_package_boundaries: false)
```
rafaelfranca, line 48:
```suggestion
        honor_package_boundaries: false)
```

exterm: Mhh, that's strange, don't we have a rubocop that should catch these things?

shageman:
I found myself wanting to use this again today. @rafaelfranca @exterm - any interest in merging?

shageman:
@exterm: Done! Consolidated commits and use "ignore nested packages" language everywhere

exterm:
Just ran your branch on our main monolith to make sure nothing broke - also tried the `--packages` option; looking good. Thank you for your contribution (and patience) Stefan.

Issue #26, opened 2020-10-01
[Bug Report] Better handling of duplicate entries under load_paths


gy:
**Description**
I this is a usability issue. I accidentally had a duplicate entry under the packwerk load_path configuration which resulted in a cryptic error.

**To Reproduce**
```yml
load_paths:
- file_a
- file_b
...
- file_a
```

```
📦 Packwerk is running validation...

Validation failed ❗
Load path cache in [Project]/packwerk.yml incorrect!
Paths missing from file:

Extraneous load paths in file:
```

**Expected Behaviour**

The two options I see are:

- Let the validation pass by ignoring [duplicates](https://github.com/Shopify/packwerk/blob/main/lib/packwerk/application_validator.rb#L53) and display a warning.
- Improve the error message.

**Version Information**
 - Packwerk 1.0.0
 - Ruby 2.6.5

wildmaples:
Ah yes, thank you for reporting this! The error message does seem to be cryptic for this case. 

I think it's best if we improved the error message so users know there is a duplicated load path in the list. This can be done perhaps with a `Set` comparison of the expected and actual load paths.

PR #137, opened 2021-08-27
ignore duplicated load paths

exterm:
Fix https://github.com/Shopify/packwerk/issues/26

Just ignore dulicated load paths. They don't break anything.

I did however add a `uniq.` when initializing the config, since we don't want to slow down analysis with an unnecessarily large set of load paths.

## Type of Change

- [X] Bugfix
- [ ] New feature

## Checklist

### DIFF
--- a/lib/packwerk/application_validator.rb
+++ b/lib/packwerk/application_validator.rb
@@ -35,9 +35,9 @@ module Packwerk
     end
 
     def check_autoload_path_cache
-      expected = @application_load_paths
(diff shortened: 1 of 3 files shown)


doodzik (approved):

exterm:
I though about a warning but then decided that it's just not worth adding. I guess if you automatically generate the list and accidentally add more entries every time it could lead to parsing the file becoming really slow, but... we're deep in edge case territory here.

PR #138, opened 2021-09-01
Specify `bundle exec` when running `packwerk` commands

andyw8:
Without `bundle exec`, I am getting this failure:

```
/Users/andyw8/.gem/ruby/3.0.2/gems/packwerk-1.3.1/lib/packwerk/package_set.rb:27:in `package_paths':
uninitialized constant #<Class:Packwerk::PackageSet>::Bundler (NameError)
```

### DIFF
--- a/TROUBLESHOOT.md
+++ b/TROUBLESHOOT.md
@@ -14,9 +14,9 @@ Packwerk can give feedback via continuous integration (CI) if you have it set up
 
 You can specify folders or packages in Packwerk commands for a shorter run time:
 
-     packwerk check components/your_package
+     bundle exec packwerk check components/your_package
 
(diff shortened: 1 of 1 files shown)


rafaelfranca:
The documentation leaves out `bundle exec` by design. I consider `bundle exec` a smell and should be avoided. Applications should be using a binstub instead, but it is outside packwerk documentation teaching people how to use bundler, so I left both `bundle exec` and `bin/packwerk` out of the documentation. Thank you for the pull request.

rafaelfranca:
Good point, that issue is actually a different one. Bundler is in reality a runtime dependency but is being defined as development dependency only. We need to change that

PR #145, opened 2021-10-13
[Discussion] Put files into buckets of responsibilities/seams

kenzan100:

Trying to find a seam within the Packwerk module.

We have some rough exploration of how it works (see below diagram). This PR will acts as a self-annotated PR to invite more discussions on what module structure Packwerk should enjoy.

### Responsibilities and its dependencies
Screen Shot 2021-10-13 at 12 55 52 PM

### Data/object transformation journey
Screen Shot 2021-10-13 at 12 55 58 PM

A couple of motivations to do this;

- We're interested in what kind of contracts (interface) Packwerk can and should expose, so that 3rd party integration becomes possible. (`checkers`?)
- We're also interested in documenting the high-level overview of "how it works", much like explained in https://matklad.github.io/2021/02/06/ARCHITECTURE.md.html

Currently, the files completely ignore the proposed namespace. No need to commit editing each file until we're comfortable finalizing the proposed change.

- Does a file belong to an appropriate directory?
- Naming of a directory (namespace) sounds right to you?
- What should be the pattern to bridge these seams? (how to interact/orchestrate between them)

## Type of Change

Discussion only. The merge-able artifact should be created as a separate PR.

### DIFF
--- a/lib/packwerk/run_context.rb
+++ b/lib/packwerk/run_context.rb
@@ -60,13 +60,29 @@ module Packwerk
 
       # Inside NodeProcessor - @reference_extractor.reference_from_node(node, ancestors: ancestors, file_path: @filename)
       # 2. node to constant
-      # 3. constant to reference
       @constant_name_inspectors.each do |inspector|
         constant_name = inspector.constant_name_from_node(node, ancestors: ancestors)
         break if constant_name
       end
 
-      reference_from_constant(constant_name, node: node, ancestors: ancestors, file_path: file_path) if constant_name
+      # Inside ReferenceExtractor - reference_from_constant(constant_name, node: node, ancestors: ancestors, file_path: file_path) if constant_name
+      # 3. constant to reference
+      namespace_path = Node.enclosing_namespace_path(node, ancestors: ancestors)
+      return if local_reference?(constant_name, Node.name_location(node), namespace_path)
+
+      constant =
+        @context_provider.context_for(
+          constant_name,
+          current_namespace_path: namespace_path
+        )
+      return if constant&.package.nil?
+
+      relative_path = Pathname.new(file_path).relative_path_from(@root_path).to_s
+
+      source_package = @context_provider.package_from_path(relative_path)
+      return if source_package == constant.package
+
+      Reference.new(source_package, relative_path, constant)
 
       # Inside NodeProcessor
       # 4. reference to an offence


lib/packwerk/checking_references/checkers/README.md
```diff
@@ -0,0 +1,19 @@
+# Packwerk::Checkers
+
+- Checker implements `invalid_reference?(reference)`, which takes a `Packwerk::Reference` and returns Boolean.
```
exterm, line 3: Nitpick: This is accurate, but it duplicates the interface definition in `Packwerk::Checker`, so I think in the final docs we should just point out that the interface exists instead of putting the details here

lib/packwerk/configurations/README.md
```diff
@@ -0,0 +1,3 @@
+### Configurations
+
+Includes validators which are outside of primary `check` runs.
```
exterm, line 3: I would probably go with the comment in https://github.com/Shopify/packwerk/pull/144/files#diff-5ca9dcd54fd0b9eeef94e00ba6adc7822dc83e5dd2bfb9e8d96fd3c356de8b8aR10-R11

lib/packwerk/extracting_references/README.md
```diff
@@ -0,0 +1,4 @@
+# Extracting references
+
+- Given a `AST::Node`, this module extracts `Packwerk::Reference`.
```
exterm, line 3: If you put it like this, it sounds like everything here works in the context of a single node. However, `ParsedConstantDefinitions` [works on a whole syntax tree](https://github.com/Shopify/packwerk/pull/144/files#diff-8538012fa017b7f4e422be58e6e8adbda8cf1fdba6bbdd45df7b41d8ad5afa64R7).

lib/packwerk/run_context.rb
```diff
@@ -52,7 +54,32 @@ def initialize(
 
     sig { params(file: String).returns(T::Array[T.nilable(::Packwerk::Offense)]) }
     def process_file(file:)
-      file_processor.call(file)
+      # 1. file path to node
```
exterm, line 57: it's more like file path to AST, but that contains nodes, so... you're not wrong

lib/packwerk/run_context.rb
```diff
@@ -52,7 +54,32 @@ def initialize(
...
+      # 4. reference to an offence
+      @checkers.each_with_object([]) do |checker, violations|
+        next unless checker.invalid_reference?(reference)
+        offense = Packwerk::ReferenceOffense.new(
+          location: Node.location(node),
+          reference: reference,
+          violation_type: checker.violation_type
+        )
+        violations << offense
+      end
```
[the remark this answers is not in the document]
exterm: Can we eliminate `RunContext` and let `ParseRun` orchestrate?

exterm:
I think in general this makes sense but we can talk more tomorrow. I think there'll be more interesting questions coming up when you start putting it together.

lib/packwerk/run_context.rb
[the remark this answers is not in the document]
kenzan100: yup, certainly! I'll first focus on isolating the responsibilities of data transformation within the `RunContext`, then will follow up with the PR to eliminate the redundancy between `RunContext` and `ParseRun`.

PR #147, opened 2021-10-14
Refactor - simplify orchestration layer

kenzan100:

First review-ready PR as an artefact of https://github.com/Shopify/packwerk/pull/145.
It's a part of [Packwerk upkeep attempts](https://github.com/Shopify/packwerk/pull/144).

### 1. Extract checkers into its own module.

Biggest extractable seam/interface seems to be the implementation of `Checkers`: We have two default checkers: `DependencyChecker` and `PrivacyChecker`. They're extracted into its own namespace.

#### 2. Decouple reference extraction from offence checks.

`FileProcessor` was having too much responsibilities, of converting file_path, into AST nodes, into a list of offences.
Now seams are clearer; **`file_path` -(`AST::Node`)-> `References` -(`ReferenceChecker`)-> `Offences`.**

1. Overall quality after the separation of concerns.
2. Naming quality modules/methods.

There're self-annotated comments in the diff for one-off topics.

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist


lib/packwerk/file_processor.rb
```diff
@@ -16,24 +18,43 @@ def initialize(node_processor_factory:, parser_factory: nil)
       @parser_factory = parser_factory || Packwerk::Parsers::Factory.instance
     end
 
+    sig do
+      params(file_path: String).returns(
+        T::Array[
+          T.any(
+            Packwerk::Reference,
+            Packwerk::Offense,
```
kenzan100, line 26: `FileProcessor` still have chance to early-return Offence, in cases of `UnknownFileTypeResult` or `ParseError`.

I don't particular like this shared responsibility of the class, and more than happy to revisit the further separation.

lib/packwerk/file_processor.rb
```diff
@@ -16,24 +18,43 @@ def initialize(node_processor_factory:, parser_factory: nil)
...
-      parser = @parser_factory.for_path(file_path)
-      return [UnknownFileTypeResult.new(file: file_path)] if parser.nil?
+      return [UnknownFileTypeResult.new(file: file_path)] if parser(file_path).nil?
 
-      node = File.open(file_path, "r", external_encoding: Encoding::UTF_8) do |file|
-        parser.call(io: file, file_path: file_path)
-      rescue Parsers::ParseError => e
-        return [e.result]
-      end
+      node = convert_to_ast_node(file_path)
```
dougedey-shopify, line 34:
```suggestion
      return [] unless (node = convert_to_ast_node(file_path))
```

That way you can do an early return and remove the if block below (personal choice entirely)

lib/packwerk/file_processor.rb
```diff
@@ -16,24 +18,42 @@ def initialize(node_processor_factory:, parser_factory: nil)
...
+      return [] unless node
+
+      references = []
+      node_processor = @node_processor_factory.for(filename: file_path, node: node)
+      node_visitor = Packwerk::NodeVisitor.new(node_processor: node_processor)
+      node_visitor.visit(node, ancestors: [], result: references)
```
exterm, line 40: Nitpick: I'd probably put these three lines into their own method too while we're at it

exterm: `references_from_ast` or something like that

lib/packwerk/file_processor.rb
```diff
@@ -16,24 +18,42 @@ def initialize(node_processor_factory:, parser_factory: nil)
...
-        node_visitor = Packwerk::NodeVisitor.new(node_processor: node_processor)
+      references
+    rescue Parsers::ParseError => e
+      [e.result]
+    end
+
+    private
 
-        node_visitor.visit(node, ancestors: [], result: result)
+    def convert_to_ast_node(file_path)
```
exterm, line 49: The name is not wrong, but as the node also contains children, I think it makes more sense to talk about a tree than a single node.

So, just `convert_to_ast` would be a better fit IMO.

Or even `parse_into_ast` to be more specific.

lib/packwerk/reference_to_offense/reference_checker.rb
```diff
@@ -0,0 +1,33 @@
...
+        return [reference] unless reference.is_a?(Packwerk::Reference)
+
+        @checkers.each_with_object([]) do |checker, violations|
+          next unless checker.invalid_reference?(reference)
+          offense = Packwerk::ReferenceOffense.new(
+            location: reference.node_location,
+            reference: reference,
+            violation_type: checker.violation_type
+          )
+          violations << offense
```
exterm, line 28: this one is new?

lib/packwerk/reference_to_offense/checkers/privacy_checker.rb
```diff
@@ -0,0 +1,57 @@
...
+        sig do
+          params(privacy_option: T.nilable(T.any(T::Boolean, T::Array[String])))
+            .returns(T::Boolean)
+        end
+        def enforcement_disabled?(privacy_option)
+          [false, nil].include?(privacy_option)
+        end
+      end
+    end
+  end
```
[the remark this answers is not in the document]
kenzan100: yup, they're identical line-by-line, except for the module nesting.

lib/packwerk/file_processor.rb
kenzan100: references array declaration can also be internal to that private method.

lib/packwerk.rb
```diff
@@ -77,4 +74,18 @@ module Generators
...
+    autoload :ReferenceChecker
+
+    module Checkers
+      extend ActiveSupport::Autoload
+
+      autoload :Checker
+      autoload :DependencyChecker
+      autoload :PrivacyChecker
+    end
+  end
```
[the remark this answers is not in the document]
kenzan100: I followed the pattern introduced at https://github.com/Shopify/packwerk/pull/122.

> ruby-native autoloading

It's super-charged by `ActiveSupport::Autoload`, so module nesting is necessary to infer the correct path: `reference_to_offense/checkers/...`.

> All statements in this file wil lbe registered at load time anyway, right?

I may not be understanding the intent of the comment. What alternative would you suggest here?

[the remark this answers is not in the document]
exterm: Ah, turns out I didn't know how `ActiveSupport::Autoload` works. Makes perfect sense then.

exterm:
I don't think 1-2% are statistically significant, so all good

lib/packwerk/reference_to_offense/reference_checker.rb
```diff
@@ -0,0 +1,33 @@
+# typed: true
+# frozen_string_literal: true
+
+module Packwerk
+  module ReferenceToOffense
+    class ReferenceChecker
```
exterm, line 6: It's a bit confusing that `ReferenceChecker` has a very similar name to `PrivacyChecker` and `DependencyChecker` yet is not a `Checker`.

I can't think of a good alternative name right now though...

dougedey-shopify (approved):
looks good, can you add a Changelog entry?

rafaelfranca:
> looks good, can you add a Changelog entry?

Changelog entry for a refactor? If it needs a changelog entry it is not a refactor.

kenzan100:
> > looks good, can you add a Changelog entry?
> 
> Changelog entry for a refactor? If it needs a changelog entry it is not a refactor.

I've looked into https://github.com/Shopify/packwerk/blob/main/USAGE.md, but none of the change here, would affect public facing (a bit ambiguous term) functionality.

lib/packwerk.rb
```diff
@@ -77,4 +74,18 @@ module Generators
     autoload :InflectionsFile
     autoload :RootPackage
   end
+
+  module ReferenceToOffense
```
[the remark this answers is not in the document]
kenzan100: 👍 changed it to `ReferenceChecking`. Thx for the detailed thought exchange 🙏

rafaelfranca (approved):

kenzan100:
🎉 thank you so much for the reviewers.

PR #148, opened 2021-10-15
[WIP] experiment with using Ripper as a parser

exterm:
Make packwerk faster.

Use a faster parser. Previous experiments have shown that `packwerk check` spends upwards of 80% of its time parsing ruby code.

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist

### DIFF
--- a/lib/packwerk/parsers/ripper.rb
+++ b/lib/packwerk/parsers/ripper.rb
@@ -1,4 +1,4 @@
-# typed: true
+# typed: false
 # frozen_string_literal: true
 
 require "ripper"


exterm:
Got some really useful context from Kevin Newton; I can actually hook directly into Ripper to build a parser that returns an `::AST`. He's got some docs at https://kddnewton.com/ripper-docs/ and an example parser he wrote at https://github.com/prettier/plugin-ruby/blob/main/src/ruby/parser.rb.

PR #152, opened 2021-10-21
Promote bundler to a runtime dependency

jordanstephens:

I was trying to run packwerk in a CI container without installing our whole gem bundle, but `/path/to/packwerk check` fails with `uninitialized constant #<Class:Packwerk::PackageSet>::Bundler (NameError)` if bundler isn't loaded. This suggested to me that `bundler` should be promoted from a development dependency to a runtime dependency. It looks like @rafaelfranca mentioned the same notion month or so ago here: https://github.com/Shopify/packwerk/pull/138#issuecomment-914585018.

Ideally we could have run it without `bundler` even being installed, but this will be a step forward for now.

```diff
+  spec.add_dependency("bundler")
-  spec.add_development_dependency("bundler")
```

I think this is pretty straightforward, but let me know if I'm overlooking anything.

## Type of Change

- [x] Bugfix
- [ ] New feature

## Checklist

- [x] I have updated the documentation accordingly. - I don't believe a documentation change is necessary.
- [x] I have added tests to cover my changes. - If you have an idea of an effective test that I could add here, let me know.
- [x] It is safe to rollback this change. - Sure, it wouldn't be any worse than it is now. 👍

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -91,6 +91,7 @@ PATH
       activesupport (>= 5.2)
       ast
       better_html
+      bundler
       constant_resolver
       parallel
       parser
@@ -237,7 +238,6 @@ PLATFORMS
   x86_64-darwin-20
 
 DEPENDENCIES
-  bundler
   byebug
   constant_resolver
   m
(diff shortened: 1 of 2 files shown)


doodzik (approved):

doodzik:
> @doodzik I just signed the CLA. Should it have refreshed automatically, or did I do something wrong?

I think you have to push a new commit to the branch to refresh it.

exterm (approved):

exterm:
Thank you for your contribution, Jordan!

PR #153, opened 2021-10-21
Enable spring by default when executing packwerk init

AlexEvanczuk:
# Summary
This is a preamble to https://github.com/Shopify/packwerk/pull/150. As discussed in the thread, we want to make `bin/packwerk init` generate a spring-enabled bin so users can execute commands like `bin/packwerk check` and `bin/packwerk update-deprecations` more quickly.

# Commits
- change references to packwerk CLI command throughout
  - This commit is all about changing the language in the docs to encourage people to use the generated binstub.
- generate binstub along with other commands
  - This commit is about generating the binstub along with the root package, configuration, and inflections

This is a pre-requisite to removing the need to maintain a list of custom inflections and load_paths, since we will soon be asking the rails application itself for these things on the fly. We want to make it easy for people to use `spring` with `packwerk` so this doesn't cause a performance degradation when running `check` or `update-deprecations` on single files or directories.

In the short-term, we simply generate the binstub so `spring` helps make loading the rails application quicker.
In the long-term, if we can remove our dependency on zeitwerk (https://github.com/Shopify/constant_resolver/issues/26), we can remove the need to load rails altogether.

Making sure my understanding of how spring plays into things is correct.

## Type of Change

- [ ] Bugfix
- [x] New feature

### Additional Release Notes

The docs suggest using `bin/packwerk`. Users who do not have a binstub yet can generate one with `bundle exec packwerk init`

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,8 @@ Or install it yourself as:
 
     $ gem install packwerk
 
-3. Run `packwerk init` to generate the configuration files
+2. Run `bundle binstub packwerk` to generate the binstub
+3. Run `bin/packwerk init` to generate the configuration files
 
 ## Usage
 
--- a/USAGE.md
+++ b/USAGE.md
@@ -42,9 +42,12 @@ The [package principles](https://en.wikipedia.org/wiki/Package_principles) page 
 
 ## Getting started
 
-After including Packwerk in the Gemfile, you can generate the necessary files to get Packwerk running by executing:
+After including Packwerk in the Gemfile, you will first want to generate a binstub:
+You can do this by running `bundle binstub packwerk`, which will generate a [binstub](https://bundler.io/man/bundle-binstubs.1.html#DESCRIPTION) at `bin/packwerk`.
 
-    packwerk init
+Then, you can generate the necessary files to get Packwerk running by executing:
+
+    bin/packwerk init
 
 Here is a list of files generated:
 
@@ -68,6 +71,12 @@ Packwerk reads from the `packwerk.yml` configuration file in the root directory.
 | load_paths           | All application autoload paths            | list of load paths |
 | custom_associations  | N/A                                       | list of custom associations, if any |
 | parallel             | true                                      | when true, fork code parsing out to subprocesses |
+
+## Setting up Spring
+
+[Spring](https://github.com/rails/spring) is a preloader for Rails. Because `packwerk` loads `Rails`, it can be sped up dramatically by enabling spring.  Packwerk supports the usage of Spring.
+Firstly, spring needs to know about the packwerk spring command when spring is loading. To do that, add `require 'packwerk/spring_command'` to `config/spring.rb` in your application.
+Secondly, to enable Spring, first run `bin/spring binstub packwerk` which will "springify" the generated binstub. 
 
 ### Using a custom ERB parser
 
@@ -134,7 +143,7 @@ Any new inflectors should be added to `config/inflections.yml`.
 
 There are some criteria that an application must meet in order to have a valid package system. These criteria include having a valid autoload path cache, package definition files, and application folder structure. The dependency graph within the package system also has to be acyclic.
 
-We recommend setting up the package system validation for your Rails application in a CI step (or through a test suite for Ruby projects) separate from `packwerk check`.
+We recommend setting up the package system validation for your Rails application in a CI step (or through a test suite for Ruby projects) separate from `bin/packwerk check`.
 
 Use the following command to validate the application:
 
@@ -237,7 +246,7 @@ You can also specify folders for a shorter run time:
 
 ![](static/packwerk_check.gif)
 
-In order to keep the package system valid at each version of the application, we recommend running `packwerk check` in your CI pipeline.
+In order to keep the package system valid at each version of the application, we recommend running `bin/packwerk check` in your CI pipeline.
 
 See: [TROUBLESHOOT.md - Sample violations](TROUBLESHOOT.md#Sample-violations)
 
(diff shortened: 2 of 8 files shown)


exterm:
To add some context: This is a partial revert of https://github.com/Shopify/packwerk/pull/128

lib/packwerk/generators/templates/packwerk
```diff
@@ -0,0 +1,23 @@
...
+# Command line arguments needs to be duplicated because spring modifies it
+packwerk_argv = ARGV.dup
+
+begin
+  load(File.expand_path("spring", __dir__))
+rescue LoadError => e
+  raise unless e.message.include?("spring")
+end
+
+require File.expand_path("../config/environment", __dir__)
```
exterm, line 18: One improvement that I'd like to keep from Rafael's PR is to start the app only when we need it.

`ApplicationLoadPaths` starts the app, so we don't need to start it here. I think you can just remove this line.

lib/packwerk/generators/binstub.rb
```diff
@@ -0,0 +1,42 @@
...
+      end
+
+      def generate
+        @out.puts("📦 Generating binstub...")
+        generate_packwerk_validate_script
+      end
+
+      private
+
+      def generate_packwerk_validate_script
```
exterm, line 25: this is not part of the revert, you added this, right? I wonder why we didn't need this before.

We also had the `SpringCommand` before, which I'm not quite sure what it does 🤷🏼

AlexEvanczuk: It was part of the revert:
It was called `ApplicationValidation`, which was generated when initializing packwerk here: https://github.com/Shopify/packwerk/pull/128/files#diff-ac5ae86965a90253f8d378798aa32f4041ec009d6b79c4420e88be6e67108e88L80

That application validation generation copied the binstub here: https://github.com/Shopify/packwerk/pull/128/files#diff-5cc7180b28a34feeec40fbab8714e5bb0dd61d6db8f1554084f63f9d2ca1e9b2L39

I just called it a `binstub` because that's really its main point. Before it was called `ApplicationValidation` I think because it was intended to be what you called when validating the application, but now it's what we will expect people to call always regardless of if validating or if running `bin/packwerk update-deprecations`, etc.

README.md
```diff
@@ -47,7 +47,7 @@ Or install it yourself as:
 
     $ gem install packwerk
 
-3. Run `packwerk init` to generate the configuration files
+3. Run `bundle exec packwerk init` to generate the configuration files
```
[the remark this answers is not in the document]
exterm: yeah, `init` definitely doesn't work without `bundle exec` on master.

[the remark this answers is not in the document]
technicalpickles: It works by having `gem install packwerk`. We have it in our `Gemfile`.

exe/packwerk
```diff
@@ -3,4 +3,10 @@
 
 require "packwerk"
 
-Packwerk::Cli.new(style: Packwerk::OutputStyles::Coloured.new).run(ARGV.dup)
+# Needs to be run in test environment in order to have test helper paths available in the autoload paths
```
AlexEvanczuk, line 6: This is what the previous binstub template did

AlexEvanczuk:
After discussing at length with @exterm , we've made changes to the original implementation. I kept the old PR description for posterity.

Namely, we decided:
A) We leverage `bundle binstub packwerk` to generate the binstub. This is generally best practice and allows `packwerk` to support less and avoid pitfalls where other systems like bundler expect a certain type of binstub.
B) We use `bin/spring binstub packwerk` to "springify" the binstub
C) We require the user to `require 'packwerk/spring_command'` in `config/spring.rb` to actually properly register the packwerk spring command.

I tested this locally and it looks good in our application!

README.md
```diff
@@ -47,7 +47,8 @@ Or install it yourself as:
 
     $ gem install packwerk
 
-3. Run `packwerk init` to generate the configuration files
+2. Run `bundle binstub packwerk` to generate the binstub
+3. Run `bin/packwerk init` to generate the configuration files
```
exterm, line 51: should we add instructions on how to set up spring?

exterm: Oh sorry I see those exist further down

USAGE.md
```diff
@@ -42,9 +42,12 @@ The [package principles](https://en.wikipedia.org/wiki/Package_principles) page
 
 ## Getting started
 
-After including Packwerk in the Gemfile, you can generate the necessary files to get Packwerk running by executing:
+After including Packwerk in the Gemfile, you will first want to generate a binstub:
+You can do this by running `bundle binstub packwerk`, which will generate a [binstub](https://bundler.io/man/bundle-binstubs.1.html#DESCRIPTION) for packwerk.
```
exterm, line 46:
```suggestion
You can do this by running `bundle binstub packwerk`, which will generate a [binstub](https://bundler.io/man/bundle-binstubs.1.html#DESCRIPTION) for packwerk at `bin/packwerk`.
```

USAGE.md
```diff
@@ -69,6 +72,10 @@ Packwerk reads from the `packwerk.yml` configuration file in the root directory.
 | custom_associations  | N/A                                       | list of custom associations, if any |
 | parallel             | true                                      | when true, fork code parsing out to subprocesses |
 
+## Setting up Spring
+
+[Spring](https://github.com/rails/spring) is a preloader for Rails. Because `packwerk` loads `Rails`, it can be sped up dramatically by enabling spring.  Packwerk supports the usage of Spring. To enable Spring, first run `bin/spring binstub packwerk` which will "springify" the generated binstub. Secondly, spring needs to know about the packwerk spring command when spring is loading. To do that, add `require 'packwerk/spring_command'` to the bottom of `config/spring.rb` in your application.
```
exterm, line 77: I my testing, it looks like I need to add the require to the spring config first, before I generate the binstub:

```
 $ bin/spring binstub packwerk
The 'packwerk' command is not known to spring.
```

Also, is there a reason we should recommend to add it to the _bottom_ of the file? Why not just say ```to `config/spring.rb` in your application.```?

test/unit/spring_command_test.rb
```diff
@@ -0,0 +1,20 @@
...
+    test "registers command with Spring when loaded" do
+      require "packwerk/spring_command"
+
+      command = Spring.command("packwerk")
+
+      assert_not_nil(command, message: "packwerk command not registered with Spring")
+      assert_equal("packwerk", command.exec_name)
+      assert_equal("packwerk", command.gem_name)
+      assert_equal("test", command.env(nil))
+    end
```
exterm, line 18: I think you copied this test over from the previous PR, right?

Do you think it is useful?

USAGE.md
AlexEvanczuk: Noted -- I think I didn't properly capture this order dependency. Let me clarify to require and THEN run `bin/spring binstub packwerk`.

No -- I don't know of any reason to add it to the bottom. Let me remove that requirement!

USAGE.md
AlexEvanczuk: Sure, changing

AlexEvanczuk:
Modified this PR @exterm !

test/unit/spring_command_test.rb
exterm: I agree. The test is not very useful but as we don't expect the command to change often, it also doesn't incur a huge cost.

exterm:
Just did a final test of `bin/packwerk check` startup times before and after this change by running on a single, very small ruby file within our monolith.

As `packwerk check` doesn't boot the application yet, I expected either no difference between the springified and non-springified binstub, or even the springified binstub to be a little slower.

However, these are the results:

- springified binstub: ~3.5s
- unspringified binstub: ~4.7s

The speedup must be just from spring caching the bundle then. Pretty cool.

exterm:
I have looked over your changes again and they make perfect sense to me. If you squash the commits into some small number that makes sense, I think this is ready to merge! 👏🏼

exterm (approved):

AlexEvanczuk:
Thanks @exterm -- I think a single commit makes sense. Just squashed, let me know if you agree or disagree and then I will land or make changes.

Excited to get this through. Next up is to rebase https://github.com/Shopify/packwerk/pull/150 off of this, then remove the custom inflections change as well, then cut a new release.

exterm:
yep, I was just running a full `check` of our monolith with the springified binstub to make sure this doesn't break any subtle edge cases - and then got pulled into other things.

PR #157, opened 2021-11-22
Remove need for custom inflections YML file

AlexEvanczuk:
# Summary
See https://github.com/Shopify/packwerk/pull/150 for more context

TLDR: Now that rails is loading with spring during every packwerk run, we can remove the complicated setup of putting inflections into a YML file and loading them at runtime from that YML file.

Remove need for custom inflections YML

Load directly from Rails (`ActiveSupport::Inflections`).

Am I missing anything in the interface difference between the old inflector and the rails inflector?

## Type of Change

- [ ] Bugfix
- [x] New feature

Technically this is non-breaking, depending on your interpretation. Because of `bin/packwerk validate`, we verify that the rails inflections and the cached inflections in the YML file are the same, so things out of the gate should continue to work.

### Additional Release Notes

## Checklist


Gemfile.lock
```diff
@@ -87,7 +87,7 @@ GIT
 PATH
   remote: .
   specs:
-    packwerk (1.3.2)
+    packwerk (1.4.0)
```
exterm, line 90: Let's do the version bump separately. I think I'll bump to 2.0, but we can discuss it separately, this PR can be merged without the bump.

USAGE.md
```diff
@@ -102,45 +100,6 @@ end
...
-  - ['analysis', 'analyses']
-  - ['reserve', 'reserves']
-uncountable:
-  - 'payment_details'
-singular:
-  - [!ruby/regexp /status$/, 'status']
-```
-
-Any new inflectors should be added to `config/inflections.yml`.
-
```
exterm, line 144: 🔥 ❤️

lib/packwerk/configuration.rb
```diff
@@ -47,13 +46,33 @@ def initialize(configs = {}, config_path: nil)
 
         warn(warning)
       end
+
+      inflection_file = File.expand_path(configs["inflections_file"] || "config/inflections.yml", @root_path)
+      if configs["inflections_file"]
+        warning = <<~WARNING
+          DEPRECATION WARNING: The 'inflections_file' key in `packwerk.yml` is deprecated.
+          This value is no longer cached, and you can remove the key from `packwerk.yml`.
+          You can also delete #{configs["inflections_file"]}.
```
exterm, line 55: Just deleting it will actually break your app 😱 

I think we want an actual upgrade guide in a `UPGRADING.md` to get from 1.x to 2.0. We can do that in a separate PR, I'm happy to draft something

exterm:
I love how much code you're removing here.

- reducing implementation complexity
- reducing configuration complexity
- actually speeding things up, the way it looks (through the re-introduction of spring)

lib/packwerk/configuration.rb
exterm: I think it's a straightforward migration (as long as we explain it well) so it's fine to just make it a breaking change

AlexEvanczuk: That makes sense -- should we keep one or both warnings actually? I have the warning about the `load_paths` key to push people to have simpler config, but technically it can still exist. The `inflections_file` key warning is similar. Wondering if we should keep that one? The other one is a bit useful to because it reminds the user to remove the file if it still exists, although perhaps as part of 2.0 we can just remove all three warnings and include the removal of those keys and files and migrations to `inflections.rb` as part of the upgrade path. Let me know what makes sense to you.

exterm:
This is due for a rebase now that the branch it's based on has been merged.

AlexEvanczuk:
> This is due for a rebase now that the branch it's based on has been merged.

Rebased!

AlexEvanczuk:
> Haven't had time to look at it in detail, but I'm now getting a `` /tmp/bundle/ruby/3.0.0/bundler/gems/packwerk-9f8631e8afc1/lib/packwerk/formatters/progress_formatter.rb:19:in `started': undefined method `pluralize' for "file":String (NoMethodError) `` in CI when executing `packwerk check` using this branch.
> 
> Will find some time to investigate.

Can you try now @exterm ?

I think removing `require "active_support/inflector/inflections"` caused that method to no longer exist, so I pushed a new commit (squashed into the same one) that adds:
```
# Provides String#pluralize
require "active_support/core_ext/string"
```
in `packwerk.rb`

PR #159, opened 2021-12-01
Bump packwerk to 2.0.0

AlexEvanczuk:
# Commits
- Remove need for custom inflections (this commit will be dropped when https://github.com/Shopify/packwerk/pull/157 lands)
- Add a changelog
- Bump version to 2.0.0

Bump to 2.0.0 and add a changelog to help users migrate.

I made a `CHANGELOG.md` (instead of `UPGRADING.md`) because it's a pretty well understood convention for communicating various types of changes (including upgrades). I could have put the upgrade specific stuff elsewhere, but I figured it's nice to keep it on one place, for now (perhaps future, more involved upgrades can have their own guides). This also allowed me to backfill the info on `1.4.0` which enabled spring.

Let me know if the changelog entries make sense.

## Type of Change

- [ ] Bugfix
- [x] New feature

### Additional Release Notes

## Checklist

- [x] I have added tests to cover my changes. (in a previous PR)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -87,7 +87,7 @@ GIT
 PATH
   remote: .
   specs:
-    packwerk (1.4.0)
+    packwerk (2.0.0)
       activesupport (>= 5.2)
       ast
       better_html
--- a/lib/packwerk/cli.rb
+++ b/lib/packwerk/cli.rb
@@ -53,8 +53,6 @@ module Packwerk
         output_result(parse_run(args).check)
       when "detect-stale-violations"
         output_result(parse_run(args).detect_stale_violations)
-      when "update"
-        update(args)
       when "update-deprecations"
         output_result(parse_run(args).update_deprecations)
       when "validate"
@@ -112,11 +110,6 @@ module Packwerk
 
       @out.puts(result)
       success
-    end
-
-    def update(paths)
-      warn("`packwerk update` is deprecated in favor of `packwerk update-deprecations`.")
-      output_result(parse_run(paths).update_deprecations)
     end
 
     def output_result(result)
--- a/lib/packwerk/version.rb
+++ b/lib/packwerk/version.rb
@@ -2,5 +2,5 @@
 # frozen_string_literal: true
 
 module Packwerk
-  VERSION = "1.4.0"
+  VERSION = "2.0.0"
 end
--- a/test/integration/custom_executable_test.rb
+++ b/test/integration/custom_executable_test.rb
@@ -129,11 +129,6 @@ module Packwerk
         end
       end
 
-      test "'update' gives deprecation warning" do
-        _out, err = capture_io { assert_successful_run("update") }
-        assert_match(/`packwerk update` is deprecated/, err)
-      end
-
       private
 
       def assert_successful_run(command)


AlexEvanczuk:
> Shopify uses github releases instead of a `CHANGELOG`, that's our convention. An [entry for 1.4.0](https://github.com/Shopify/packwerk/releases/tag/v1.4.0) exists.

Ah okay -- let me remove the `CHANGELOG.md`!

AlexEvanczuk:
@exterm 
I dropped the commit with the changelog. Here is the original for reference which we can use in the github releases (or we can change as you'd like):

```markdown

# Changelog
All notable changes to this project will be documented in this file.

The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).

## [Unreleased]
- This section will contain features, bug fixes, deprecations, changes, etc. that are not yet in a released version.

## [2.0.0] - 2021-12-01
### Changed

- `bin/packwerk update`, which was previously deprecated, has now been removed in favor of `bin/packwerk update-deprecations`.
- Packwerk is no longer required in the production gemfile once your inflections have been migrated back to using `inflections.rb` (see below).
- We no longer require `load_paths` in `packwerk.yml`. You can simply delete the `load_paths` key as it will not be read anymore. Instead, Packwerk will ask Rails for load paths. If you're using spring, make sure to properly set up spring (see [1.4.0](#140---2021-11-24)) to keep packwerk fast.
- We no longer require a custom `inflections.yml` file. Instead, you'll want to revert BACK to using `inflections.rb` as you would have done prior to adopting packwerk. To do this, you'll want to convert back to using the plain [ActiveSupport Inflections API](https://api.rubyonrails.org/classes/ActiveSupport/Inflector/Inflections.html)

Given the following example `inflections.yml`, here is an example `inflections.rb` that would follow. Tip: if you're using git, you can run `git log config/inflections.yml`, find the first commit that introduced `inflections.yml`, find the COMMIT_SHA, and then run `git show COMMIT_SHA` to see what your inflections file looked like before (note that you may have changed `inflections.yml` since then, though).

`config/inflections.yml`
```yml
# List your inflections in this file instead of `inflections.rb`
# See steps to set up custom inflections:
# https://github.com/Shopify/packwerk/blob/main/USAGE.md#Inflections

acronym:
 - 'HTML'
 - 'API'

singular:
 - ['oxen', 'oxen']

irregular:
 - ['person', 'people']

uncountable:
 - 'fish'
 - 'sheep'
```

`config/initializers/inflections.rb`
```ruby
ActiveSupport::Inflector.inflections(:en) do |inflect|
  inflect.acronym 'HTML'
  inflect.acronym 'API'

  inflect.singular 'oxen', 'oxen'

  inflect.irregular 'person', 'people'

  inflect.uncountable 'fish'
  inflect.uncountable 'sheep'
 end
```

## [1.4.0] - 2021-11-24
### Fixed

- Packwerk now integrates properly with Spring and `bin/packwerk` will not produce extra output from spring. If you'd like to enable this, see `USAGE.md` "Setting up Spring." To migrate, you may consider regenerating your binstub. First, `rm bin/packwerk`, then run `bundle binstub packwerk` to regenerate the binstub. Make sure to `require 'packwerk/spring_command'` in `spring.rb` as `USAGE.md` suggests. At that point, you can run `bundle exec spring binstub packwerk` to springify your binstub!
```

AlexEvanczuk:
> 👍🏼 I think I'll extract some of this content into an `UPGRADING.md` that we can then link to from the release notes.

Sounds good to me!

exterm (approved):

exterm:
I'll merge and release later today, have to focus on something else for a while

exterm:
OK I was still waiting for the resolution of a problem with an internal app and packwerk 1.4, but as it seems that it's unrelated and we haven't had any other new error reports, I feel confident releasing 2.0.

If it turns out that 1.4 introduced a problem, we can always fix forward.

Issue #158, opened 2021-11-22
[Bug Report] Explicitly private constants without `::` are ignored


exterm:
**Description**

If explicitly private constants are specified without the `::` prefix, they will not be matched when checking for privacy violations.
Specifying them in this way just does not have any effect.

**To Reproduce**

In a package `a` defining `MyConstant` at the top level (equivalent to `::MyConstant`), set `enforce_privacy: ["MyConstant"]`.
From another package `b`, add a reference to `MyConstant` or `::MyConstant`.
Execute `packwerk check --packages=b`.

**Expected Behaviour**
Packwerk reports an error for the reference to the private constant.

**Actual Behavior**
Packwerk reports no error.

**Version Information**
 - Packwerk: 1.3.2
 - Ruby 3.0.2p107

**Additional Context**
It works just fine if `enforce_privacy` is set to `["::MyConstant"]` instead.

I see two possible solutions:

- Enhance the `ApplicationValidator` to reject explicitly private constants that are not explicitly top-level - or in other words, only accept values that start with `::`
- Better: Change the comparison code to that it assumes all explicitly private constants are top level.

exterm:
Actually not that sure anymore the second proposed solution is better. It would look strange to have constants with and without `::` prefix in the same file.

PR #160, opened 2021-12-01
[Bugfix] Allow for explicitly private constants without `::` prefix

timbogit:
Resolve issue #158 where private constants that are not honored unless prefixed with the top-level namespace operator`::`

Enhanced the logic in `ApplicationValidator` to prefix the constant names found in the `package.yml`s `enforce_privacy` config section with the  top-level namespace operator`::`, unless they already were thusly prefixed.

Is this fix correct for all use cases of the `enforce_privacy` config setting? 
Meaning: can one always assume that the constant names listed in this section are presumed to be absolute constant name references (referenced by their fully-qualified name, starting at the top-level namespace)?

## Type of Change

- [x] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist

### DIFF
--- a/lib/packwerk/application_validator.rb
+++ b/lib/packwerk/application_validator.rb
@@ -64,7 +64,7 @@ module Packwerk
     end
 
     def top_level_constant(name)
-      name.start_with?('::') ? name : "::#{name}"
+      name.start_with?("::") ? name : "::#{name}"
     end
 
     def check_package_manifest_syntax


timbogit:
Could I please get someone to help with looking into why the CLA signing has not worked for me?
Any help appreciated! (maybe @exterm ... given you filed the issue this PR is tried to address?)
🙇

exterm:
oops - sorry for not looking into this earlier. Looks like your close / reopen procedure seems to have worked though.

I have not found time to review this yet, it might have to wait until next week.

exterm:
Sorry, I know I said otherwise in the issue, but your test demonstrates well that it looks confusing to have some constants with and some without the prefix. It suggests some of them are relative references, when indeed they _can't_ be.

exterm:
> explicitly fail with a hint that all constants in the enforce_privacy section need to be prefixed with a `::`.

We could do that in the `validate` command.

I don't think we need to go through a deprecation for this as it's an easy fix and we're going to release `2.0.0` with breaking changes soon anyway, so we could add this to that release.

exterm:
I'm running out of time for the 2.0 release before the holidays so I've decided to release today.

I still think we can do this as a breaking change as it's a really easy fix.

Issue #164, opened 2021-12-09
[Bug Report] Max RSS is Significantly Higher in Version 2.0


jordanstephens:
**Description**
Memory usage in packwerk v2.0 is significantly higher than it was in v1.4—so much higher that we will need to increase the resource class of our CI nodes in order for us to run packwerk v2.0. It looks like the max RSS for the whole process tree has increased from ~850 MB to about ~2.5GB for our use-case.

I ran gnu `time -v` on `bin/packwerk check` in our application working directory using both versions of packwerk and saw the following results:

<table>
<tr>
	<td>Version 1.4</td>
	<td>Version 2.0</td>
</tr>
<tr>
    <td>

> Command being timed: "bin/packwerk check"
> User time (seconds): 79.86
> System time (seconds): 12.22
> Percent of CPU this job got: 719%
> Elapsed (wall clock) time (h:mm:ss or m:ss): 0:12.79
> Average shared text size (kbytes): 0
> Average unshared data size (kbytes): 0
> Average stack size (kbytes): 0
> Average total size (kbytes): 0
> **Maximum resident set size (kbytes): 83560**
> **Average resident set size (kbytes): 0**
> Major (requiring I/O) page faults: 0
> Minor (reclaiming a frame) page faults: 399579
> Voluntary context switches: 10258
> Involuntary context switches: 179007
> Swaps: 0
> File system inputs: 0
> File system outputs: 0
> Socket messages sent: 0
> Socket messages received: 0
> Signals delivered: 36
> Page size (bytes): 4096
> Exit status: 0

</td>
<td>

> Command being timed: "bin/packwerk check"
> User time (seconds): 81.06
> System time (seconds): 11.77
> Percent of CPU this job got: 381%
> Elapsed (wall clock) time (h:mm:ss or m:ss): 0:24.34
> Average shared text size (kbytes): 0
> Average unshared data size (kbytes): 0
> Average stack size (kbytes): 0
> Average total size (kbytes): 0
> **Maximum resident set size (kbytes): 334968**
> **Average resident set size (kbytes): 0**
> Major (requiring I/O) page faults: 5862
> Minor (reclaiming a frame) page faults: 1233804
> Voluntary context switches: 35237
> Involuntary context switches: 156020
> Swaps: 0
> File system inputs: 0
> File system outputs: 0
> Socket messages sent: 16
> Socket messages received: 16
> Signals delivered: 40
> Page size (bytes): 4096
> Exit status: 0
</td>
</table>

[...]

PR #165, opened 2021-12-13
Diagnostics for packwerk memory bloat issue

AlexEvanczuk:
# Summary
This PR is in response to https://github.com/Shopify/packwerk/issues/164

The goal is to assess the feasibility of *no longer* requiring Rails in the main packwerk process so that when forking the process in order to parse files over multiple CPUs, we do not create a memory bloat from cloned Rails applications.

# Approach
1) I pulled the `ApplicationLoadPaths` step into a rake task. Spring automatically springifies rake tasks (more in comments within PR) so we still get the speedup of a springified rails boot.
2) I shell out to that rake task, forking the process.
3) I run some diagnostics to confirm that the main thread continues to not have Rails loaded, and check the size of memory space to confirm.

# Results summary
This appears to have been successful. 

In the main process, our PID is 6775, the memory is about 50000KB, and Rails is not defined.
After forking, the PID is 6788 and memory is 320000KB, and Rails is defined.
Right before we dump the load paths, our PID changs again to 6798 (as a result of spring, I confirm by running `DISABLE_SPRING=1 bin/packwerk check` and note that the PID does not change again, also note the spring output confirms the PID)
Load paths and inflections are dumped correctly, and Rails is defined.
After jumping back to the main process, our PID is back to 6775, our memory back to 50000KB, and Rails is no longer defined.

The full results are at the bottom.

# Followup

If we like this approach, I'll make the following changes (sequencing TBD)
1) The dumped load paths and inflections will be read in from STDOUT (as JSON) and stored in the configuration.
2) Load paths can continue to be read with exactly the same interface (an array of strings on `Configuration`).
3) Inflections can be read via the interface that was removed here https://github.com/Shopify/packwerk/pull/157. Note, importantly, that this interface is now simply an implementation detail, and it is not exposed to the client. It's just used as a wrapper around the JSON, similar to how it previously read the loaded `inflections.yml` file previously.
4) Release 2.1.0. It should not technically be a breaking change, because it should *just work* for the client. However, we would recommend the client unspringify their `bin/packwerk` binstub, since they won't get the bug fix unless they do that.
5) Update the docs to remove the recommendation to springify the binstub, and instead encourage the general usage of spring to speed up the forked Rails process.

# Results of diagnostic check
```
DevSpace(ae-use-local-packwerk) ./ $ bin/packwerk check
fatal: not a git repository (or any of the parent directories): .git
Diagnostics: {:msg=>"Requiring railtie if Rails is defined", :file=>"packwerk.rb", :pid=>6775, :rails_is_defined=>false, :memory=>49516}
📦 Packwerk is inspecting 38010 files
Diagnostics: {:msg=>"About to execute \"bin/rake packwerk:dump_rails_dependencies_to_json\"", :file=>"configuration.rb", :pid=>6775, :rails_is_defined=>false, :memory=>54676}
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
Running via Spring preloader in process 6798
Diagnostics: {:msg=>"Requiring railtie if Rails is defined", :file=>"packwerk.rb", :pid=>6788, :rails_is_defined=>true, :memory=>326392}
[TEST PROF INFO] Spring detected
Diagnostics: {:msg=>"Extracting load paths and inflections", :file=>"dump_rails_dependencies_to_json.rake", :pid=>6798, :rails_is_defined=>true, :memory=>522176}
Load path count: 753
Inflections (acronyms for brevity): {"ssnvs"=>"SSNVS", "noc"=>"NOC", "svb"=>"SVB", "pgp"=>"PGP", "pnc"=>"PNC", "bai"=>"BAI", "hi"=>"HI", "csv"=>"CSV", "epd"=>"EPD", "tdi"=>"TDI", "sdi"=>"SDI"}
Diagnostics: {:msg=>"Successfully extracted Rails data", :file=>"dump_rails_dependencies_to_json.rake", :pid=>6798, :rails_is_defined=>true, :memory=>522348}
Diagnostics: {:msg=>"Finished executing command.", :file=>"configuration.rb", :pid=>6775, :rails_is_defined=>false, :memory=>54512}
Traceback (most recent call last):
        14: from bin/packwerk:29:in `<main>'
        13: from bin/packwerk:29:in `load'
        12: from /var/www/components/packwerk/exe/packwerk:12:in `<top (required)>'
        11: from /usr/local/bundle/ruby/2.7.0/gems/sorbet-runtime-0.5.9291/lib/types/private/methods/_methods.rb:269:in `block in _on_method_added'
        10: from /usr/local/bundle/ruby/2.7.0/gems/sorbet-runtime-0.5.9291/lib/types/private/methods/call_validation.rb:161:in `validate_call'
         9: from /usr/local/bundle/ruby/2.7.0/gems/sorbet-runtime-0.5.9291/lib/types/private/methods/call_validation.rb:161:in `bind_call'
         8: from /var/www/components/packwerk/lib/packwerk/cli.rb:40:in `run'
         7: from /usr/local/bundle/ruby/2.7.0/gems/sorbet-runtime-0.5.9291/lib/types/private/methods/_methods.rb:269:in `block in _on_method_added'
         6: from /usr/local/bundle/ruby/2.7.0/gems/sorbet-runtime-0.5.9291/lib/types/private/methods/call_validation.rb:161:in `validate_call'
         5: from /usr/local/bundle/ruby/2.7.0/gems/sorbet-runtime-0.5.9291/lib/types/private/methods/call_validation.rb:161:in `bind_call'
         4: from /var/www/components/packwerk/lib/packwerk/cli.rb:53:in `execute_command'
         3: from /var/www/components/packwerk/lib/packwerk/parse_run.rb:45:in `check'
         2: from /var/www/components/packwerk/lib/packwerk/parse_run.rb:62:in `find_offenses'
         1: from /var/www/components/packwerk/lib/packwerk/run_context.rb:30:in `from_configuration'
/var/www/components/packwerk/lib/packwerk/configuration.rb:85:in `load_paths': Diagnostic tests complete (RuntimeError)
```

### DIFF
--- a/lib/packwerk.rb
+++ b/lib/packwerk.rb
@@ -7,6 +7,9 @@ require "fileutils"
 
 # Provides String#pluralize
 require "active_support/core_ext/string"
+require 'packwerk/diagnostics'
+Packwerk::Diagnostics.log('Requiring railtie if Rails is defined', __FILE__)
+require 'packwerk/railtie' if defined?(Rails)
 
 module Packwerk
   extend ActiveSupport::Autoload
--- a/lib/packwerk/configuration.rb
+++ b/lib/packwerk/configuration.rb
@@ -79,10 +79,10 @@ module Packwerk
     end
 
     def load_paths
-      T.unsafe(Process).fork do
-        `bin/rake packwerk:dump_rails_dependencies_to_json`
-      end
-      raise 'blah'
+      Packwerk::Diagnostics.log('About to execute "bin/rake packwerk:dump_rails_dependencies_to_json"', __FILE__)
+      puts `bin/rake packwerk:dump_rails_dependencies_to_json`
+      Packwerk::Diagnostics.log('Finished executing command.', __FILE__)
+      raise 'Diagnostic tests complete'
     end
 
     def parallel?
--- a/lib/packwerk/configuration.rb
+++ b/lib/packwerk/configuration.rb
@@ -79,7 +79,10 @@ module Packwerk
     end
 
     def load_paths
-      @load_paths ||= ApplicationLoadPaths.extract_relevant_paths(@root_path, "test")
+      T.unsafe(Process).fork do
+        `bin/rake packwerk:dump_rails_dependencies_to_json`
+      end
+      raise 'blah'
     end
 
     def parallel?
--- a/lib/packwerk/railtie.rb
+++ b/lib/packwerk/railtie.rb
@@ -1,0 +1,13 @@
+require 'packwerk'
+require 'rails'
+
+module Packwerk
+  class Railtie < Rails::Railtie
+    railtie_name :packwerk
+
+    rake_tasks do
+      path = File.expand_path(__dir__)
+      Dir.glob("#{path}/tasks/**/*.rake").each { |f| load f }
+    end
+  end
+end
--- a/lib/packwerk/tasks/packwerk/dump_rails_dependencies_to_json.rake
+++ b/lib/packwerk/tasks/packwerk/dump_rails_dependencies_to_json.rake
@@ -1,0 +1,39 @@
+# typed: strict
+
+module Packwerk
+  # This rake task is in a class so that it can be strictly typed.
+  class TaskLoader
+    include Rake::DSL
+    extend T::Sig
+
+    sig { void }
+    def create_tasks!
+      namespace(:packwerk) do
+        #
+        # The purpose of this task is to give a rake command to output our dependencies on Rails in JSON format.
+        # The reason we want this is because we don't want our `bin/packwerk` process to load Rails, since when we run Parallel.flat_map (in `parse_run.rb`),
+        # we fork the process. `parallel` (https://github.com/grosser/parallel#processes) will create a new forked process that duplicates the memory space,
+        # meaning we have X processes each with their own copy of the Rails app. We want to avoid this, so we want to only grab what we need from Rails in a single process,
+        # and have the forked processes remain lightweight.
+        # See https://github.com/Shopify/packwerk/issues/164 for bug report associated with this.
+        #
+        # Note that by default, `spring` will automatically springify all rake tasks run via `bin/rake` (https://github.com/rails/spring#rake).
+        # So we should expect to see the fast execution of this task by the parent process.
+        #
+        # Lastly -- we do not expect clients to be using this task. It is only exposed so that the client-run `bin/packwerk` command can fork and run this command to get the STDOUT.
+        # We reflect this by running this command with `WARNING="This is private API." bin/rake packwerk:dump_rails_dependencies_to_json`. Open to other suggestions
+        # for incorporating this in a more conventionally private way.
+        #
+        desc('This prints out Rails dependencies to STDOUT.')
+        task(dump_rails_dependencies_to_json: :environment) do |_task, _args|
+          # TODO: Pass this in as an argument to the rake task
+          root_path = File.expand_path('.')
+          puts "HERE WE ARE!"
+          puts ApplicationLoadPaths.extract_relevant_paths(root_path, "test").join("\n")
+        end
+      end
+    end
+  end
+end
+
+Packwerk::TaskLoader.new.create_tasks!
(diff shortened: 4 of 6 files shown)


lib/packwerk.rb
```diff
@@ -7,6 +7,9 @@
 
 # Provides String#pluralize
 require "active_support/core_ext/string"
+require 'packwerk/diagnostics'
+Packwerk::Diagnostics.log('Requiring railtie if Rails is defined', __FILE__)
+require 'packwerk/railtie' if defined?(Rails)
```
AlexEvanczuk, line 12: With this pattern, packwerk will add its rake task if run via `bin/rake`, but otherwise won't interact with rails if rails isn't already loaded.

lib/packwerk/configuration.rb
```diff
@@ -79,7 +79,10 @@ def initialize(configs = {}, config_path: nil)
     end
 
     def load_paths
-      @load_paths ||= ApplicationLoadPaths.extract_relevant_paths(@root_path, "test")
+      Packwerk::Diagnostics.log('About to execute "bin/rake packwerk:dump_rails_dependencies_to_json"', __FILE__)
+      puts `bin/rake packwerk:dump_rails_dependencies_to_json`
+      Packwerk::Diagnostics.log('Finished executing command.', __FILE__)
+      raise 'Diagnostic tests complete'
```
AlexEvanczuk, line 85: I raise to finish the test, but the plan here is to capture the results from shelling out, parse them as JSON, and incorporate them into the configuration. From there, we can use the simple Ruby primitives rather than the Rails stuff.

lib/packwerk/railtie.rb
```diff
@@ -0,0 +1,13 @@
+require 'packwerk'
+require 'rails'
+
+module Packwerk
+  class Railtie < Rails::Railtie
+    railtie_name :packwerk
+
+    rake_tasks do
+      path = File.expand_path(__dir__)
+      Dir.glob("#{path}/tasks/**/*.rake").each { |f| load f }
```
AlexEvanczuk, line 10: We could probably be more explicit and just add the specific rake task we want to include.

I used this as a reference: https://gist.github.com/ntamvl/7a6658b4cd82d6fbd15434f0a9953411

lib/packwerk/tasks/packwerk/dump_rails_dependencies_to_json.rake
```diff
@@ -0,0 +1,43 @@
...
+        # we fork the process. `parallel` (https://github.com/grosser/parallel#processes) will create a new forked process that duplicates the memory space,
+        # meaning we have X processes each with their own copy of the Rails app. We want to avoid this, so we want to only grab what we need from Rails in a single process,
+        # and have the forked processes remain lightweight.
+        # See https://github.com/Shopify/packwerk/issues/164 for bug report associated with this.
+        #
+        # Note that by default, `spring` will automatically springify all rake tasks run via `bin/rake` (https://github.com/rails/spring#rake).
+        # So we should expect to see the fast execution of this task by the parent process.
+        #
+        # Lastly -- we do not expect clients to be using this task. It is only exposed so that the client-run `bin/packwerk` command can fork and run this command to get the STDOUT.
+        # We reflect this by running this command with `WARNING="This is private API." bin/rake packwerk:dump_rails_dependencies_to_json`. Open to other suggestions
```
AlexEvanczuk, line 24: I haven't added this yet, but plan to. Open to other ideas.

lib/packwerk/tasks/packwerk/dump_rails_dependencies_to_json.rake
```diff
@@ -0,0 +1,43 @@
...
+        # for incorporating this in a more conventionally private way.
+        #
+        desc('This prints out Rails dependencies to STDOUT.')
+        task(dump_rails_dependencies_to_json: :environment) do |_task, _args|
+          # TODO: Pass this in as an argument to the rake task
+          root_path = File.expand_path('.')
```
AlexEvanczuk, line 30: i.e. this root will come from the configuration.

PR #167, opened 2021-12-18
Add caching to reduce runtime by 85% (all files) and 42% (single file)

AlexEvanczuk:
# Summary
This is a proof of concept only. This is an attempt to incorporate caching into packwerk. In what I believe is the ideal strategy, this looks like, for the ideal case (primed cache), it reduces parsing time by 85% (from 3m30s to 30s in our codebase) on all files and by 42% (2.4s to 1.4s in our codebase) on a single file.

To use:
`EXPERIMENTAL_PACKWERK_CACHE=1 bin/packwerk check`

Note -- I'm totally fine with this *not* landing, and I don't think it makes sense yet! This was mostly just a fun side project and I totally understand if it doesn't make sense for us to add this complexity!

I don't think we can consider landing this yet until I can find a way to make the cache smaller (right now its 293M because it stores so much unnecessary data). I think if we want to go about this, we would want to pursue an alternate strategy where we probably want to cache a simpler, lower-level data structure (the results of the AST parse). See below in the comments for more.

# Further Details
I wanted to investigate what it would be like to add a cache to packwerk.
Right now, `bin/packwerk check` takes 3 and a half minutes on our codebase, which makes it quite a chore to run.

Until we get a faster parser (like Sorbet), speeding this up makes updating deprecations less toilsome for engineers.

I was curious to see if we can get this fast enough to be in a pre-commit hook. So far, I don't think we've reached that bar, but the speedup is substantial.

# Why this is complicated
A cache is only (perfectly) accurate if it is a pure function and the cache can be completely (or partially, in a targeted way) busted when one of its inputs changes.

A very simple cache here might be to cache all of the offenses in a file, and the cache is always hit if we can find offenses that are mapped to by the Digest of the file contents.

There are several other inputs into this. Namely, something is only an offense if the source package protects against privacy or dependency violations, so as an additional input, we would need to keep a hash of each package.yml and its digest, and also store the package.ymls that a file depends on to check for offenses. There's more -- if a file has a reference to SomeModel, that SomeModel can be in any package, in theory. If a constant is moved to another pack (something which we encourage with packwerk because reorganizing files is so easy), then the source package would change. For this, we'd also need to cache, alongside the `package.yml` of the package that defines the content, but the source file itself. As long as each file exists, we can assume *that* is the file that defines the constant, because packwerk does not permit multiple files to ambiguously define the same constant (https://github.com/Shopify/constant_resolver/blob/e78af0c8d5782b06292c068cfe4176e016c51b34/lib/constant_resolver.rb#L122). 

Therefore, the cache is pretty complicated. The cache is hit if the following is true:
- The file contents are the same
- The file contents of all of the source locations AND packages of each referenced constant are the same (and those files still exist)

# Approaches
I considered a couple of approaches:

## Approach 1: Per file cache
Pros to this approach:
- Each cached file is smaller, which may avoid some types of scaling issues with this cache. 
- In theory, the cache could be shared per file.  That is, if a team wanted to, they could check this cache into version control so the cache is pre-warmed for users. In practice, I doubt users would do this, as it would create lots of file diffs with each PR.
- Incrementally adding to the cache is *fast*. Instead of having to dump the whole cache file each time, we can just dump the additional things.

Cons to this approach:
- At least three file IOs for *each* file that is being parsed. One to get the digest, one to read the cache, and one the cache -- seems to be a lot slower. There are also file IOs for each file/package that the file depends on. Note that we only read each file contents once in the cache, since the cache holds the file content digests in memory.

## Approach 2: One cache for all files
Pros:
- One file IO for each file (one to get the digest), then two file IOs for the cache (one to read the whole cache, one to write the whole cache)
- Sharing the cache would only result in changes to one file, which is maybe a bit more well compartmentalized since its in one file.

Cons:
- Possibly could create a large file IO operation with very large codebases.
- We have to write the entire cache each time, even if only one file has been changed.
- Even when we only pass in one file (e.g. `bin/packwerk check path/to/filer.b`, we need to load the entire cache). The speed of the cache does not scale with the number of files being checked.

## Approach 3: Hybrid cache
In this approach, I wondered if we could have a hybrid cache where we have one large cache produced by the first run of `update-deprecations`, and then apply the incremental cache on a per file-basis. In theory, we could even merge the two caches if there were more than X (1000?) per file caches. Practically, the user can just run `rm -rf tmp/cache/packwerk` and rerun update deprecations to effectively merge the caches and get a bit of a speedup perhaps. I did not implement this approach.

Pros:
- Could be the fastest, since it reads the fewest files, and only writes what is incrementally necessary.

Cons:
- Very complicated implementation, prone to highly variable performance

# Performance across approaches for `bin/packwerk check`
|                  	| Cold cache (all cache misses) 	| Warm Cache (one cache miss) 	| Hot cache (zero cache misses) 	|
|------------------	|-------------------------------	|-----------------------------	|-------------------------------	|
| No cache         	| ~3m30s                        	| n/a                         	| n/a                           	|
| Monolithic cache 	| ~4m16s                        	| ~1m30s                      	| ~1m30s                        	|
| Cache-per file   	| ~4m24s                        	| 0m30s-0m45s                    	| 0m30s-0m45s                    	|
| Hybrid cache     	| did not investigate           	| did not investigate         	| did not investigate           	|

# Performance across approaches for `bin/packwerk check path/to/file.rb`
|                  	| Cold Cache          	| Hot cache           	|
|------------------	|---------------------	|---------------------	|
| No cache         	| ~2.4s               	| n/a                 	|
| Monolithic cache 	| did not investigate 	| did not investigate 	|
| Cache-per file   	| ~2.8s               	| ~1.4s               	|
| Hybrid cache     	| did not investigate 	| did not investigate 	|

# Memory Usage
We should probably do a quick memory profile of this! With the cache, we allocate a lot of extra objects, so it's possible that this has a non-neglible effect on memory.

# Decision
I went with the per file cache. The monolithic cache is very poorly optimized for `bin/packwerk check path/to/file.rb` and doesn't allow us to only redump the cache when the cache changes. The implementation is simplest and it offers the fastest speeds.

# Deployment
If we are interested in this, we can of course have this be opt-in via a `cache` flag in `packwerk.yml` (similar to `parallel`). This can also be considered an experimental feature that may be removed, and perhaps we can have it activated via an environment variable like `ENABLE_EXPERIMENTAL_PACKWERK_CACHE`. This is the way I have it implemented today, but happy to enable it via the configuration flag too (perhaps `experimental_cache`).

# Type of Change

- [ ] Bugfix
- [x] New feature

# Checklist

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -100,6 +100,10 @@ end
 Packwerk::Parsers::Factory.instance.erb_parser_class = CustomParser
 ```
 
+## Using the experimental cache
+Packwerk ships with an experimental cache to help speed up file parsing. You can run packwerk with `EXPERIMENTAL_PACKWERK_CACHE=1 bin/packwerk ...` to turn this on.
+This will write to `tmp/cache/packwerk`.
+
 ## Validating the package system
 
 There are some criteria that an application must meet in order to have a valid package system. These criteria include having a valid autoload path cache, package definition files, and application folder structure. The dependency graph within the package system also has to be acyclic.
--- a/USAGE.md
+++ b/USAGE.md
@@ -102,7 +102,7 @@ Packwerk::Parsers::Factory.instance.erb_parser_class = CustomParser
 
 ## Using the experimental cache
 Packwerk ships with an experimental cache to help speed up file parsing. You can run packwerk with `EXPERIMENTAL_PACKWERK_CACHE=1 bin/packwerk ...` to turn this on.
-This will write to `tmp/cache/packwerk`.
+This will write to `tmp/cache/packwerk`. It is not recommended to use the cache in your test suite, as this cache is experimental and its best to ensure CI is running with no cache.
 
 ## Validating the package system
 
--- a/lib/packwerk.rb
+++ b/lib/packwerk.rb
@@ -15,6 +15,7 @@ module Packwerk
   autoload :ApplicationValidator
   autoload :AssociationInspector
   autoload :OffenseCollection
+  autoload :Cache
   autoload :Cli
   autoload :Configuration
   autoload :ConstantDiscovery
--- a/lib/packwerk/cache.rb
+++ b/lib/packwerk/cache.rb
@@ -3,10 +3,12 @@
 
 #
 # There are some known bugs in this cache:
-# 1) The cache should be busted if the contents of `packwerk.yml` change, since custom associations can affect what is considered a violation
+# 1) The cache should be busted if the contents of `packwerk.yml` change, since custom associations
+# can affect what is considered a violation
 # 2) The cache should be busted if inflections change.
 #
-# In practice, we think these things change rarely enough that when they do change, a user can just run `rm -rf tmp/cache/packwerk` to reset the cache
+# In practice, we think these things change rarely enough that when they do change, a user can
+# just run `rm -rf tmp/cache/packwerk` to reset the cache
 #
 module Packwerk
   class Cache
@@ -22,13 +24,13 @@ module Packwerk
       ).returns(T::Array[RunContext::ProcessedFileResult])
     end
     def self.with_cache(files, root_path:, &block)
-      if ENV['EXPERIMENTAL_PACKWERK_CACHE']
+      if ENV["EXPERIMENTAL_PACKWERK_CACHE"]
         cache = Private.new(files: files, root_path: root_path)
         uncached_files = cache.uncached_files
         Debug.out("Using cache - #{cache.cached_file_count} files are cached, #{uncached_files.count} are not")
         Debug.out("First 5 uncached files: #{uncached_files.first(5).inspect}")
         uncached_results = block.call(uncached_files)
-        cache.cache_results(uncached_files, uncached_results)
+        cache.cache_results(uncached_results)
 
         uncached_results + cache.cached_results
       else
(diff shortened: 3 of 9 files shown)


AlexEvanczuk:
I did some more first principles analysis of performance with packwerk and wanted to share some of the results.

I used a flamegraph to record `bin/packwerk check` (in serial so it could more accurately sample the work in the same process).

Here's what I ran to get this flamegraph:
```
# Get rbspy, the tool used to sample a ruby process
wget https://github.com/rbspy/rbspy/releases/download/v0.9.1/rbspy-x86_64-unknown-linux-gnu.tar.gz
# Unpack the file
tar xvf rbspy-x86_64-unknown-linux-gnu.tar.gz
cp rbspy-x86_64-unknown-linux-gnu /usr/local/bin/rbspy
chmod +x /usr/local/bin/rbspy
# Run without spring, as I believe spring pulls some things into a separate process which affects our ability to accurately sample
DISABLE_SPRING=1 /usr/local/bin/rbspy record --format speedscope --file data.json -- bin/packwerk check
# Tool used to view the flamegraph
npm install -g speedscope
speedscope data.json
```

## Results in time order
Screen Shot 2021-12-20 at 12 10 19 PM

## Results in left heavy
Screen Shot 2021-12-20 at 12 12 50 PM

## Analysis
In general, we can see here that the vast majority of the time is spent in two steps:
1) Parsing the ruby file
2) Traversing through the tree and collecting references.

Note that today, collecting references uses the constant resolver, and we create a custom packwerk 'reference' to establish an edge. Note that if we were to ever use a sorbet parser, we would need to take in a parsed AST that doesn't know about references/offenses and turn it into something else. So I think this could be a good refactor that doubles as giving us something that is much simpler to cache.

AlexEvanczuk:
I tried this other implementation here: https://github.com/Shopify/packwerk/pull/168

The result I believe is much simpler.

PR #173, opened 2022-01-22
Improve typing across codebase

AlexEvanczuk:
# Summary
This PR adds lots of types. In preparation for fixing bugs or thinking about new features, this makes it easier to think about and execute on refactoring the system in a safe way.

This PR can be viewed commit by commit. I generally tried to type one or two clear things for each commit.
I'd be happy to break this into smaller PRs if desired.

# Performance Analysis
Best I can tell, this doesn't have a meaningful/statistically significant impact on performance:
Screen Shot 2022-02-17 at 10 30 32 AM

Add types to improve readability and refactorability of codebase.

Add sorbet sigs based on existing system behavior.

The integrity of the types.

## Type of Change

- [ ] Bugfix
- [ ] New feature

## Checklist


lib/packwerk/constant_discovery.rb
```diff
@@ -52,10 +63,10 @@ def context_for(const_name, current_namespace_path: [])
 
       package = @packages.package_from_path(constant.location)
       ConstantContext.new(
-        constant.name,
-        constant.location,
-        package,
-        package&.public_path?(constant.location),
```
AlexEvanczuk, line 58: Note that this save navigation is removed since the package is considered non-nilable.

I could still add this back with `T.unsafe(package)&.public_path`, but in various other parts of the codebase we expect the package to be non-nilable (wherever we call `constant.package`). Those places would error if the package did not exist.

When we get the package using `package_from_path`, we default to `root` if another package cannot exist, so this plus the fact that other parts would break makes me think this is okay.

lib/packwerk/reference_checking/checkers/dependency_checker.rb
```diff
@@ -20,9 +20,9 @@ def violation_type
             .returns(T::Boolean)
         end
         def invalid_reference?(reference)
-          return false unless reference.source_package
```
AlexEvanczuk, line 23: Same comment as https://github.com/Shopify/packwerk/pull/173/files#r790152320

lib/packwerk/reference_offense.rb
```diff
@@ -25,10 +29,11 @@ def initialize(reference:, violation_type:, location: nil)
 
     private
 
+    sig { params(reference: Reference, violation_type: ViolationType).returns(String) }
     def build_message(reference, violation_type)
       violation_message = case violation_type
       when ViolationType::Privacy
-        source_desc = reference.source_package ? "'#{reference.source_package}'" : "here"
```
AlexEvanczuk, line 31: Same comment as https://github.com/Shopify/packwerk/pull/173/files#r790152320

test/support/factory_helper.rb
```diff
@@ -16,6 +16,7 @@ def build_reference(
       destination_package,
       public_constant
     )
-    Packwerk::Reference.new(source_package, path, constant, source_location)
+    Packwerk::Reference.new(source_package: source_package, relative_path: path, constant: constant,
+source_location: source_location)
```
rafaelfranca, line 20: Wrong indentation.

lib/packwerk/application_validator.rb
```diff
@@ -116,26 +141,27 @@ def check_application_structure
...
 
+    sig { returns(Result) }
     def check_acyclic_graph
-      edges = package_set.flat_map do |package|
-        package.dependencies.map { |dependency| [package, package_set.fetch(dependency)] }
+      edges = @package_set.flat_map do |package|
+        package.dependencies.map { |dependency| [package, @package_set.fetch(dependency)] }
       end
-      dependency_graph = Packwerk::Graph.new(*T.unsafe(edges))
+      dependency_graph = Graph.new(*T.unsafe(edges))
```
rafaelfranca, line 155:
```suggestion

      dependency_graph = Graph.new(*T.unsafe(edges))
```

lib/packwerk/configuration.rb
```diff
@@ -37,7 +37,7 @@ def from_packwerk_config(path)
       :include, :exclude, :root_path, :package_paths, :custom_associations, :config_path, :cache_directory
     )
 
-    def initialize(configs = {}, config_path: nil)
+    def initialize(configs = {}, config_path: DEFAULT_CONFIG_PATH)
       @include = configs["include"] || DEFAULT_INCLUDE_GLOBS
       @exclude = configs["exclude"] || DEFAULT_EXCLUDE_GLOBS
       root = config_path ? File.dirname(config_path) : "."
```
rafaelfranca, line 43: If the config path is required now (at least that is what the commit message says) do we need this line?

lib/packwerk/constant_discovery.rb
```diff
@@ -15,32 +15,50 @@ module Packwerk
...
     # Analyze a constant via its name.
     # If the constant is unresolved, we need the current namespace path to correctly infer its full name
-    #
-    # @param const_name [String] The unresolved constant's name.
-    # @param current_namespace_path [Array<String>] (optional) The namespace of the context in which the constant is
-    #   used, e.g. ["Apps", "Models"] for `Apps::Models`. Defaults to [] which means top level.
-    # @return [Packwerk::ConstantDiscovery::ConstantContext]
+    sig do
+      params(
+        # The unresolved constant's name.
```
rafaelfranca, line 55: Will this comment be presented in the documentation? If not we should not remove the documentation in favor of a comment in the type.

lib/packwerk/offense.rb
```diff
@@ -22,9 +29,11 @@ def initialize(file:, message:, location: nil)
 
     sig { params(style: OutputStyle).returns(String) }
     def to_s(style = OutputStyles::Plain.new)
-      if location
+      # So sorbet knows it's non-nil after the if check
```
rafaelfranca, line 32: No need for this comment. If people try to remove they will know why it is there.

lib/packwerk/offense.rb
```diff
@@ -22,9 +29,11 @@ def initialize(file:, message:, location: nil)
...
+      flow_sensitive_location = location
```
rafaelfranca, line 33:
```suggestion
      location = self.location
```

lib/packwerk/parse_run.rb
```diff
@@ -83,18 +99,22 @@ def find_offenses(show_errors: false)
...
 
-    def serial_find_offenses
-      all_offenses = T.let([], T.untyped)
-      @files.each do |path|
-        offenses = yield path
-        all_offenses.concat(offenses)
+    sig { params(block: ProcessFileProc).returns(T::Array[Offense]) }
+    def serial_find_offenses(&block)
+      all_offenses = T.let([], T::Array[Offense])
+      begin
```
rafaelfranca, line 105: I think it is better to remove this `begin/end`. Just let the method rescue the exception in the entire body

lib/packwerk/parsers/parser_interface.rb
```diff
@@ -0,0 +1,20 @@
...
+# frozen_string_literal: true
+
+module Packwerk
+  module Parsers
+    module ParserInterface
+      extend T::Helpers
+      extend T::Sig
+
+      # So Sorbet knows `nil?` exists
+      include Kernel
```
rafaelfranca, line 11: We should not be including Kernel. The right way to fix this is with `requires_ancestor`

lib/packwerk/reference_checking/checkers/privacy_checker.rb
```diff
@@ -23,10 +23,11 @@ def invalid_reference?(reference)
           return false if reference.constant.public?
 
           privacy_option = reference.constant.package.enforce_privacy
-          return false if enforcement_disabled?(privacy_option)
-
+          enforcement_disabled = privacy_option == false || privacy_option.nil?
```
rafaelfranca, line 26: Why is this reimplementing `enforcement_disabled?`?

lib/packwerk/reference_extractor.rb
```diff
@@ -66,17 +66,19 @@ def self.get_fully_qualified_references_and_offenses_from(unresolved_references_
             current_namespace_path: unresolved_reference.namespace_path
           )
 
-        next if constant&.package.nil?
+        next if constant.nil?
```
rafaelfranca, line 69:
```suggestion
        next if constant.nil?

```

lib/packwerk/reference_extractor.rb
```diff
@@ -66,17 +66,19 @@ def self.get_fully_qualified_references_and_offenses_from(unresolved_references_
...
+        package_for_constant = constant.package
+        next if package_for_constant.nil?
```
rafaelfranca, line 71:
```suggestion
        next if package_for_constant.nil?
```
```suggestion

        next if package_for_constant.nil?
```

lib/packwerk/reference_offense.rb
rafaelfranca: Wouldn't this change the logic? `here` will not be returned anymore.

sorbet/rbi/gems/actionmailbox@7.0.0.alpha-d612542336d9a61381311c95a27d801bb4094779.rbi
```diff
@@ -2,7 +2,7 @@
 # This is an autogenerated file for types exported from the `actionmailbox` gem.
 # Please instead update this file by running `bin/tapioca sync`.
 
-# typed: true
+# typed: strict
```
rafaelfranca, line 5: All the files generated by sorbet should not change.

test/support/application_fixture_helper.rb
```diff
@@ -1,7 +1,10 @@
-# typed: false
+# typed: true
 # frozen_string_literal: true
 
 module ApplicationFixtureHelper
+  # So sorbet knows self.raise exists
+  include Kernel
```
rafaelfranca, line 6: `requires_ancestor` is the right way to fix this.

lib/packwerk/reference_offense.rb
AlexEvanczuk: I may be misunderstanding something, but best I can tell `source_package` is never nil -- if it can't find a matching package, it always says the source package is root. So I think `here` is dead code.

lib/packwerk/constant_discovery.rb
AlexEvanczuk: I'm not totally following this one -- are you referring to the end-user facing documentation (README and USAGE)?

When you say a comment in the type -- I have a comment over the type right now, let me know what you mean by that or if what I have now works.

lib/packwerk/offense.rb
AlexEvanczuk: Makes sense

lib/packwerk/parse_run.rb
AlexEvanczuk: This one is needed because Sorbet doesn't know that `all_offenses = ...` won't throw an error. So having the `begin` after that declaration let's Sorbet know that it's an array, not a nilable array:

lib/packwerk/reference_checking/checkers/privacy_checker.rb
AlexEvanczuk: This allowed the type to be refined through flow sensitivity. At any rate I reverted this after no longer declaring `Reference` as a `T::Struct`.

lib/packwerk/reference_extractor.rb
AlexEvanczuk: pushing this as a separate commit

test/support/application_fixture_helper.rb
AlexEvanczuk: Unfortunately it doesn't -- going to use `T.unsafe` for now

rafaelfranca: Ah, it is because `requires_ancestor` is not enable in our sorbet config. Let's add `--enable-experimental-requires-ancestor` to sorbet/config

AlexEvanczuk: Ah perfect, adding!

Issue #154, opened 2021-10-29
[Bug Report] Packwerk suggests to run `update-deprecations` when it doesn't apply


exterm:
**Description**
Packwerk now checks for both new violations and stale deprecations when running `check`. However, deprecation lists are per package, not per file, so we can only determine which deprecations are valid when checking whole packages. When `check`ing single source files, we won't find all the violations that are in the deprecation file for the package, and then packwerk (wrongly) outputs ```There were stale violations found, please run `packwerk update-deprecations` ```

**To Reproduce**
1. Choose a package that has recorded deprecations from multiple files
2. run `packwerk check some_file.rb` on just a single file
3. witness the confusing message

**Expected Behaviour**
Output
```
📦 Packwerk is inspecting 1 file
.
📦 Finished in 0.76 seconds

No offenses detected
```

**Version Information**
 - Packwerk: 1.3.2
 - Ruby 3.0.1p64

**Additional Context**
I can think of three possible solutions here:
- omit stale violations check if we're checking partial packages
- compare stale violations per file instead of per package; when just checking `some_file.rb` we would only compare the found violations to those entries in the deprecation file that are referring to the same source file
- remove stale violation check from `check`, put into a different command or an option `--check-stale`

My favourite is the middle one, make the comparison more intelligent, because then it will always "just work". The same fix could then also be applied to `update-deprecations`, which currently misbehaves in a similar way when run on a single file.

exterm:
❤️

PR #176, opened 2022-02-22
Fix issues where running `check` on a single file mistakenly reports stale violations

AlexEvanczuk:
# Summary

Resolves https://github.com/Shopify/packwerk/issues/154 by only determining if a violation is stale if `bin/packwerk check` is called with input variables including the potentially stale violation.

Resolve this bug to support `bin/packwerk check` on a single file behaving as expected.

Comparing potentially stale violations against input vars.

Is there a simpler way to accomplish this?

## Type of Change

- [x] Bugfix
- [ ] New feature

## Checklist

### DIFF
--- a/lib/packwerk/deprecated_references.rb
+++ b/lib/packwerk/deprecated_references.rb
@@ -36,29 +36,43 @@ module Packwerk
     sig { params(reference: Packwerk::Reference, violation_type: Packwerk::ViolationType).returns(T::Boolean) }
     def add_entries(reference, violation_type)
       package_violations = @new_entries.fetch(reference.constant.package.name, {})
-      entries_for_file = package_violations[reference.constant.name] ||= {}
+      entries_for_constant = package_violations[reference.constant.name] ||= {}
 
-      entries_for_file["violations"] ||= []
-      entries_for_file["violations"] << violation_type.serialize
+      entries_for_constant["violations"] ||= []
+      entries_for_constant["violations"] << violation_type.serialize
 
-      entries_for_file["files"] ||= []
-      entries_for_file["files"] << reference.relative_path.to_s
+      entries_for_constant["files"] ||= []
+      entries_for_constant["files"] << reference.relative_path.to_s
 
       @new_entries[reference.constant.package.name] = package_violations
       listed?(reference, violation_type: violation_type)
     end
 
-    sig { returns(T::Boolean) }
-    def stale_violations?
+    sig { params(for_files: T::Set[String]).returns(T::Boolean) }
+    def stale_violations?(for_files)
       prepare_entries_for_dump
+
       deprecated_references.any? do |package, package_violations|
-        package_violations.any? do |constant_name, entries_for_file|
+        package_violations_for_files = {}
+        package_violations.each do |constant_name, entries_for_constant|
+          entries_for_files = for_files & entries_for_constant["files"]
+          next if entries_for_files.none?
+
+          package_violations_for_files[constant_name] = {
+            "violations" => entries_for_constant["violations"],
+            "files" => entries_for_files.to_a,
+          }
+        end
+
+        return true if package_violations_for_files.empty?
+
+        package_violations_for_files.any? do |constant_name, entries_for_constant|
           new_entries_violation_types = @new_entries.dig(package, constant_name, "violations")
           return true if new_entries_violation_types.nil?
 
-          if entries_for_file["violations"].all? { |type| new_entries_violation_types.include?(type) }
+          if entries_for_constant["violations"].all? { |type| new_entries_violation_types.include?(type) }
             stale_violations =
-              entries_for_file["files"] - Array(@new_entries.dig(package, constant_name, "files"))
+              entries_for_constant["files"] - Array(@new_entries.dig(package, constant_name, "files"))
             stale_violations.any?
           else
             return true
@@ -93,9 +107,9 @@ module Packwerk
     sig { returns(ENTRIES_TYPE) }
     def prepare_entries_for_dump
       @new_entries.each do |package_name, package_violations|
-        package_violations.each do |_, entries_for_file|
-          entries_for_file["violations"].sort!.uniq!
-          entries_for_file["files"].sort!.uniq!
+        package_violations.each do |_, entries_for_constant|
+          entries_for_constant["violations"].sort!.uniq!
+          entries_for_constant["files"].sort!.uniq!
         end
         @new_entries[package_name] = package_violations.sort.to_h
       end
--- a/test/unit/deprecated_references_test.rb
+++ b/test/unit/deprecated_references_test.rb
@@ -40,13 +40,13 @@ module Packwerk
 
     test "#stale_violations? returns true if deprecated references exist but no violations can be found in code" do
       deprecated_references = DeprecatedReferences.new(destination_package, "test/fixtures/deprecated_references.yml")
-      assert deprecated_references.stale_violations?
+      assert deprecated_references.stale_violations?(Set.new)
     end
 
     test "#stale_violations? returns false if deprecated references does not exist but violations are found in code" do
       deprecated_references = DeprecatedReferences.new(destination_package, "nonexistant_file_path")
       deprecated_references.add_entries(build_reference, ViolationType::Dependency)
-      refute deprecated_references.stale_violations?
+      refute deprecated_references.stale_violations?(Set.new)
     end
 
     test "#stale_violations? returns false if deprecated references violation match violations found in code" do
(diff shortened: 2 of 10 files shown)


lib/packwerk/constant_discovery.rb
```diff
@@ -19,6 +19,8 @@ class ConstantDiscovery
 
     ConstantContext = Struct.new(:name, :location, :package, :public?)
 
+    # @param constant_resolver [ConstantResolver]
+    # @param packages [Packwerk::PackageSet]
```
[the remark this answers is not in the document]
AlexEvanczuk: There's this: https://github.com/dduugg/yard-sorbet

May be worth seeing if we could stop maintaining some of these YARD docs by hand as a separate effort 🤔

AlexEvanczuk:
Hey @mclark and @exterm --
I think this resolve the "packwerk check on a single file" bug from #154. If we wanted we could probably use the same approach to delete deprecated references files if 100% of the violations in the file are stale.

Basically -- right now we only create one key value pair in `OffenseCollection#deprecated_references` when calling `add_offense`. That is -- we only create the entry if there is an error for a file in a given package. Another option would be to initialize *all* of the package => deprecated references key value pairs.

Then, when we call `dump_deprecated_references_files`, instead of checking if new entries is empty, we would need to compare it against passed in files. To do this, `OffenseCollection` would need to store the input files as an instance var during initialization. Then we can ask "do the input files represent 100% of what was in `deprecated_references.yml`, and is the file now empty? If so, we think we can delete it."  I think we'd need to do a couple of tidying refactors first, because we need to change the way we think about keeping track of violations. I think we're onto something in the other PR but just wanted to share this other path.

mclark:
Hmm I think I'm leaning the other way on this one, at least assuming I understand the issues properly 😄 

`<package>/deprecated_references.yml` is an uncontextualized artifact representing the known and accepted violations for the package. If I update the deprecations, I expect that file to accurately reflect the deprecations for the _entire_ package. If we allow scoping to a single path then there are a couple consequences we need to accept:

* The file can only be deleted when packwerk is able to determine that the entire scope of the package is being evaluated.
* It makes it easier to have a _partially valid_ file. This is different from `packwerk check` which is an ephemeral result and therefore is always dependent on the exact command that was used to output it.

While it is true that generally organizations will have a CI check set up to ensure invalid `deprecated_references.yml` files don't make it to the main branch, I don't think we should rely on that or ideally even allow the generation of these files when it should be straightforward enough to avoid them by simply parsing the entire package.

Can you elaborate on why you feel the need to have file level granularity for `update-deprecations`?

Here's an idea, would it suffice to determine the minimum encompassing package set for a given set of paths and then run against just those packages? For example, if we were to run `bin/packwerk update-deprecations packs/my-pack/app/some/file.rb` we could determine that `file.rb` is part of the `my-pack` package and then evaluate all files for that package. Would that help? I feel like I don't quite understand the larger vision of what you are aiming for and that's probably the issue here.

mclark:
> The file being outdated in any way would always make it partially valid, IMO. Like if I add a new violation but don't update deprecations, the other violations in the file are still valid, so is it then partially valid?

I think it's more of a distinction at the time of execution. If I run `packwerk update-deprecations` I expect the state of the file to be valid WRT the state of the code at the time the command was run. If I can scope it to a single path, this is no longer true, I'm just updating it to a new state that will _usually_ be accurate.

I realize this may be an overly pedantic argument but until I understand the benefits of making the concession, I don't understand why we should make it.

> there is another reason to consider allowing update-deprecations to be run on single files - consistency

As it is broken now anyway, I'm not really sure the benefit here. If you are referring to consistency with other commands, I believe automatic re-scoping to the minimal enclosing package would yield functional accuracy as described above in addition to consistency across commands.

AlexEvanczuk:
Thanks all for the great engagement here!

Before I forget, let me know if you have any feedback on this particular PR or if we're comfortable merging.

I'm not sure I have a strong opinion yet on the ideas around validity of the `deprecated_references.yml` files. I'm not sure if having these files be accurate with respect to a whole pack at the time of the last run has a benefit or a guarantee that folk would need to rely on.

However – I have a proposal... since today's behavior is outright broken, I propose we make a change to no longer permit `bin/packwerk update-deprecations` with non-pack-name arguments. Note this differs from an earlier proposed solution where we permit file arguments but file the pack the files reside in. The reason for this difference is that by constraining the supported behavior to what actually works today (pack based arguments), it allows us to later come back and decide to support file-based arguments. That is, if we identify the pack the file resides in and use that, then making changes per our discussions would be a breaking change instead of a feature enhancement. I like this approach because it's basically just removing a 'feature' that never worked and giving us time and space to come back to it later.

Why have file level granularity for update-deprecations... good question! I think this may be feature creep, which is why I proposed the above as a stopgap. There are a couple of use cases I'd like to at least have the option to support in the future:
- VSCode Extension "autofix" – a user can click a violation and one of the quick fix options is to run `update-deprecations` on that file
- `danger-packwerk` auto fix – a user can respond with a certain comment to have the danger check update deprecations on that file and push a commit
- Fast user updates – for the same reasons a user wants `bin/packwerk check` on a single file to be fast, they may want it to be fast to run `update-deprecations` on a single file

These aren't rock solid reasons by any means, which is why I think for now restricting `update-deprecations` to be run on *all* packs (no arguments) or specific packs and raising an error otherwise would be a nice stop gap. What do you all think?

mclark:
> I believe we're reaching agreement as, if I understand correctly, that is what I suggested in https://github.com/Shopify/packwerk/pull/199#issuecomment-1111249186, and got a thumb up from Matthew on 😄

Sounds good to me!

> There are a couple of use cases I'd like to at least have the option to support in the future

@AlexEvanczuk thank you for the brain dump of what you were considering! This helps a lot for me to understand where you are coming from ❤️  This all makes sense, and I can understand what you are trying to achieve.

I think the core of our discussion is that both of you are working on codebases with much more mature usage of Packwerk and the list of deprecated references for each of your packages is mutating but rarely (never?) disappearing entirely. Updating it based on a single file at a time makes sense in this world, basically tweaking it as you write code.

As for me, my team is currently ramping up Packwerk usage to much higher levels than before and changing privacy enforcement and dependency enforcement settings, sometimes for individual constants. This results in _all_ offenses being removed fairly regularly! Currently, however, this means the `deprecated_references.yml` file just freezes where it was because it is not being deleted as expected when it runs out of violations. In #199 I'm hoping to fix that so the file is entirely removed when the last violations for a package are resolved.

Assuming #199 is merged, updating deprecations for individual paths means that deleting the file becomes a very inconsistent proposition. If the last violation is removed for a package, `packwerk update-deprecations` deletes `deprecated_references.yml` but `packwerk update-deprecations <path-to-changed-file.rb>` does not.

I wonder if we just make `deprecated_references.yml` optionally empty and _never_ delete it as a result of `packwerk update-deprecations` we might find the consistency we all are looking for here. This is a change that could be made in #199. Would this work for you?

lib/packwerk/deprecated_references.rb
```diff
@@ -36,28 +36,41 @@ def listed?(reference, violation_type:)
...
-        package_violations.any? do |constant_name, entries_for_file|
+        package_violations_for_files = {}
+        package_violations.each do |constant_name, entries_for_constant|
+          entries_for_files = for_files & entries_for_constant["files"]
+          next if entries_for_files.none?
+          package_violations_for_files[constant_name] = {
+            "violations" => entries_for_constant["violations"],
+            "files" => entries_for_files.to_a,
+          }
+        end
```
mclark, line 64: What do you think of extracting this filtering step into its own method?

Is there much performance impact when running stale violations checks on the entire codebase as a result of essentially copying all the violations data?

test/unit/deprecated_references_test.rb
```diff
@@ -86,20 +89,33 @@ class DeprecatedReferencesTest < Minitest::Test
...
+      assert deprecated_references.stale_violations?(Set.new([
+        "orders/app/jobs/orders/sweepers/purge_old_document_rows_task.rb",
+        "orders/app/models/orders/services/adjustment_engine.rb",
+      ]))
+    end
+
+    test "#stale_violations? returns false if violations in deprecated_references.yml exist but are not found when checking for violations" do
+      package = Package.new(name: "buyers", config: { "enforce_dependencies" => true })
+      deprecated_references = DeprecatedReferences.new(package, "test/fixtures/deprecated_references.yml")
+      assert deprecated_references.stale_violations?(
```
mclark, line 101: This assertion doesn't match the description of the test

test/unit/deprecated_references_test.rb
```diff
@@ -86,20 +89,33 @@ class DeprecatedReferencesTest < Minitest::Test
...
+        Set.new(["orders/app/jobs/orders/sweepers/purge_old_document_rows_task.rb"])
+      )
     end
 
     test "#stale_violations? returns true if violations in deprecated_references.yml exist but are not found when checking for violations" do
       package = Package.new(name: "buyers", config: { "enforce_dependencies" => true })
 
       violated_reference = build_reference(
```
mclark, line 109: Does this test pass with this reference removed?

mclark (commented):
While I'm here commenting so much about other stuff I thought I may as well review the PR in question 😆  Your changes look great, just a few minor nit picking kind of suggestions, thanks for the cleanup as always! 🙇

test/unit/deprecated_references_test.rb
AlexEvanczuk: Thanks, fixed!

lib/packwerk/deprecated_references.rb
AlexEvanczuk: I ran this against our codebase and didn't notice any huge changes in performance. I'm pretty sure the variation is just noise.

AlexEvanczuk:
@rafaelfranca Let me know if you're good with this one merging too!

AlexEvanczuk:
@rafaelfranca Addressed your comments and squashed the commits in preparation for merging!

rafaelfranca (approved):

gmcgibbon:
https://rubygems.org/gems/packwerk/versions/2.2.1

PR #178, opened 2022-02-24
Tidy up FilesForProcessing

AlexEvanczuk:
- Add typing to FilesForProcessing
- Make absolute vs relative explicit

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist

### DIFF
--- a/lib/packwerk/cli.rb
+++ b/lib/packwerk/cli.rb
@@ -122,16 +122,16 @@ module Packwerk
       result.status
     end
 
-    sig { params(paths: T::Array[String], ignore_nested_packages: T::Boolean).returns(T::Array[String]) }
-    def fetch_files_to_process(paths, ignore_nested_packages)
-      files = FilesForProcessing.fetch(
-        paths: paths,
+    sig { params(relative_file_paths: T::Array[String], ignore_nested_packages: T::Boolean).returns(T::Array[String]) }
+    def fetch_files_to_process(relative_file_paths, ignore_nested_packages)
+      absolute_file_paths = FilesForProcessing.fetch(
+        relative_file_paths: relative_file_paths,
         ignore_nested_packages: ignore_nested_packages,
         configuration: @configuration
       )
       abort("No files found or given. "\
-        "Specify files or check the include and exclude glob in the config file.") if files.empty?
-      files
+        "Specify files or check the include and exclude glob in the config file.") if absolute_file_paths.empty?
+      absolute_file_paths
     end
 
     sig { params(_paths: T::Array[String]).returns(T::Boolean) }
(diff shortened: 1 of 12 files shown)


lib/packwerk/reference_extractor.rb
```diff
@@ -94,21 +101,21 @@ def self.get_fully_qualified_references_and_offenses_from(unresolved_references_
...
+    def reference_from_constant(constant_name, node:, ancestors:, absolute_file_path:)
       namespace_path = Node.enclosing_namespace_path(node, ancestors: ancestors)
 
       return if local_reference?(constant_name, Node.name_location(node), namespace_path)
 
-      relative_path = Pathname.new(file_path).relative_path_from(@root_path).to_s
+      relative_file_path = Pathname.new(absolute_file_path).relative_path_from(@root_path).to_s
       location = Node.location(node)
 
       UnresolvedReference.new(
```
AlexEvanczuk, line 115: Note that I did not change `UnresolvedReference`'s internal `Struct` implementation to name the struct variable differently. This would not only break the cache for users, but I think it would actually raise a slightly obscure error. It would be easily resolved with `rm -rf path/to/cache`, but we shouldn't break public API here. If we want to change the implementation (not worth it IMO right now), we could bust the cache if we fail to hydrate a `UnresolvedReference`.

lib/packwerk/cli.rb
```diff
@@ -122,16 +122,16 @@ def output_result(result)
       result.status
     end
 
-    sig { params(paths: T::Array[String], ignore_nested_packages: T::Boolean).returns(T::Array[String]) }
-    def fetch_files_to_process(paths, ignore_nested_packages)
-      files = FilesForProcessing.fetch(
-        paths: paths,
+    sig { params(relative_file_paths: T::Array[String], ignore_nested_packages: T::Boolean).returns(T::Array[String]) }
+    def fetch_files_to_process(relative_file_paths, ignore_nested_packages)
+      absolute_file_paths = FilesForProcessing.fetch(
```
rafaelfranca, line 127: Those are not just `absolute_file_paths` they are the actual path for existing files. An absolute file path might exist or not. This code calculate the absolute path and return a file that exists. Maybe `files` or `found_files` is a better name.

AlexEvanczuk: Ah -- so is the distinction here similar to how a `Pathname` can be in memory, but there may not be an underlying file at that path? So what you want is to be more explicit that it's not only a path (which may or may not exist) but a path that does exist?

rafaelfranca: `absolute_files` makes sense to me.

AlexEvanczuk: I pushed up this change

rafaelfranca (approved):

PR #179, opened 2022-02-24
More cleanly separate out references from offenses when processing a file

AlexEvanczuk:
I noticed what felt like a bit of a code smell in the checkers. Namely:
```
return [reference] if reference.is_a?(Packwerk::Offense)
```

It felt like the checkers didn't really want to be receiving `Packwerk::Offense`s at all. I think a design which more clearly represents the intention of the system is that we might have errors related to parsing or otherwise getting references to checkers, and then separately we have `Packwerk::ReferenceOffense` generated by the checkers. I want to represent this by returning a `ProcessedFile` from which we can then more surgically get what we need and pass to other supporting actors.

The main impetus for this change was getting us slightly closer to an extensible interface for checkers. I think this cleans up the interface ever so slightly and thought it would make a nice little tidy up.

I returned a first class object that separately contains offenses and references.

Does this approach make sense? Is there a reason to pass offenses to the checkers that I'm missing?

## Type of Change

- [ ] Bugfix
- [ ] New feature

## Checklist

### DIFF
--- a/lib/packwerk/reference_extractor.rb
+++ b/lib/packwerk/reference_extractor.rb
@@ -51,20 +51,14 @@ module Packwerk
 
     sig do
       params(
-        unresolved_references_and_offenses: T::Array[T.any(UnresolvedReference, Offense)],
+        unresolved_references: T::Array[UnresolvedReference],
         context_provider: ConstantDiscovery
-      ).returns(T::Array[T.any(Reference, Offense)])
+      ).returns(T::Array[Reference])
     end
-    def self.get_fully_qualified_references_and_offenses_from(unresolved_references_and_offenses, context_provider)
-      fully_qualified_references_and_offenses = T.let([], T::Array[T.any(Reference, Offense)])
+    def self.get_fully_qualified_references_from(unresolved_references, context_provider)
+      fully_qualified_references = T.let([], T::Array[Reference])
 
-      unresolved_references_and_offenses.each do |unresolved_references_or_offense|
-        if unresolved_references_or_offense.is_a?(Offense)
-          fully_qualified_references_and_offenses << unresolved_references_or_offense
-
-          next
-        end
-
+      unresolved_references.each do |unresolved_references_or_offense|
         unresolved_reference = unresolved_references_or_offense
 
         constant =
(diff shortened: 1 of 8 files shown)


lib/packwerk/reference_extractor.rb
```diff
@@ -51,20 +51,17 @@ def reference_from_node(node, ancestors:, absolute_file:)
...
-      fully_qualified_references_and_offenses = T.let([], T::Array[T.any(Reference, Offense)])
-
-      unresolved_references_and_offenses.each do |unresolved_references_or_offense|
-        if unresolved_references_or_offense.is_a?(Offense)
-          fully_qualified_references_and_offenses << unresolved_references_or_offense
-
-          next
-        end
+    def self.get_fully_qualified_references_from(
+      unresolved_references,
```
rafaelfranca, line 59: Let's keep the arguments in the same line since. the line size isn't > 120 yet.

lib/packwerk/run_context.rb
```diff
@@ -76,13 +76,18 @@ def initialize(
...
+        processed_file.unresolved_references,
         context_provider
       )
       reference_checker = ReferenceChecking::ReferenceChecker.new(@checkers)
-      references_and_offenses.flat_map { |reference| reference_checker.call(reference) }
+
+      [
+        *processed_file.offenses,
+        *references.flat_map { |reference| reference_checker.call(reference) },
+      ]
```
rafaelfranca, line 90:
```suggestion
      processed_file.offenses + references.flat_map { |reference| reference_checker.call(reference) }
```

lib/packwerk/reference_extractor.rb
AlexEvanczuk: Changed!

AlexEvanczuk:
Updated @rafaelfranca , let me know if you have any other feedback!

exterm:
main monolith CI is green with this branch.

exterm:
Unless someone else swoops in and finds a problem, this looks ready to merge to me. It's a great improvement to some of the more confusing parts of packwerk's implementation. 

Can you just squash the commits @AlexEvanczuk ?

mclark (approved):
That mixture of offenses and references always bugged me, thanks so much for cleaning this up! I am definitely a fan of the `ProcessedFile` abstraction. I believe that is the right level to pull these types apart 👍

exterm:
re-ran monolith CI, still green.

PR #180, opened 2022-02-24
Change FilesForProcessing to return a T::Set[String]

AlexEvanczuk:

Change `FilesForProcessing` to return a Set, which makes fixing https://github.com/Shopify/packwerk/issues/154 simpler (see https://github.com/Shopify/packwerk/pull/178 for more info).

I returned a set, since we are already calling `uniq` on the returned files anyways. This makes testing for inclusion faster.
I used a `T.type_alias` to specify the type rather than a separate class because I wanted to continue to use the same `Set` interface but also be able to more explicitly refer to the types in signatures.

Any reason not to make this a set?

## Type of Change

- [ ] Bugfix
- [ ] New feature

## Checklist

### DIFF
--- a/lib/packwerk/files_for_processing.rb
+++ b/lib/packwerk/files_for_processing.rb
@@ -5,6 +5,8 @@ module Packwerk
   class FilesForProcessing
     extend T::Sig
 
+    AbsoluteFileSet = T.type_alias { T::Set[String] }
+
     class << self
       extend T::Sig
 
@@ -13,7 +15,7 @@ module Packwerk
           relative_file_paths: T::Array[String],
           configuration: Configuration,
           ignore_nested_packages: T::Boolean
-        ).returns(T::Array[String])
+        ).returns(AbsoluteFileSet)
       end
       def fetch(relative_file_paths:, configuration:, ignore_nested_packages: false)
         new(relative_file_paths, configuration, ignore_nested_packages).files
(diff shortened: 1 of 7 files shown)


lib/packwerk/files_for_processing.rb
```diff
@@ -47,19 +49,21 @@ def files
...
+    sig { returns(AbsoluteFileSet) }
     def custom_files
-      @custom_files ||= @relative_file_paths.flat_map do |relative_file_path|
-        absolute_file_path = File.expand_path(relative_file_path, @configuration.root_path)
-        if File.file?(absolute_file_path)
-          absolute_file_path
-        else
-          custom_included_files(absolute_file_path)
+      @custom_files ||= Set.new(
+        @relative_file_paths.flat_map do |relative_file_path|
```
rafaelfranca, line 55: Is this `flat_map` still relevant? Since the return of this block is sometimes a set isn't doing anything for that case anymore, hence the need to call `flatten` after the set is built. Maybe we should only call `flatten` after the set is built?

AlexEvanczuk: Agreed! Changed.

AlexEvanczuk:
> Instantiation set is often slower than arrays and we are doing a lot of instantiation in loops here. Did you measured to see if things are note getting slower by building all those sets?

Good question -- let me run my standard performance comparison on 20 iterations of packwerk check! I'll report back.

AlexEvanczuk:
@rafaelfranca 
I don't measure any meaningful difference in performance:

Screen Shot 2022-02-24 at 6 29 12 PM

exterm (approved):

exterm:
✅  main monolith CI is green with this branch

exterm (approved):

Issue #182, opened 2022-03-07
[Bug Report] `Errno::ENOENT: No such file or directory @ rb_sysopen - tmp/cache/packwerk/inflections` (issue with cache)


AlexEvanczuk:
**Description**
The cache periodically (rarely) fails in our CI system with the error shown in the screenshots section.

We suspect that the packwerk cache has a race condition when forked processes are initializing the cache independently.
Namely -- after one process has bust the cache and created the cache directory, another process can then bust the cache, after which the original process may try to write the cache contents of `inflections.rb`. This can result in the bug:
`Errno::ENOENT: No such file or directory @ rb_sysopen - tmp/cache/packwerk/inflections`

**To Reproduce**
We are not able to reliably reproduce this bug.

**Expected Behaviour**
We expect the cache to not raise an error. From an implementation perspective, we expect `tmp/cache/packwerk` to already exist before attempting to write the cache file.

**Screenshots**
<img width="909" alt="Screen Shot 2022-03-07 at 1 53 57 PM" src="https://user-images.githubusercontent.com/3311200/157099216-e459ed75-1646-4d04-a19d-b83ae6775101.png">

**Version Information**
 - Packwerk: 2.1.0
 - Ruby: 2.7.5

**Additional Context**
Some potential solutions we're looking at to solve this issue:

- We can initialize `Cache` within the parent process to ensure that the call to `FileUtils.rm_fr` is only run once at the start. This is what we went with in this PR: https://github.com/Shopify/packwerk/pull/183
- Alternatively, we can also drop the `rm_fr` entirely, and use a strategy that hashes based on the digest of the current inflections.

exterm:
going to close this now that the candidate fix is released with [2.1.1](https://github.com/Shopify/packwerk/releases/tag/v2.1.1). Let's reopen should the error reoccur.

PR #183, opened 2022-03-07
Fix race condition in cache implementation

AlexEvanczuk:
See https://github.com/Shopify/packwerk/issues/182

More info in issue

Is this an acceptable way to resolve cache race conditions for the short/medium term?

## Type of Change

- [x] Bugfix
- [ ] New feature

## Checklist

### DIFF
--- a/lib/packwerk/run_context.rb
+++ b/lib/packwerk/run_context.rb
@@ -71,7 +71,10 @@ module Packwerk
 
       @file_processor = T.let(nil, T.nilable(FileProcessor))
       @context_provider = T.let(nil, T.nilable(ConstantDiscovery))
-      @cache = T.let(nil, T.nilable(Cache))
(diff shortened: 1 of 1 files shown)


exterm:
Is the problem occurring frequently enough in your CI that you can meaningfully test this fix?

AlexEvanczuk:
@exterm It's pretty rare in our CI @exterm -- we're not able to reproduce reliably and haven't even with a scheduled build.

Let me know if you feel hesitations in shipping this and we can think of other strategies.

exterm:
I just encountered it in our CI 🤣 

Yeah I think this doesn't make anything worse (can you think of any drawbacks?) so it's probably fine to just ship it

exterm (approved):

AlexEvanczuk:
Thanks @exterm ! Maybe we can cut 2.1.1 for this one?

PR #189, opened 2022-03-26
Add flow chart to help explain how to think about and resolve new violations

AlexEvanczuk:
## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist


USAGE.md
```diff
@@ -19,6 +19,8 @@
 * [Checking for violations](#checking-for-violations)
 * [Recording existing violations](#recording-existing-violations)
   * [Understanding the list of deprecated references](#understanding-the-list-of-deprecated-references)
+* [Resolving New Violations](#resolving-new-violations)
+  * [Understanding how to respond to new violations](#understanding-how-to-respond-to-new-violations)
```
exterm, line 23: nitpick but I would probably put this _before_ the recording violations section

mclark:
On a related note, GitHub now natively supports mermaid.js as of February this year https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/

exterm:
Sorry - yes I can now see it. Just haven't had time to review.

AlexEvanczuk:
> Sorry - yes I can now see it. Just haven't had time to review.

Ok no rush @exterm ! Thank you !

exterm:
Other than that, it looks great to me. Would be interested in @rafaelfranca opinion.

AlexEvanczuk:
Wow that was a very redundant box @exterm . The general idea is that when I recommend people use packwerk to modularize, I say there are two steps:
1) State what your boundaries by creating packages, putting files in them, exposing public API, and stating dependencies
2) Allow packwerk to reflect on how accurate the stated boundaries are via violations.

If (1) doesn't make sense, then we shouldn't try to "reduce violations," we should first iterate on the quality of the boundaries.

I also changed the gusto specific reference to `packs` and said "its public folder."

Here's the CSV of shape data. As mentioned before it doesn't appear to be importable into lucid chart, but the intent is to at least reduce the cost of changing this documentation by having the text available:
[Packwerk Violation Flow Chart.csv](https://github.com/Shopify/packwerk/files/8400813/Packwerk.Violation.Flow.Chart.csv)

exterm:
> If (1) doesn't make sense, then we shouldn't try to "reduce violations," we should first iterate on the quality of the boundaries.

I love that framing.

rafaelfranca:
I'm not sure if there is much value of including a link to a PDF in our readme. I really doubt people will read this. Maybe we should move to the wiki?

AlexEvanczuk:
Wanted to bump this @rafaelfranca

AlexEvanczuk:
Good points @rafaelfranca . The current form is not amenable to change which is a pretty bad property for documentation!

I'll turn this flow chart into a new file. I'll try to represent the flow chart but as a bulleted list instead of a PDF, and I'll try to make it a bit more concise as well. From there we can make edits as we see fit!

Might take a little bit -- about to head OOO for 2 weeks but might pop in and push that change.

rafaelfranca (approved):

Issue #186, opened 2022-03-17
Packwerk doesn't support custom root namespaces for root directories in the load path


mclark:
In Zeitwerk, the load path is list of directories that it calls "root directories". They are "root" because any files that reside in them are expected to represent modules that are part of the root namespace. If an object is nested in some more specific namespace, it needs to be placed in a subdirectory of one of these roots. For example, in the following, `lib` is a "root directory" with the following structure and the ruby files have the noted expected constants:

```
.
├── lib
│   ├── my_namespace
│   │   └── my_file.rb            # MyNamespace::MyFile
│   └── my_other_file.rb          # MyOtherFile
```

The thing is, Zeitwerk also supports setting a "default namespace" to each root directory via [`push_dir`](https://github.com/fxn/zeitwerk/blob/03529893edd471ad4b05c06577b3fab8ab087dc7/lib/zeitwerk/loader/config.rb#L107-L111) which is called when the Zeitwerk loader is being configured. So in the above, it was assumed `lib` was added to the loader via `loader.push_dir("lib")` but it could also be added with `loader.push_dir("lib", MyCompany)` which means the constants inside would be expected to be declared as `MyCompany::MyNamespace::MyFile` and `MyCompany::MyOtherFile`.

At GitHub we rely on this feature to work around some extremely legacy directories in our load path that do not conform to Zeitwerk's expectations. Up until now, we have worked around them in Packwerk by skipping some and "hacking" at others. At this point, we'd like to just add support for custom namespaces to Packwerk so we can be confident in Packwerk's operation.

While I haven't dived into the Packwerk's source for this yet, I suspect this would mean moving [`Packwerk::ApplicationLoadPaths`](https://github.com/Shopify/packwerk/blob/1876786fc8584839f955f09d59b2773605890e58/lib/packwerk/application_load_paths.rb#L27) from `config.autoload_paths`, `eager_load_paths`, and `autoload_once_paths` to `Rails.autoloaders` to get our list of root directories as a hash of paths to modules. By default, these modules would all just be `Object` as they are in Zeitwerk, but it would allow us to override that default as needed for certain paths.

I realize this would also necessitate a change to `ConstantResolver` as well and have a WIP branch [here](https://github.com/Shopify/constant_resolver/compare/master...mclark:default-namespace-support). This branch uses strings instead of module instances for the namespaces in order to keep us as disentangled from a dependency standpoint as possible.

[...]

PR #190, opened 2022-04-01
Load paths hash

mclark:

As described in https://github.com/Shopify/packwerk/issues/186, Zeitwerk supports the notion of a "default" namespace for each root directory. This is represented by a `Hash` of root directory path strings to constants representing the default module in `Zeitwerk::Loader#root_dirs`. This pull request updates Packwerk so that it treats the group of load paths as a hash rather than an Array and uses the `#root_dirs` of each autoloader in `Rails.autoloaders` instead.

As Packwerk relies on Rails' `config.autoload_paths` and friends lists to get the list of load paths to draw on for its constant discovery process, and these are just lists of strings, they needed to be abandoned in favour of relying on the autoloaders instead which are hashes of root dir paths to default modules. This triggered a number of changes, mostly to method signatures as the new hash worked its way through the call stacks to eventually arrive in the `ConstantResolver` gem (already updated in https://github.com/Shopify/constant_resolver/pull/27).

After that, most of the effort just went into updating the tests to conform to the new expectations:

* Removed `RailsPaths` in favour of using plain old Mocha mocks
* Removed uniqueness and method signature tests that were impossible to fail thanks to the use of Hashes and Sorbet
* Initialized the dummy application so `Rails.autoloaders` would be populated
* Changed two tests from `ActiveSupport::TestCase` to `Minitest::Test` to be consistent with all other tests and to suppress log output caused by initializing the dummy application.

I believe @exterm is bumping `constant_resolver` as I write this so expect the git dependency for `constant_resolver` to go away and be replaced once that is done.

I am not an expert on rails or load paths. Is eager loading still supported properly? Can we suppress the logging in a better way than outputting "nothing" to standard out?

## Type of Change

- [ ] Bugfix
- [x] New feature

### Additional Release Notes

## Checklist


Gemfile
```diff
@@ -9,7 +9,8 @@ gemspec
 
 gem("spring")
 gem("rails", github: "rails/rails", branch: "main")
-gem("constant_resolver", require: false)
+gem("constant_resolver", git: "https://github.com/Shopify/constant_resolver.git",
+    ref: "ba7fd11288566e2ecdd70e0946dcfb20c5341b27", require: false)
```
exterm, line 13: 0.2.0 is released now.

mclark: Updated! Thank you! ❤️

test/support/rails_application_fixture_helper.rb
```diff
@@ -28,36 +18,28 @@ def use_template(template)
...
+    Rails.application.config.stubs(:root).returns(root)
   end
 
   private
 
   def set_load_paths_for_minimal_template
-    RailsPaths.autoload(to_app_paths("/components/sales/app/models"))
-    RailsPaths.eager_load([])
-    RailsPaths.autoload_once([])
+    main = Zeitwerk::Loader.new
```
mclark, line 28: I overlooked the fact that this is introducing a direct dependency on Zeitwerk here until I saw [these tests failing](https://github.com/Shopify/packwerk/runs/5792482153?check_suite_focus=true). Given that Packwerk depends on Zeitwerk's structural assumptions (and indeed is even named after it?) I don't think adding an explicit dependency on Zeitwerk would be that much of a stretch. However, I'm tempted to just use a mock object since we only need the one `:root_dirs` method here anyway.

Do others have thoughts on the best direction here?

lib/packwerk/application_load_paths.rb
```diff
@@ -18,36 +18,30 @@ def extract_relevant_paths(root, environment)
...
-        Rails.application.railties
-          .select { |railtie| railtie.is_a?(Rails::Engine) }
-          .push(Rails.application)
-          .flat_map do |engine|
-            paths = (engine.config.autoload_paths + engine.config.eager_load_paths + engine.config.autoload_once_paths)
-            paths.map(&:to_s).uniq
-          end
+        Rails.autoloaders.inject({}) do |h, loader|
+          h.merge(loader.root_dirs)
+        end
```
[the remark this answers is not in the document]
mclark: So it turns out this doesn't work in Rails 6.0 unless `Rails.autoloaders.zeitwerk_enabled?` returns `true`. If it is false, `Rails.autoloaders` will be an empty enumerable.

So now we have to choose one of:

* For all the places that we stub out `config.autoload_paths` and friends we add matching tests to cover when zeitwerk is both enabled and disabled.
* Explicitly add the requirement that packwerk is only compatible with Rails applications which are zeitwerk enabled. This is the case for Rails 6.x codebases that have `Rails.autoloaders.zeitwerk_enabled? == true` or Rails 7.0+ codebases.

Personally, based on our discussions about adding Zeitwerk as a dev dependency in the [test in this diff](https://github.com/Shopify/packwerk/pull/190/files#r840817911) I feel we could pretty safely say that Packwerk requires `zeitwerk_enabled?` to be true. This would retain Rails 6.x support without requiring us to duplicate a whole bunch of test.

As ["Classic mode"](https://github.com/rails/rails/commit/0d523d83657ce7066f25d87f6f094e804590e1e8) isn't coming back, and Packwerk is based on the conventions of Zeitwerk, unless this causes problems for any known users, I'd suggest we go with the addition of this "soft" requirement.

test/rails_test_helper.rb
```diff
@@ -16,4 +18,7 @@ def self.skeleton(*path)
     skeleton("vendor", "cache", "gems", "example", "models"),
   ]
   config.root = skeleton(".")
+  config.logger = Logger.new(STDOUT)
 end
+
+T.cast(Dummy, Rails::Application).initialize!
```
exterm, line 24: This assumes `Dummy` to be a value of type `Rails::Application`, but it is not - it is a subclass.

mclark:
> Looks like we don't have a type spec for Rails::Application... maybe tapioca gem rails can fix it?

I couldn't really get it working, probably my lack of experience with `tapioca`.. Anyway, using `Rails.application.initialize!` instead seems to have done the trick 👍

lib/packwerk/application_load_paths.rb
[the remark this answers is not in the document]
exterm: Maybe we should make it more obvious, but packwerk already officially requires zeitwerk to be enabled https://github.com/Shopify/packwerk#prerequisites

I'll accept a PR to reword that to make it more obvious

test/rails_test_helper.rb
```diff
@@ -8,6 +8,8 @@ def self.skeleton(*path)
     ROOT.join("test", "fixtures", "skeleton", *path).to_s
   end
 
+  config.autoloader = :zeitwerk
```
exterm, line 11: 👍🏼 This is the correct solution IMO

test/const_node_inspector_test.rb
```diff
@@ -5,7 +5,7 @@
 require "parser_test_helper"
 
 module Packwerk
-  class ConstNodeInspectorTest < ActiveSupport::TestCase
+  class ConstNodeInspectorTest < Minitest::Test
```
exterm, line 8: why this change?

exterm (approved):

test/const_node_inspector_test.rb
mclark: The tests in these two classes started writing formatted output, probably as a result of the Dummy app being initialized. This initially created a log file in the dummy skeleton until I changed it to `stdout`. To silence them completely, I figured I'd just make them consistent with the rest of the tests.
Alternatively, maybe we could configure the Dummy app with some sort of `NullLogger` I suppose?

exterm: That seems like a good solution 🤔  TBH I don't remember why these tests are using `ActiveSupport::TestCase` and others do not. But it seems like it would be confusing for someone to start using `ActiveSupport::TestCase` and thereby inadvertently generating logs.

mclark: Done in 805a77d14179e742ada7ff7a51ad6f19171e7329 👍

exterm:
our main monolith CI is green against this branch. I've also done some manual testing and it's looking great.

PR #194, opened 2022-04-14
Filter custom paths according to include

thomasmarshall:

Custom paths are [already filtered by the `exclude` option](https://github.com/Shopify/packwerk/blob/72209689adef3f6c55776a2a948999ef42ec3819/lib/packwerk/files_for_processing.rb#L47), but this commit ensures that the `include` option is respected too. This is particularly useful for the `packwerk-vscode` extension, which [attempts to run](https://github.com/Gusto/packwerk-vscode/blob/b51eb9535dd95e17a7f17fc618dd7beab449a979/package.json#L58) `packwerk check $filename` on save.

For example:

```yml
include:
  - a/**/*.rb
  - b/**/*.rb
```

I would expect `packwerk check c/some/file.rb` to show the "No files found or given" message, as it would if `exclude` was set. This commit ensures that happens.

I chose to always filter custom paths according to the `include` option, because that seemed to mirror the behaviour of the `exclude` option. However, there might be valid use-cases for `packwerk check some/not/included/file.rb` that I haven't considered. If that's a common use-case, maybe I should approach this differently? Maybe the `packwerk-vscode` extension should be making decisions about when to run `packwerk check` instead?

Are there common/valid use-cases for `packwerk check` that I am making impossible with this change? I guess if someone is relying on the existing behaviour then this might be a breaking change?

## Type of Change

- [x] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist

### DIFF
--- a/lib/packwerk/files_for_processing.rb
+++ b/lib/packwerk/files_for_processing.rb
@@ -41,7 +41,7 @@ module Packwerk
       include_files = if custom_files.empty?
         configured_included_files
       else
-        custom_files
+        configured_included_files & custom_files
       end
 
       include_files - configured_excluded_files
--- a/test/unit/files_for_processing_test.rb
+++ b/test/unit/files_for_processing_test.rb
@@ -76,6 +76,19 @@ module Packwerk
       refute_any_match(files, Set.new([File.join(@configuration.root_path, "components/timeline", "**/*.rb")]))
     end
 
+    test "fetch with custom paths for files includes only include glob in custom paths" do
+      files = ::Packwerk::FilesForProcessing.fetch(
+        relative_file_paths: [
+          "components/sales/app/models/order.rb",
+          "components/sales/app/views/order.html.erb",
+        ],
+        configuration: @configuration
+      )
+      included_file_patterns = @configuration.include.map { |pattern| File.join(@configuration.root_path, pattern) }
+
+      assert_all_match(files, included_file_patterns)
+    end
+
     private
 
     def assert_all_match(files, patterns)


exterm:
It is a breaking change IMO, so at least we should bump the version accordingly.

@AlexEvanczuk @mclark Would this cause any problems at Gusto or Github?

AlexEvanczuk:
I tried this on Gusto's codebase and `bin/packwerk update-deprecations` produced no changes to `deprecated_references.yml` files.

Note Gusto is also not yet on the version of sorbet that packwerk 2.1.1 requires, but to test this on our codebase, I just pulled this branch locally and made some small tweaks. Wanted to include what I did in case GitHub or Shopify needs to do the same.

### In `packwerk`
#### In `packwerk.gemspec`
1. Downgraded `sorbet-runtime to the version in our application, in our case:
```ruby
spec.add_dependency("sorbet-runtime", '0.5.9889')
```

### In our application
#### `Gemfile`
1. Moved `packwerk` to:
```ruby
path('../packwerk') do
  gem 'packwerk'
end
```

2. `bundle update constant_resolver --conservative` (we were on 0.1.5)
2. `bundle install`

mclark:
@thomasmarshall I needed to merge `main` into this branch to verify your fix against the GitHub monolith as we now require 2.2.0 and it appears this branch is based on a slightly older commit. I then tested with just an include path to verify your changes took (they did!) and then ran `bin/packwerk update` and `bin/packwerk update-deprecations` on our standard configuration. Everything appears to be working as expected 🎉 

Thanks for the fix @thomasmarshall! 👍

gmcgibbon:
This is still a breaking change, but I think it is a good change. The next release will have to be `3.0.0` to denote the breaking change.

PR #195, opened 2022-04-15
make FilesForProcessing return relative paths

AlexEvanczuk:

The intent of this PR is to get us closer to being able to fix the issue with `bin/packwerk check` not working when run on a single file (it mistakenly picks up stale violations).

To do this, we want to compare the files passed into `ParseRun` against what is in the `deprecated_references.yml` files.

All file references in `deprecated_references.yml` files are paths *relative to the root*. Therefore by changing parse run to use relative paths, we:
A) Simplify some of the code which no longer has to deal with translating to and from absolute paths.
B) Allows us to compare against `deprecated_references.yml` in a follow-up PR.

Am I missing something where the behavior relies on absolute paths?

## Type of Change

- [ ] Bugfix
- [ ] New feature

## Checklist

### DIFF
--- a/lib/packwerk/cli.rb
+++ b/lib/packwerk/cli.rb
@@ -128,17 +128,17 @@ module Packwerk
       params(
         relative_file_paths: T::Array[String],
         ignore_nested_packages: T::Boolean
-      ).returns(FilesForProcessing::AbsoluteFileSet)
+      ).returns(FilesForProcessing::RelativeFileSet)
     end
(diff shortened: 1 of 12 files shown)


exterm (approved):
looks great!

AlexEvanczuk:
Thanks for merging @exterm . I'm OOO the rest of the week, but I'll try to finish off this bug fix next or the following week.

Issue #192, opened 2022-04-08
Any way of warning when extraneous dependencies exist in a package.yml file?


oskarpearson:
**Description**

Running `packwerk check` should ideally show a warning/error when a package.yml file includes items in the `dependencies:` block that are not actually dependencies in the code.

**To Reproduce**

Start with a codebase with multiple packages in it, and where packwerk check and packwerk validate are both passing as expected.

Modify one of the package's package.yml files (package_a) to add a dependency on a different package (package b). Ensure that you don't create any cyclic loops in this experiment. Don't make any ruby code changes - the intention is to just add an extraneous dependency to package_b

Run packwerk validate and packwerk check, and note that the output indicates that the configuration is valid.

**Expected Behaviour**

I'd expect a warning or error saying that there's an extraneous/unused dependency on package_b in package_a.

**Screenshots**

N/A

**Version Information**

 - Packwerk: v2.1.1
 - Ruby 2.7.4p191

**Additional Context**

As our developers change code, they sometimes inadvertently remove the code that previously triggered a dependency. Ideally, packwerk would let them know this, so that they can remove the obsolete line from the package.yml

exterm:
🤔  Maybe it's because most of our packages are huge, but I've rarely seen this happen.

There's also the complication that packwerk so far doesn't have a concept of warnings, and I wouldn't want to make this an error.

That said, I understand how it could be useful.

PR #196, opened 2022-04-15
Move checker specific logic into checker

AlexEvanczuk:
The idea of packwerk being extendible to incorporate new "checkers" has been bounced around for a while now. 

Certain new issues that have come up (e.g. https://github.com/Shopify/packwerk/issues/192) demonstrate the desire for this sort of capability (in this case, the checker would enforce that we only list dependencies if there is at least one reference to a pack).

Other interesting checkers could be:
- visibility checker: allow packages to specify who they are visible to support using packages to divide up subsystems where only one pack contains the externally facing public API
- layered architecture: allow packages to specify what "layer" they are, such as "utilities," "platform", "business logic" and more. The user could configure the permitted order and the checker could ensure that utilities never rely on the business logic.
- serializable types at the boundaries: if we added a piece of data into reference (if the reference is defined in a sorbet signature), a client could configure a checker to ensure that public APIs only expose serializable types, such as `T::Struct`, at the boundaries, rather than ActiveRecord objects.

To make this possible, we need to make sure a couple of things are true:
### ✅ We want to make sure any part any checker logic is colocated within the checker. 

`lib/packwerk/reference_offense.rb` right now builds a message based on the violation type. We want this because in a world where the other checkers can exist, those checkers define their own message. Similarly, we want to make sure that no other part of the system is coupled to checker *implementation* and only to the interface.

### Packwerk contains an API to inject new checkers.
There are a variety of ways to support this. The simplest might be to support a `Packwerk.configure { |config| config.checkers << [...] }` type of API.

### The interface for a checker is sufficiently flexible to support a variety of checks.
Note that file processing runs in parallel, but checking happens in serial. This means each checker ends up receiving every reference. I have a PR to change the interface for a checker to receive the array of every single reference. This would support checkers that can do things like the above (e.g. check if pack A refers to any constants in pack B and if not, creates a stale dependency offense).

Is there any reason why we wouldn't want this behavior in the checker itself?

## Type of Change

- [ ] Bugfix
- [ ] New feature

## Checklist

### DIFF
--- a/lib/packwerk/reference_checking/checkers/checker.rb
+++ b/lib/packwerk/reference_checking/checkers/checker.rb
@@ -8,13 +8,26 @@ module Packwerk
         extend T::Sig
         extend T::Helpers
 
-        interface!
+        abstract!
 
         sig { returns(ViolationType).abstract }
         def violation_type; end
 
         sig { params(reference: Reference).returns(T::Boolean).abstract }
         def invalid_reference?(reference); end
+
+        sig { params(reference: Reference).returns(String).abstract }
+        def message(reference); end
+
+        sig { params(reference: Reference).returns(String) }
+        def standard_help_message(reference)
+          standard_mesage = <<~EOS
+            Inference details: this is a reference to #{reference.constant.name} which seems to be defined in #{reference.constant.location}.
+            To receive help interpreting or resolving this error message, see: https://github.com/Shopify/packwerk/blob/main/TROUBLESHOOT.md#Troubleshooting-violations
+          EOS
+
+          standard_mesage.chomp
+        end
       end
     end
   end
--- a/test/unit/offense_collection_test.rb
+++ b/test/unit/offense_collection_test.rb
@@ -9,7 +9,11 @@ module Packwerk
 
     setup do
       @offense_collection = OffenseCollection.new(".")
-      @offense = ReferenceOffense.new(reference: build_reference, violation_type: ViolationType::Dependency)
+      @offense = ReferenceOffense.new(
+        reference: build_reference,
+        violation_type: ViolationType::Dependency,
+        message: "some message"
+      )
     end
 
     test "#add_violation adds entry and returns true" do
@@ -53,12 +57,22 @@ module Packwerk
         .with(package, "./buyer/deprecated_references.yml")
         .returns(deprecated_references)
 
-      offense = Packwerk::ReferenceOffense.new(reference: reference, violation_type: ViolationType::Dependency)
+      offense = Packwerk::ReferenceOffense.new(
+        reference: reference,
+        violation_type: ViolationType::Dependency,
+        message: "some message"
+      )
+
       assert @offense_collection.listed?(offense)
     end
 
     test "#listed? returns false if constant is not listed in file " do
-      offense = Packwerk::ReferenceOffense.new(reference: build_reference, violation_type: ViolationType::Dependency)
+      offense = Packwerk::ReferenceOffense.new(
+        reference: build_reference,
+        violation_type: ViolationType::Dependency,
+        message: "some message"
+      )
+
       refute @offense_collection.listed?(offense)
     end
   end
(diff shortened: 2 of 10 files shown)


test/unit/offense_collection_test.rb
```diff
@@ -53,12 +57,22 @@ class OffenseCollectionTest < Minitest::Test
...
+
       assert @offense_collection.listed?(offense)
     end
 
     test "#listed? returns false if constant is not listed in file " do
-      offense = Packwerk::ReferenceOffense.new(reference: build_reference, violation_type: ViolationType::Dependency)
+      offense = Packwerk::ReferenceOffense.new(
+        reference: build_reference,
+        violation_type: ViolationType::Dependency,
+        message: "some message"
```
exterm, line 73: almost feels like there is a `Violation` type missing here that could generate the message on demand, so that instead of injecting a string, we would inject the capability to generate the string 🤔

exterm: But that can be a separate PR, if you agree it makes sense. And maybe it is a premature abstraction, I haven't thought too much about it.

The smell for me is that we're injecting two things that are tightly coupled to each other. We would never pass a `ViolationType::Dependency` with a message for a privacy error.

AlexEvanczuk: For sure, I think this speaks to Matt's thought: https://github.com/Shopify/packwerk/pull/196#discussion_r853115253

I think what might make most sense would be the checkers create the offense themselves rather than calling a bunch of separate checker interface methods. I do think there is value in having a reference offense have a `violation_type` as part of its API since I can see a client wanting to react differently to different types of violations.

I agree that we'd never want to pass a dependency violation type for a privacy error, although I'm not sure how we can prevent this systematically while simultaneously allowing clients to know what type of violation type it is 🤔

test/unit/parse_run_test.rb
```diff
@@ -73,12 +74,17 @@ class ParseRunTest < Minitest::Test
...
+        message: "some message",
+        violation_type: ViolationType::Privacy
+      )
+
       DeprecatedReferences.any_instance.stubs(:listed?).returns(true)
       out = StringIO.new
       parse_run = Packwerk::ParseRun.new(
         absolute_file_set: Set.new(["some/path.rb"]),
-        configuration: Configuration.new({ "parallel" => false }),
+        configuration: Configuration.new({ "parallel" => false }, config_path: nil),
```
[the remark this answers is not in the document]
AlexEvanczuk: Huh -- I'm not sure. I removed it!

exterm:
This looks good to me, I just need to find a minute to run it in monolith CI

lib/packwerk/reference_checking/checkers/checker.rb
```diff
@@ -8,13 +8,26 @@ module Checker
...
+        def message(reference); end
+
+        sig { params(reference: Reference).returns(String) }
+        def standard_help_message(reference)
+          standard_mesage = <<~EOS
+            Inference details: this is a reference to #{reference.constant.name} which seems to be defined in #{reference.constant.location}.
+            To receive help interpreting or resolving this error message, see: https://github.com/Shopify/packwerk/blob/main/TROUBLESHOOT.md#Troubleshooting-violations
+          EOS
+
+          standard_mesage.chomp
```
rafaelfranca, line 29: Why the chomp?

rafaelfranca (approved):
Thank you @AlexEvanczuk. I tested against our app and it passes all checks. Feel free to merge this PR yourself.

lib/packwerk/reference_checking/checkers/checker.rb
AlexEvanczuk: Every HEREDOC adds a new line at the bottom. Before this was all part of a single heredoc with one new line. Now this heredoc is interpolated in another heredoc for each checker, which added two lines, so I removed the extra here to maintain the same output (as determined by the expectations of the tests).

Issue #49, opened 2020-10-23
Update command should `deprecated_references.yml` file when violations are gone


wildmaples:
When all the violations are resolved from a package, `packwerk update` does not delete the file, leaving the file stale. `packwerk update` should somehow be able to remove stale files completely.

rafaelfranca:
TODO:

* [ ] Finish https://github.com/Shopify/packwerk/compare/main...rm-fix-packages-with-no-violations with feedback from https://github.com/Shopify/packwerk/pull/199#issuecomment-1199785945

PR #199, opened 2022-04-20
Delete deprecated references files when empty

mclark:

I'm trying to fix https://github.com/Shopify/packwerk/issues/49. In short, if a package _had_ deprecated references, but they have since been resolved, the `deprecated_references.yml` file should be removed.

I altered `ParseRun#update_deprecations` to first _delete_ the `deprecated_references.yml` files, if they exist, of each package in the `RunContext#package_set`. This way the existing logic will only recreate the files which still have deprecations.

I originally tried a different approach to this. I think that fundamentally `Package` objects should own their `DeprecatedReferences` objects, but that turned into a much larger refactor that isn't really needed quite yet, at least I didn't think so. This PR instead at least addresses the core pain point and I didn't feel that the tech debt potential was onerous.

Are we cool with the `ParseRun` cleaning up existing deprecated references files before asking the deprecated reference instances it has to write out their data?

## Type of Change

- [x] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist


lib/packwerk/parse_run.rb
```diff
@@ -45,6 +46,12 @@ def detect_stale_violations
     sig { returns(Result) }
     def update_deprecations
       offense_collection = find_offenses
+
+      @run_context.package_set.each do |package|
```
AlexEvanczuk, line 50: I wonder if we could do a small tidy to reuse some things in `lib/packwerk/offense_collection.rb`...

I was thinking we could have a method:
```
# replaces `deprecated_references_for`, a private method in `OffenseCollection`
deprecated_references = Packwerk::DeprecatedReferences.for(package)
```

Then we could call:
```
deprecated_references.delete_if_exists
```

Which is of course just:
```
File.delete(@filepath) if File.exist?(@filepath)
```

Also note that in `DeprecatedRefererences#dump`, I think we can remove the `if @new_entries.empty` portion given this new implementation.

AlexEvanczuk:
This seems like a nice solution for this issue.

The point about packages owning their own deprecated references files is interesting although I'm not sure how it would solve it in this case. We'd probably need to something similar to what I'm trying to do to solve the "packwerk check fails on a single file" issue where we compare the files passed into `update-deprecations` with the files in a package and/or the files represented in `deprecated_references.yml`. However I think approach is a lot more complex here so I really like what you have here, especially since I think that was the intent of `if @new_entries.empty; # delete the file`, which I believe is actually dead code.

lib/packwerk/parse_run.rb
mclark: I'd really like to be able to just be able to do something like `package_set.each { |package| package.deprecated_references.persist! }` but I think we're just too far away from that. I can do the change you suggested, it does at least centralize the deprecated reference file IO in the one class which I like. Thanks for taking the time to look at this!

mclark:
Yeah I agree it would have been much better! Unfortunately, deprecated references are only instantiated in `OffenseCollection` for the packages that have violations. So when we iterate over these files, we only know about the ones that should still exist! This is why @AlexEvanczuk suspected it was dead code.

Instead of passing in the `root_path` to the `OffenseCollection`, we could pass in a `PackageSet` and use that to build the deprecated references for all the packages in the project. This would then cause that code to no longer be dead and it would handle deleting the files.

Thoughts on this proposal?

exterm:
> pass in a PackageSet and use that to build the deprecated references for all the packages in the project.

We wouldn't have offenses for all packages though if we don't check all of them, right?

exterm:
Actually, that is a good point - what happens if we run `update-deprecations` on _parts_ of the code base? Would this PR still delete all deprecation files?

I think what we're missing is a concept of "which packages are being checked", which is not super trivial considering that we can specify files to check, and thus partial packages.

Reminds me of the work @AlexEvanczuk has been doing for fixing false positive stale violations.

exterm:
Some of our packages are so large that checking single files is a convenient way to decrease wait times locally. Also, the VS Code integration relies on packwerk analyzing single files.

exterm:
I guess what we could do is say that `update-deprecations` can not take single files.

rafaelfranca:
I don't remember why I didn't finish this branch https://github.com/Shopify/packwerk/compare/main...rm-fix-packages-with-no-violations (I don't even remember if it works), but I think it is a simpler solution for this problem. What do you think?

AlexEvanczuk:
Hey @rafaelfranca I like that approach too. My only two pieces of feedback are:
- Iterating over all files (40K+) and using `Pathname#relative_path_from` can actually be pretty slow. Not sure how slow in practice but may be worth benchmarking
- Right now we implicitly "register" targeted deprecated references files because we memoize the deprecated references file in a hash one we get it. I wonder if we could/should make this a bit more explicit internally so its easier to see what is going on.

Want to open a PR from that branch?

PR #201, opened 2022-04-21
Bump sorbet and sorbet-runtime to address change of public API on newest version of sorbet-runtime

AlexEvanczuk:
- bundle update sorbet sorbet-runtime --conservative
- bundle update sorbet static
- fix issue
- Set explicit version for sorbet runtime

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist


lib/packwerk/package_set.rb
```diff
@@ -12,7 +12,7 @@ class PackageSet
     extend T::Generic
     include Enumerable
 
-    Elem = type_member(fixed: Package)
+    Elem = type_member { { fixed: Package } }
```
exterm, line 15: Ruby syntax strikes again... This is a block returning a hash. It took me a second to realize that.

rporrasluc (approved):

PR #202, opened 2022-05-23
Remove `digest` dependency from gemspec

iMacTia:
`digest` is a standard gem and as such is bundled with Ruby 3.0+, while in previous Ruby versions it's part of stdlib.
Having the gem explicitly added to the gemspec can cause issues downstream on projects using older ruby versions.
This is because it's actually possible to load the code twice, once from the stdlib and once from the gem.

Removing the dependency should have absolutely no effect on the code.
For people running Ruby 2.7 or previous, the `require "digest"` in `cache.rb` will load the stdlib file, while for those running Ruby 3.0+ it will load the standard gem.

Cross-compatibility with Ruby 2 and Ruby 3 

## Type of Change

- [x] Bugfix
- [ ] New feature

## Checklist

cc @rafaelfranca @AlexEvanczuk

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -93,7 +93,6 @@ PATH
       better_html
       bundler
       constant_resolver (>= 0.2.0)
-      digest
       parallel
       parser
       sorbet-runtime (>= 0.5.9914)
@@ -119,7 +118,6 @@ GEM
     concurrent-ruby (1.1.8)
     constant_resolver (0.2.0)
     crass (1.0.6)
-    digest (3.1.0)
     erubi (1.10.0)
     globalid (0.4.2)
       activesupport (>= 4.2.0)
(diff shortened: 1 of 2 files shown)


iMacTia:
CLA is signed, not sure how to re-run the CI now though 😄

AlexEvanczuk:
This looks fine to me, thanks @iMacTia !

gmcgibbon:
> while for those running Ruby 3.0+ it will load the standard gem.

```
> gem uninstall digest
Gem digest-3.1.0 cannot be uninstalled because it is a default gem
```

Hm, so `digest` is a default gem that can't be uninstalled. TIL. Seems kind of confusing that you can load default gems twice, but I guess Ruby 2 has no knowledge of this new paradigm and since [the gemspec of digest supports Ruby 2](https://github.com/ruby/digest/blob/f8f7c6a2c5950371f3fef27d624c1f92294d864b/digest.gemspec#L26), it just installs it like any other gem.

Issue #209, opened 2022-07-11
[Bug Report] `packwerk validate` fails with uninitialized constant <Class:Packwerk::PackageSet>::Bundler


dmitri-minkin:
**Description**
`packwerk validate` fails on minimalistic new Rails app

**To Reproduce**

- Create new folder for the application
- In that folder create new Rails app: `rails new sportsball`
- Add `gem 'packwerk'` to Gemfile
- Change ruby version to 3.1.2 in .ruby-version and Gemfile
- Call bundler to install gems: `bundle install`
- Add packwerk binstub: `bundle binstub packwerk`
- Initialize packwerk: `bin/packwerk init`
- Validate that packwerk is installed correctly: `packwerk validate`

Notice the failure:
> 📦 Packwerk is running validation...
/Users/{user}/.gem/ruby/3.1.2/gems/packwerk-2.2.0/lib/packwerk/package_set.rb:45:in `package_paths': uninitialized constant #<Class:Packwerk::PackageSet>::Bundler (NameError)

          .push(Bundler.bundle_path.join("**").to_s)
                ^^^^^^^
        from /Users/{user}/.gem/ruby/3.1.2/gems/sorbet-runtime-0.5.10151/lib/types/private/methods/call_validation.rb:161:in `bind_call'
        from /Users/{user}/.gem/ruby/3.1.2/gems/sorbet-runtime-0.5.10151/lib/types/private/methods/call_validation.rb:161:in `validate_call'
        from /Users/{user}/.gem/ruby/3.1.2/gems/sorbet-runtime-0.5.10151/lib/types/private/methods/_methods.rb:270:in `block in _on_method_added'
        from /Users/{user}/.gem/ruby/3.1.2/gems/packwerk-2.2.0/lib/packwerk/package_set.rb:24:in `load_all_from'
        from /Users/{user}/.gem/ruby/3.1.2/gems/sorbet-runtime-0.5.10151/lib/types/private/methods/call_validation.rb:161:in `bind_call'
        from /Users/{user}/.gem/ruby/3.1.2/gems/sorbet-runtime-0.5.10151/lib/types/private/methods/call_validation.rb:161:in `validate_call'
        from /Users/{user}/.gem/ruby/3.1.2/gems/sorbet-runtime-0.5.10151/lib/types/private/methods/_methods.rb:270:in `block in _on_method_added'
[...]

PR #210, opened 2022-07-12
update explanation in USAGE for packwerk validate

dmitri-minkin:
`packwerk validate` command fails, instead we should use `bin/packwerk validate`
In the "Usage" explanation we mention `packwerk validate` command that will fail, see the [issue](https://github.com/Shopify/packwerk/issues/209) which is closed but not fixed.

changing the explanation is the easiest way to fix

wording

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -204,17 +204,17 @@ It will be a dependency violation when `components/shop_identity` tries to refer
 
 After enforcing the boundary checks for a package, you may execute:
 
-    packwerk check
+    bin/packwerk check
 
 Packwerk will check the entire codebase for any new or stale violations.
 
 You can also specify folders for a shorter run time. When checking against folders all subfolders will be analyzed, irrespective of nested package boundaries.
 
-    packwerk check components/your_package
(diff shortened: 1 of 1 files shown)


gmcgibbon (commented):
There are other references of `packwerk` that should be the binstub in the usage doc. Can you update those too?

gmcgibbon:
> Do you want to change it anyway ?

Yes. The docs should be consistently using the binstub as should anyone's usage of packwerk.

dmitri-minkin:
in the [issue](https://github.com/Shopify/packwerk/issues/209) I also provide the version:

Packwerk: 2.2.0

rafaelfranca:
Ok. Got it. That issue is already fixed in the version of the gem in the main branch. It isn't released yet, that is why you are still getting the error, you are not using the version with the fix.

So while using binstubs is a good practice, this PR isn't about fixing that issue.

rafaelfranca (approved):

PR #217, opened 2022-08-08
Try improving RESOLVING_VIOLATIONS.md

exterm:
While I agree that the PDF format wasn't ideal, and [mermaid is not a good fit](https://github.com/exterm/packwerk/issues/1) for the complexity of this flow, I think the first markdown version of this is too far from the user experience of the [original diagram](https://github.com/Shopify/packwerk/blob/fce44ae46adaa97cc097a647bf7cace112b8b12d/docs/new_violation_flow_chart.pdf). This is my attempt at getting a little closer to that.

My approach was to flatten the structure and make it more explicit, as well as add a sprinkle of emoji.

Does this improve the document from a user's perspective?

## Type of Change

- [ ] Bugfix
- [ ] New feature

## Checklist

- ~[ ] I have updated the documentation accordingly.~
- ~[ ] I have added tests to cover my changes.~

### DIFF
--- a/RESOLVING_VIOLATIONS.md
+++ b/RESOLVING_VIOLATIONS.md
@@ -1,50 +1,81 @@
-# Resolving violations
-## Understanding how to respond to new violations
+# Resolving Violations
 
-When you have a new dependency or privacy violation, what do you do?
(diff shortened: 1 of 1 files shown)


exterm:
Oh cool that comment triggered a re-run 😄

PR #220, opened 2022-08-11
Simplify packwerk path determination mechanics

shageman:
While working on https://github.com/Shopify/packwerk/pull/218 I noticed that there is a bit of cruft in the methods that are working together to determine the load paths to be included by packwerk. They pass around hashes of loaders, when all that is needed is pathname (sometimes as `Pathname`, sometimes as `String`).

I switched the params from using hashes to using arrays to clarify the intent of what is going on.

Nothing specific

## Type of Change

- [ ] Bugfix
- [ ] New feature

## Checklist

### DIFF
--- a/lib/packwerk/application_load_paths.rb
+++ b/lib/packwerk/application_load_paths.rb
@@ -9,7 +9,7 @@ module Packwerk
     class << self
       extend T::Sig
 
-      sig { params(root: String, environment: String).returns(T::Hash[String, Module]) }
+      sig { params(root: String, environment: String).returns(T::Array[String]) }
       def extract_relevant_paths(root, environment)
         require_application(root, environment)
         all_paths = extract_application_autoload_paths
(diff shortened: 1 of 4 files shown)


rafaelfranca (approved):
Love the simplification. Thank you

It seems the email you used to commit isn't attached to your GitHub account, so our CLI bot can't find if you signed it or not. Do you mind to fix that?

shageman:
I would love to. However, I had to change the email address associated with my account. I suspect Shopify needs to change the primary key behind the CLA form from `github_username` to `github_username+email`, so I can fill the CLA out again (which I don't mind doing)... or something like that

shageman:
Ah, ok. I will take a look.

shageman:
@rafaelfranca thanks for the tip! That worked!

mclark:
I saw this one too late, but the idea behind those hashes is to preserve Packwerk's knowledge of the "default namespace" for each root directory. GitHub relies on this feature for a number of odd older load dirs.

Can this be reverted?

rafaelfranca:
Of course. But I don't see that information being used by packwerk itself, is it?

mclark:
I'm about to board a plane at the moment but the original change is in https://github.com/Shopify/packwerk/pull/190 and has some context.

I'll investigate the full story later when I am back at a computer as it's been a while since I worked on that one. I suspect I probably messed up one of the tests in that PR.

mclark:
So just to close the loop on this one, https://github.com/Shopify/packwerk/pull/190/files#diff-8e43c83d1320f09335dd0d66cb5cc51646ccd11cabf6a3a696b1b4411b415ffdR41 is the key change. Basically, constant resolver [now supports](https://github.com/Shopify/constant_resolver#default-namespaces) a hash of load paths to default namespaces and Packwerk leverages that to support them for monoliths that use this behaviour.

I will see about adding a test to ensure this functionality is not lost again, it's disappointing nothing failed!

PR #222, opened 2022-08-11
Upgrade all gems and correct related code

rafaelfranca:
Only changes to code were related to new versions of rubocop-shopify and rubocop-sorbet.


Gemfile
```diff
@@ -21,4 +21,9 @@ group :development do
   gem("byebug", require: false)
   gem("minitest-focus", require: false)
   gem("ruby-lsp", "~> 0.2.1")
+
+  gem("m")
```
mclark, line 24: Is adding these dependencies via the gem file better practice now than using the gemspec dev dependencies, or is this being done to satisfy Sorbet?

mclark (commented):
I see you merged this already, but here's a review anyway. Looks good 👍

PR #227, opened 2022-08-17
Rename Node to NodeHelpers and add signatures

rafaelfranca:
While working in changing the parser to use SyntaxTree as an experiment, I noticed that the `AST::Node` abstraction was leaking everywhere. So I decided to refactor this code a little bit to make that work easier.

The first step was to rename the `Packwerk::Node` to `Packwerk::NodeHelpers`. This will free that name to hold a node abstraction to be used internally by packwerk and avoid the AST::Node implementation to leak everywhere.

Later I added method signature to those helpers so I can more easily refactor the code.


gmcgibbon (approved):

AlexEvanczuk:
Hey @rafaelfranca I'm curious how your experiment turns out. I did a basic proof of concept with SyntaxTree as an experiment and my experiment showed it actually did not produce an improvement in performance (it actually made it worse). There are some reasons why things may have changed:
- I didn't actually run syntax parser in its totality -- I just measured how long it took to parse a file to attempt to establish a lower bound for performance.
- My PR may have bugs in terms of the way I measured performance. I didn't put a ton of time into it... unfortunately did not share my results at the time, which I think would have been useful). I was able to dig it out of my local `git reflog` and pushed up the PR without changes https://github.com/Shopify/packwerk/pull/228.
- `SyntaxTree` may have merged performance improvements since then.

In general, very excited about the prospect of making `packwerk` faster. I've always been curious how a blazingly fast generalized ruby parser could unblock a whole ecosystem of ruby tools. Stripe is very understandably unable to support a generalized ruby parser extracted from Sorbet! However it is cool to think that fully-qualified constant resolution is a small subset of its entire process that includes a type check which takes <3 seconds for us. A constant only AST parser based on sorbet could reduce packwerk's runtime to a fraction while simultaneously removing (shifting) a lot of complexity from packwerk!

AlexEvanczuk (approved):

AlexEvanczuk:
Amazing, thanks for the additional context @rafaelfranca !

PR #233, opened 2022-10-17
Colocate privacy API on Packwerk::Package

alexevanczuk:
Much like https://github.com/Shopify/packwerk/pull/232, this moves us closer to #219 by colocating all concerns around privacy on `Packwerk::Package` by putting them in `PrivacyChecker`

I did not yet remove the old API on `Packwerk::Package`, since that would be a breaking API change, which I'd like to do all at once when we move `PrivacyChecker` to its own gem. When that happens, that gem can have a nice and easy to use API that can replace the old APIs folks were using on `Packwerk::Package`.

Let me know if this approach makes sense to you!

I'm wondering if I should just go ahead and remove the old API from `Packwerk::Package`. While it's probably best to assume folks *are* using it, I also wouldn't be surprised if very few or no clients are actually using them (`packwerk` itself is using them in very few places).

## Type of Change

- [ ] Bugfix
- [ ] New feature

## Checklist

### DIFF
--- a/lib/packwerk/reference_checking/checkers/privacy_checker.rb
+++ b/lib/packwerk/reference_checking/checkers/privacy_checker.rb
@@ -6,6 +6,9 @@ module Packwerk
     module Checkers
       # Checks whether a given reference references a private constant of another package.
       class PrivacyChecker
+        extend ActiveSupport::Autoload
+        autoload :PrivacyProtectedPackage
+
         extend T::Sig
         include Checker
 
--- a/lib/packwerk/package.rb
+++ b/lib/packwerk/package.rb
@@ -17,17 +17,15 @@ module Packwerk
     sig { returns(T::Array[String]) }
     attr_reader :dependencies
 
+    sig { returns(T::Hash[T.untyped, T.untyped]) }
+    attr_reader :config
+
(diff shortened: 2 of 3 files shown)


lib/packwerk/reference_checking/checkers/privacy_checker.rb
```diff
@@ -11,6 +11,39 @@ class PrivacyChecker
 
         VIOLATION_TYPE = T.let("privacy", String)
 
+        class << self
```
gmcgibbon, line 14: Does it make sense to put this in a nested different class or module? This logic isn't really related to the checker, it belongs to the package.

alexevanczuk: Eventually, this whole class will implement a `Packwerk::PluginInterface`, which will connect it to the rest of the packwerk runtime. I think that it might make sense to better factor out the privacy implementation into more logical parts when it comes time to do that, but I was hoping at this stage to just move all privacy related things into a pile in one place so that it can be deleted from packwerk.

Let me know if you think think that makes sense as a stepping stone.

gmcgibbon (approved):

PR #235, opened 2022-10-17
Use Zeitwerk for autoloading

alexevanczuk:
`packwerk` is based on the `zeitwerk` autoloader – figured we might as well use it!
We're already using `autoload`, which `zeitwerk` does for us automatically.

I followed the instructions in https://github.com/fxn/zeitwerk#synopsis

Any reason we don't want to do this?

## Type of Change

- [ ] Bugfix
- [ ] New feature

## Checklist

### DIFF
--- a/lib/packwerk.rb
+++ b/lib/packwerk.rb
@@ -8,81 +8,9 @@ require "fileutils"
 # Provides String#pluralize
 require "active_support/core_ext/string"
 
+require "zeitwerk"
+loader = Zeitwerk::Loader.for_gem
+loader.setup
+
 module Packwerk
-  extend ActiveSupport::Autoload
-
-  autoload :ApplicationLoadPaths
-  autoload :ApplicationValidator
-  autoload :AssociationInspector
-  autoload :OffenseCollection
-  autoload :Cache
-  autoload :Cli
-  autoload :Configuration
-  autoload :ConstantDiscovery
-  autoload :ConstantNameInspector
-  autoload :ConstNodeInspector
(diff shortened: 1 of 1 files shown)


lib/packwerk.rb
```diff
@@ -8,81 +8,9 @@
...
-  end
-
-  module Formatters
-    extend ActiveSupport::Autoload
-
-    autoload :OffensesFormatter
-    autoload :ProgressFormatter
-  end
+require "zeitwerk"
+loader = Zeitwerk::Loader.for_gem
```
gmcgibbon, line 12: Usually the Zeitwerk loader is exposed via a reader on the root module (eg. https://github.com/dry-rb/dry-monitor/blob/195c1324b7d3edfd00bd9eed8c61fb4ccc14af5f/lib/dry/monitor.rb#L23) but since this runs statically we probably don't need this unless we want to support plugin gems in the future.

gmcgibbon (commented):
This LGTM, but can you provide a benchmark proving that this doesn't impact performance?

lib/packwerk.rb
alexevanczuk: Oh interesting! Yeah, we can always make that change if that ever becomes a capability we wanted to support 👍

gmcgibbon:
If it consistently adds ~1 second to boot time, I don't see a clear reason to add Zeitwerk yet unless the complexity of the project calls for it. Since this is a static analysis tool, I'd rather optimize for speed.

PR #236, opened 2022-10-17
Run CI on push, which includes pull_request

alexevanczuk:
Reduce unnecessary test runs.

### Before
Screen Shot 2022-10-17 at 9 08 58 AM
### After
Screen Shot 2022-10-17 at 9 17 07 AM

I removed `pull_request` from `ci.yml`, as it appears `push` includes `pull_request` events.

Is there some event I'm missing here?

## Type of Change

- [ ] Bugfix
- [ ] New feature

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,6 @@
 name: CI
 
-on: [push, pull_request]
+on: [push]
 
 jobs:
   tests:


rafaelfranca:
Regular users can't push to branches on this repository, so would not their PRs no run any CI?

rafaelfranca:
Yes. That is the case https://github.com/Shopify/packwerk/pull/239. Reverting.

alexevanczuk:
Thank you @rafaelfranca !

PR #237, opened 2022-10-18
Prepare for 2.2.1 release

gmcgibbon:
### New Features

None.

### Bug Fixes

- Fix issue where incorrect violations are found when checking single file (#176)
- Remove digest dependency (#202)
- Filter custom paths with include patterns (#194)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    packwerk (2.2.0)
+    packwerk (2.2.1)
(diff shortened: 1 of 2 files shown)


gmcgibbon:
We decided to defer 3.0 since the breaking change is actually correct behaviour with a small blast radius. The branch name is confusing so I just pushed in https://github.com/Shopify/packwerk/commit/acd8e910805a05a398937b6cc77bbd63f5ab8fa5

PR #238, opened 2022-10-19
Fix stale violation detection

alexevanczuk:
Original issue: https://github.com/Shopify/packwerk/issues/154
Attempted fix: https://github.com/Shopify/packwerk/pull/176
Release with attempted fix: https://github.com/Shopify/packwerk/releases/tag/v2.2.1

I realized that the attempted fix had a bug after pulling into Gusto's codebase. I wrote two more tests (one which failed with the old implementation) and then fixed the failing test).

I decided to fix the issue inline and not do any further refactors or structural changes that would have made the issue more obvious. I think it would be nice to better factor that code but I'm not quite sure how to do that or if it makes sense yet.

Let me know if you'd prefer some more tidying of the code that had the bug to help make it easier to spot this sort of thing.

## Type of Change

- [x] Bugfix
- [ ] New feature

## Checklist

### DIFF
--- a/test/unit/parse_run_test.rb
+++ b/test/unit/parse_run_test.rb
@@ -230,6 +230,120 @@ module Packwerk
       assert result.status
     end
 
+    test "#check does not list stale violations when run on a single file with violations, even if the containing package has violations" do
+      use_template(:minimal)
+      file_to_check = "components/source/some/path.rb"
+      other_file = "components/source/some/other/path.rb"
+
+      source_package_name = "components/source"
+      write_app_file("#{source_package_name}/deprecated_references.yml", <<~YML.strip)
+        ---
+        "components/destination":
+          "::SomeName":
+            violations:
+            - privacy
+            files:
+            - #{other_file}
+            - #{file_to_check}
+          "::SomeOtherName":
+            violations:
+            - privacy
+            files:
+            - #{other_file}
+      YML
+
+      out = StringIO.new
+      parse_run = Packwerk::ParseRun.new(
+        relative_file_set: Set.new([file_to_check]),
+        configuration: Configuration.new({ "parallel" => false }),
+        progress_formatter: Packwerk::Formatters::ProgressFormatter.new(out)
+      )
+
+      offense = ReferenceOffense.new(
+        reference: build_reference(path: file_to_check),
+        message: "some message",
+        violation_type: Packwerk::ReferenceChecking::Checkers::PrivacyChecker::VIOLATION_TYPE
+      )
+
+      RunContext.any_instance.stubs(:process_file).returns([offense])
+      result = parse_run.check
+
+      expected_output = <<~EOS
+        📦 Packwerk is inspecting 1 file
+        .
+        📦 Finished in \\d+\\.\\d+ seconds
+      EOS
+
+      assert_match(/#{expected_output}/, out.string)
+
+      expected_message = <<~EOS
+        No offenses detected
+        No stale violations detected
+      EOS
+      assert_equal expected_message, result.message
+
+      assert result.status
+    end
+
+    test "#check does not list stale violations when run on a single file with no exising violations, but one new violation" do
+      use_template(:minimal)
+      file_to_check = "components/source/some/path.rb"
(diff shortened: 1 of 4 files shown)


lib/packwerk/deprecated_references.rb
```diff
@@ -64,11 +63,17 @@ def stale_violations?(for_files)
           }
         end
 
-        return true if package_violations_for_files.empty?
+        # We `next false` because if we cannot find existing violations for `for_files` within
+        # the `deprecated_references.yml` file, then there are no violations that
+        # can be considered stale.
+        next false if package_violations_for_files.empty?
```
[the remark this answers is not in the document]
alexevanczuk: My new test adds a violation, which allows this code to be hit and the test failure to emerge.

lib/packwerk/deprecated_references.rb
```diff
@@ -64,11 +63,17 @@ def stale_violations?(for_files)
...
 
         package_violations_for_files.any? do |constant_name, entries_for_constant|
           new_entries_violation_types = @new_entries.dig(package, constant_name, "violations")
-          return true if new_entries_violation_types.nil?
+          # If there are no NEW entries that match the old entries `for_files`,
+          # @new_entries is from the list of violations we get when we check this file.
+          # If this list is empty, we also must have stale violations.
+          next true if new_entries_violation_types.nil?
```
alexevanczuk, line 76: Technically this was behaviorally right before, but I think it's better practice to `next true` within an `any?` loop rather than `return true`, which breaks us out of the function entirely. I think this lends itself better to refactorable code.

test/unit/parse_run_test.rb
```diff
@@ -230,6 +230,120 @@ class ParseRunTest < Minitest::Test
...
+      expected_message = <<~EOS
+        No offenses detected
+        No stale violations detected
+      EOS
+      assert_equal expected_message, result.message
+
+      assert result.status
+    end
+
+    test "abc #check does not list stale violations when run on a single file with no exising violations, but one new violation" do
```
gmcgibbon, line 288:
```suggestion
    test "#check does not list stale violations when run on a single file with no existing violations, but one new violation" do
```

test/unit/parse_run_test.rb
```diff
@@ -230,6 +230,120 @@ class ParseRunTest < Minitest::Test
       assert result.status
     end
 
+    test "#check does not list stale violations when run on a single file with SOME violations, even if the containing package has violations" do
```
gmcgibbon, line 233:
```suggestion
    test "#check does not list stale violations when run on a single file with multiple violations, even if the containing package has violations" do
```
Or do you mean a violation from a different package? I'm not sure I'm understanding the test correctly.

gmcgibbon (commented):
This doesn't seem to cause behaviour differences for us. Can you please verify this fixes the bug on your end before we merge this? Also let's keep version bumps and bug fixes in separate commits.

test/unit/parse_run_test.rb
alexevanczuk: Basically the file that we run on has violations in this test, versus in the other test the file we passed in didn't have any violations. I can just remove `SOME` entirely I think.

alexevanczuk:
> This doesn't seem to cause behaviour differences for us. Can you please verify this fixes the bug on your end before we merge this? Also let's keep version bumps and bug fixes in separate commits.

I think it would cause a behavioral difference if you ran `bin/packwerk check path/to/file.rb` on a `file.rb` that has at least one violation in a package that has other violations other than that file.

When we do that on `main`, I get:
```
[TEST PROF INFO] Spring detected
Running via Spring preloader in process 53301
📦 Packwerk is inspecting 1 file
.
📦 Finished in 0.71 seconds

No offenses detected
There were stale violations found, please run `packwerk update-deprecations`
```

On this branch, I get:
```
Running via Spring preloader in process 53483
📦 Packwerk is inspecting 1 file
.
📦 Finished in 0.74 seconds

No offenses detected
No stale violations detected
```

So it looks like it's working!

I also removed the version bumps 👍

gmcgibbon (approved):

PR #242, opened 2022-10-25
[#221] Rename deprecated_references.yml to package_todo.yml

alexevanczuk:
This addresses https://github.com/Shopify/packwerk/discussions/221
More detail in the thread about the "why" for this change.
Also see the thread for the "when" on this change (i.e. when would we want to merge it, since it breaks public API).

- As agreed on in the thread, I chose `package_todo.yml` as the name for these new files.
- I added a new `bin/packwerk todo` command. I also considered the subcommands `update`, `update-todo`, but as this command is run pretty often, I wanted it to be really easy to type, and I didn't think the addition of "update" added much clarity anyways.
- I double-read the old and the new file. Initially, I wanted to release this in three steps (as listed in the discussion). However, there is too much output that refers to `package_todo.yml` that I thought it would be confusing to users to not see the warning but to see references to this new file they didn't know about. Therefore I thought it was better UX to do this in two steps: change and warn and then remove the old thing.

Let me know if this sequencing makes sense.

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

To transition to this version, it's just a matter of renaming `deprecated_references.yml` files in your project.
```
Dir['**/deprecated_references.yml'].each{|f| File.rename(f, f.gsub('deprecated_references', 'package_todo')) }
```

## Checklist


RESOLVING_VIOLATIONS.md
```diff
@@ -3,27 +3,27 @@
 Violations can be [recorded as a deprecation](#recording-violations) or (better!) [eliminated](#eliminating-violations).
 
 ## Recording Violations
-💡 New privacy and dependency violations are never hard-blocked. There are many very valid reasons to run `bin/packwerk update-deprecations`, adding new violations to `deprecated_references.yml` files. Even if you feel your reason might not be "valid," if your judgement says adding the violation and shipping your change will produce positive impact, trust your gut.
+💡 New privacy and dependency violations are never hard-blocked. There are many very valid reasons to run `bin/packwerk todo`, adding new violations to `package_todo.yml` files. Even if you feel your reason might not be "valid," if your judgement says adding the violation and shipping your change will produce positive impact, trust your gut.
 
 ### Emergency Fixes
 ❔ Is it a revert or is there a lot of urgency because you are fixing a production bug impacting customers?
 
-➡️ Simply run `bin/packwerk update-deprecations`, and address the violation when the customer issue is resolved.
```
alexevanczuk, line 11: Before there were 19 characters to type, now there are 4!

USAGE.md
```diff
@@ -235,28 +235,28 @@ See: [RESOLVING_VIOLATIONS.md](RESOLVING_VIOLATIONS.md)
 
 For existing codebases, packages are likely to have existing boundary violations.
 
-If so, you will want to stop the bleeding and prevent more violations from occuring. The existing violations in the codebase can be recorded in a [deprecated references list](#Understanding_the_list_of_deprecated_references) by executing:
+If so, you will want to stop the bleeding and prevent more violations from occuring. The existing violations in the codebase can be recorded in a [todo list](#understanding-the-package-todo-file) by executing:
```
alexevanczuk, line 238: I love "todo list" rather than "deprecated references list"

lib/packwerk/cli.rb
```diff
@@ -67,7 +69,8 @@ def execute_command(args)
           Subcommands:
             init - set up packwerk
             check - run all checks
-            update-deprecations - update deprecated references
+            todo - update package_todo.yml files
```
gmcgibbon, line 72: Personally, I think `update-todo` with a shorthand of `update` would make the most sense. From rubocop's perspective, it is `regenerate-todo`. Since `todo` by itself isn't a verb, I would suggest keeping the update wording in the command.

lib/packwerk/offense_collection.rb
```diff
@@ -71,17 +71,28 @@ def outstanding_offenses
...
     end
 
     sig { params(package: Packwerk::Package).returns(String) }
-    def deprecated_references_file_for(package)
-      File.join(@root_path, package.name, "deprecated_references.yml")
+    def package_todo_file_for(package)
+      if Pathname.new(@root_path).join(package.name, "deprecated_references.yml").exist?
+        warning = <<~WARNING
+          DEPRECATION WARNING: `deprecated_references.yml` files have been renamed to
+          `package_todo.yml`. Please see https://github.com/Shopify/packwerk/releases
```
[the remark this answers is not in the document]
gmcgibbon: - Is it possible to rename them automatically when we run check or update?
- You'll want a require for `Pathname` at the top of this file since we're using it now.
- Since we depend on Active Support, we should probably leverage it for deprecation warnings.

lib/packwerk/cli.rb
alexevanczuk: Ah yeah I see your point. I really loved the terseness of `todo` so thought maybe it was worth the tradeoff (of it not being a verb) since it's used so often, but `update` is only two more letters I suppose!! I'll change to `update` and `update-todo`.

alexevanczuk:
@gmcgibbon Responded to all your feedback and pushed up changes!

lib/packwerk/offense_collection.rb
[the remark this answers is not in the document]
gmcgibbon: > It is possible, but I'd prefer not to do this. I think this would be really surprising for users.

If you log a message when it happens explaining the change I don't feel like it would be confusing. We wouldn't have to code an extra command or point to release notes if it _just worked_.

> This is the same API we use in lib/packwerk/sanity_checker.rb and lib/packwerk/configuration.rb for warning about deprecated API, so I wanted to be consistent with that.

Noted. We should be consistent 👍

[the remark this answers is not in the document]
alexevanczuk: Okay! I'm down to play around with it in a separate PR. We can feel out what makes most sense there.

RESOLVING_VIOLATIONS.md
```diff
@@ -3,27 +3,27 @@
...
 
 ## Recording Violations
-💡 New privacy and dependency violations are never hard-blocked. There are many very valid reasons to run `bin/packwerk update-deprecations`, adding new violations to `deprecated_references.yml` files. Even if you feel your reason might not be "valid," if your judgement says adding the violation and shipping your change will produce positive impact, trust your gut.
+💡 New privacy and dependency violations are never hard-blocked. There are many very valid reasons to run `bin/packwerk update`, adding new violations to `package_todo.yml` files. Even if you feel your reason might not be "valid," if your judgement says adding the violation and shipping your change will produce positive impact, trust your gut.
 
 ### Emergency Fixes
 ❔ Is it a revert or is there a lot of urgency because you are fixing a production bug impacting customers?
 
-➡️ Simply run `bin/packwerk update-deprecations`, and address the violation when the customer issue is resolved.
+➡️ Simply run `bin/packwerk update`, and address the violation when the customer issue is resolved.
```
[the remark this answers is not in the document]
alexevanczuk: Okay sure thing

lib/packwerk/cli.rb
```diff
@@ -67,7 +71,9 @@ def execute_command(args)
           Subcommands:
             init - set up packwerk
             check - run all checks
-            update-deprecations - update deprecated references
+            update-todo - update package_todo.yml files
+            update - Shorthand for update-todo
+            update-deprecations - update package_todo.yml files (deprecated, use todo instead)
```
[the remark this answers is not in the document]
alexevanczuk: I'm not totally sure what those are used but it might be worth auditing at some point 🤔 I think the former is used to basically find old violations that should be removed. Note sure about `generate_configs` without digging in further.

lib/packwerk/cli.rb
mclark: Naming is hard! 😅  Normally I'd leave a conversation like this untouched as there's rarely value in muddying the waters, but this is the name of a subcommand and therefore pretty important so I thought I'd share my thoughts.

I also appreciate plain `todo` as it is brief and the intentions are clear. Grammatical consistency between subcommands is nice to have but not a hill I'd die on. `update-todo` feels a little unnecessarily wordy to me.

@gmcgibbon thanks for bringing up Rubocop's prior art! I also appreciate consistency and as Packwerk is very similar to Rubocop from the beginner's perspective, copying it is definitely a good thought.

Finally, another option that could work is dropping the `todo` command entirely and just having `packwerk check --todo`. Essentially updating the todo files is just a side effect or mode of a check anyway... 🤔 

My preferences, in order would be:

1. `packwerk check --todo` 
2. `packwerk check --regenerate-todo`
3. `packwerk todo`
4. `packwerk regenerate-todo`
5. `packwerk update-todo`

`packwerk update` feels too vague to me so I think I'd prefer to not support it at all honestly.

mclark (commented):
Thank you both for putting up with my nitpicking opinions! Thank you Alex for shepherding this change, I think everyone is really happy with this improved naming! ✨ 

I added 👀 to a few places that had inconsistent command names, just to make sure we came back to them. I left a couple in there for fun as it seemed redundant to spam this PR with the emoji.

lib/packwerk/cli.rb
```diff
@@ -57,7 +57,11 @@ def execute_command(args)
       when "detect-stale-violations"
         output_result(parse_run(args).detect_stale_violations)
       when "update-deprecations"
-        output_result(parse_run(args).update_deprecations)
+        output_result(parse_run(args).update_todo)
+      when "update-todo"
+        output_result(parse_run(args).update_todo)
+      when "todo"
```
gmcgibbon, line 63: Todo isn't a good shorthand because it isn't a verb. It should be `update`.

lib/packwerk/cli.rb
```diff
@@ -57,7 +57,11 @@ def execute_command(args)
       when "detect-stale-violations"
         output_result(parse_run(args).detect_stale_violations)
       when "update-deprecations"
-        output_result(parse_run(args).update_deprecations)
+        output_result(parse_run(args).update_todo)
```
gmcgibbon, line 60: Need to raise a deprecation warning here.

gmcgibbon (commented):
LGTM after these two things are done.

lib/packwerk/cli.rb
alexevanczuk: Alrighty I changed it back. I felt like Matt C. had some convincing points here that I think are worth responding to, but would love to be able to move forward on this too.

lib/packwerk/cli.rb
alexevanczuk: Added a warning.

alexevanczuk:
Changed those two things @gmcgibbon !

lib/packwerk/cli.rb
gmcgibbon: Personally `todo` doesn't make sense to me because it ins't a verb and may be misleading due to the mutative nature of the command. `todo` also feels wrong because the root of the command is "updating" the todo, so `update` makes sense to me as a partial completion of the full name and the verb of what you're actually doing.

I'm a fan of `packwerk check --update-todo` or `--regenerate-todo` though. Since there's no to type, we can settle on `update` for now and change it later if we find it necessary. If this is all about typing less characters we can always introduce `packwerk u`.

lib/packwerk/cli.rb
```diff
@@ -57,7 +57,17 @@ def execute_command(args)
       when "detect-stale-violations"
         output_result(parse_run(args).detect_stale_violations)
       when "update-deprecations"
-        output_result(parse_run(args).update_deprecations)
+        warning = <<~WARNING
```
gmcgibbon, line 60:
```suggestion
        warning = <<~WARNING.squish
```

lib/packwerk/cli.rb
```diff
@@ -57,7 +57,17 @@ def execute_command(args)
...
+          DEPRECATION WARNING: `update-deprecations` is deprecated in favor of
+          `update-todo` or just `update`
```
gmcgibbon, line 62:
```suggestion
          `update-todo`.
```

lib/packwerk/offense_collection.rb
```diff
@@ -71,17 +73,28 @@ def outstanding_offenses
...
+        package_todo_file_for(package),
       )
     end
 
     sig { params(package: Packwerk::Package).returns(String) }
-    def deprecated_references_file_for(package)
-      File.join(@root_path, package.name, "deprecated_references.yml")
+    def package_todo_file_for(package)
+      if Pathname.new(@root_path).join(package.name, "deprecated_references.yml").exist?
+        warning = <<~WARNING
```
gmcgibbon, line 87:
```suggestion
        warning = <<~WARNING.squish
```
Should this be one line?

gmcgibbon (approved):

lib/packwerk/cli.rb
alexevanczuk: Thanks @gmcgibbon and @mclark for both sharing more thoughts. Although I merged this, if @gmcgibbon agrees with being consistent with `tapioca`, I could put up a PR to switch over, given we haven't released anything yet (only merged this PR). Otherwise, we can leave as is.

PR #247, opened 2022-10-31
Pull privacy concerns out of packwerk

alexevanczuk:
# Do not merge yet
I think sequencing should look something like this:
- Release 2.3.0, with a note about deprecating `deprecated_references.yml` and `bin/packwerk update-deprecations` in favor of `package_todo.yml` and `bin/packwerk update-todo`, and a migration hint:
```ruby
Dir['**/deprecated_references.yml'].each{|f| File.rename(f, f.gsub('deprecated_references', 'package_todo')) }
```
- Land this
- Land https://github.com/Shopify/packwerk/pull/248
- Release 3.0.0

This PR removes privacy concerns out of the `packwerk` repository. It moves them to `packwerk-privacy` in this PR: https://github.com/rubyatscale/packwerk-privacy/pull/1

I completely removed all of the privacy related code along with tests that should be moved to the other repo.

Is there something that I didn't remove that I should, or something I shouldn't have removed that I did?

## Type of Change

- [ ] Bugfix
- [x] New feature: Extendible checkers and validators!

### Additional Release Notes

Users will need to include `packwerk-privacy` in their `Gemfile` and add `packwerk-privacy` to the `require` directive in `packwerk.yml`. USAGE.md includes more detail for how to do this.

This PR will require a major version bump, and the upgrade guide should go into a bit more detail on this.
We'll also want to couple a major version bump with the removal of references to `deprecated_references.yml`.

## Checklist


test/fixtures/package_todo.yml
```diff
@@ -2,7 +2,7 @@
 buyers:
   "::Buyers::Document":
     violations:
-    - dependency
+    - some_checker_type
```
alexevanczuk, line 5: This was originally used to test when a dependency violation turned into a privacy violation. I changed it to privacy originally and then realized we could use any arbitrary string to represent a different checker's violation.

alexevanczuk:
@rafaelfranca @gmcgibbon I think this is in a good state for review now!

lib/packwerk/validator_interface.rb
```diff
@@ -0,0 +1,44 @@
...
+
+module Packwerk
+  module ValidatorInterface
+    extend ActiveSupport::Autoload
+    autoload :Helpers
+
+    extend T::Sig
+    extend T::Helpers
+
+    include Helpers
```
gmcgibbon, line 16: This will load helpers immediately, so there is no need to autoload it. Also curious if we should leave it up to the concrete implementation to include this if they need it?

lib/packwerk/cli.rb
```diff
@@ -154,11 +154,15 @@ def validate(_paths)
...
-      Packwerk::ApplicationValidator.new(
-        config_file_path: @configuration.config_path,
-        configuration: @configuration,
-        environment: @environment,
+    def validator
+      Packwerk::ApplicationValidator.new
+    end
+
+    sig { returns(PackageSet) }
+    def package_set
```
gmcgibbon, line 162: I had to define the package set on the run_context in parse_run to cleanup todos. I wonder if we DRY things up somehow.

USAGE.md
```diff
@@ -290,3 +247,11 @@ require:
 `packwerk` will directly call `require` with these paths.
 You can prefix local files with a dot to define them relative to `packwerk.yml`, or you can use absolute paths.
 You can also reference the name of a gem.
+
+## Examples
+### Privacy Checker
+[`packwerk-extensions`](https://github.com/rubyatscale/packwerk-extensions) (originally extracted from `packwerk`) can be used to help define and enforce public API boundaries of a package. See the README.md for more details. To use this, add it to your `Gemfile` and then require it via `packwerk.yml`:
```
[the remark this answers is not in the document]
alexevanczuk: This is something I discussed with @mclark . Originally we called the gem `packwerk-privacy` but he made a good point that (A) clients would likely not want to install a bunch of different gems to be able to use extra checkers and (B) from a maintenance perspective it's simpler to maintain a repo of several checkers than have one repo for each checker. The analogy he used was rubocop, or different rubocop enhancement gems which ship with several cops.

lib/packwerk/validator_interface.rb
alexevanczuk: When you say there is no need to autoload it -- do you mean replace with a vanilla require statement (since something needs to require it somehow)? I made that change.

alexevanczuk: I'm inclined to keep it because the fact that it's in a separate module is mostly just an implementation detail -- in fact I wonder if it would be better to just inline it into `AbstractValidator`. I think a lot of validators will need these methods (as the existing ones did) and wanted to avoid an extra step for the clients to be able to implement they validation.

That being said I'm open to changing this so the concrete implementation has to include it if you feel strongly about it!

USAGE.md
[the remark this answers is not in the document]
gmcgibbon: Understood.

> clients would likely not want to install a bunch of different gems to be able to use extra checkers

I know what you mean, and borrowing conventions from rubocop, extension gems often ship with a collection of cops instead of just one. I think what I'm concerned about more is the name `packwerk-extensions` because it sounds too all-encompassing. I would assume there will be extensions built in the future that aren't in this package.

>  from a maintenance perspective it's simpler to maintain a repo of several checkers than have one repo for each checker.

Yes, but you can do that in a monorepo of separately released packages. If you only have one checker in mind there's nothing stopping you from making a gem that includes several other more specific gems later on (this is basically what Rails is). If you find a bug in one of the checkers, you can fix, bump, and release the specific checker and not be blocked by work being done in other checkers.

What I'm getting at is, I wouldn't expect `packwerk-extensions` to ship with just one checker, and I think the theming around the gem name should be thought of before you release it. I ultimately have no say in this though, so feel free to do what makes sense to you!

gmcgibbon (approved):

lib/packwerk/abstract_validator.rb
```diff
@@ -0,0 +1,81 @@
...
+        T.unsafe(@validators).map(&:new)
+      end
+    end
+
+    sig { abstract.returns(T::Array[String]) }
+    def permitted_keys
+    end
+
+    sig { abstract.params(package_set: PackageSet, configuration: Configuration).returns(ApplicationValidator::Result) }
+    def call(package_set, configuration)
```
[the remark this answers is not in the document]
alexevanczuk: Makes a lot of sense, I'll make this change next week before merging.

lib/packwerk/abstract_validator.rb
```diff
@@ -0,0 +1,81 @@
...
+    end
+
+    sig { params(configuration: Configuration, setting: T.untyped).returns(T.untyped) }
```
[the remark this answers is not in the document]
alexevanczuk: Good call, I think going with string is good, at least until we have a use case for anything else.

mriddle:
For those interested, the reasoning behind these changes are outlined in the follow blog post:
https://railsatscale.com/2024-01-26-a-packwerk-retrospective#packwerk-as-a-privacy-enforcer

> Unfortunately, while easy to use, Packwerk’s privacy checks introduced several problems. Some of these were problems of implementation: the checks required a separate app/public directory for code that was meant to be public API. This broke Rails conventions on file layout by introducing a folder under app that denoted privacy level instead of architecture concepts. Confusion around where files should go resulted in new subdirectories being created for controllers and jobs, duplicating those that already existed under app. As public API, these subdirectories should have been documented and well thought-out, but Packwerk didn’t encourage this level of detail. Thus, we ended up with endless poorly-documented public code that was never meant to be public in the first place.

> Yet there was a deeper problem, namely that privacy checks had transformed Packwerk into something it was never intended to be: an API design tool. Packwerk was being used to ensure that packages communicated via blessed entrypoints, whereas its original purpose was to define and enforce a dependency graph. Package A using package B’s code (even its public API) is not acceptable if package A doesn’t depend on B, yet we found developers were focusing on the design of their APIs over the dependencies in their code. This was drawing attention away from the problem the tool had been created to solve.

Given these issues, privacy checks were removed from Packwerk with the [release of version 3.0](https://github.com/Shopify/packwerk/releases/tag/v3.0.0).

PR #252, opened 2022-11-01
Drop EOL ruby

gmcgibbon:

Drop EOL rubies and test against a more recent version of Rails.

Edit CI YAML and gemfiles.

Should we test against Rails and Ruby edge?

## Type of Change

- [ ] Bugfix
- [ ] New feature

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,13 +12,9 @@ jobs:
           - gemfiles/Gemfile-rails-6-0
           - gemfiles/Gemfile-rails-6-1
         ruby:
-          - 2.6
           - 2.7
           - 3.0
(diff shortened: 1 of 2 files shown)


gmcgibbon:
> It's a bit of a tradeoff because (if I'm understanding it correctly) it could result in failures in unrelated PRs (i.e. failures can happen anytime once a new commit is merged), but it can always be turned off later if we changed our mind.

The edge checks could be optional checks that don't dictate if CI to be green. It would just be for visibility. I can add this in a separate PR, and maybe consider adding Zeitwerk head since we use private API (this would tip us off to changes earlier). I want CI to be fixed ASAP, so I'm going to merge this.

PR #258, opened 2022-11-07
Require Zeitwerk 2.6.1

gmcgibbon:

Followup on https://github.com/Shopify/packwerk/pull/251.

Require Zeitwerk 2.6.1 so that we can always use the public introspection API to get application autoload paths.

The loader approach previously used was good, but if we can just require a later version of Zeitwerk, we don't need to write branching logic for each version.

Is there any reason not to impose a minimum version of Zeitwerk for the gem?

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist


alexevanczuk (commented):
This seems fine but would we call this a breaking change? We're removing support for <2.6.4, so I thought this could be released in the major version.

rafaelfranca (approved):

PR #261, opened 2022-11-14
Allow Checker to define the location of its violations

alexevanczuk:
This allows a `Checker` implementation to define the location of its own violations. This is important for the extracted privacy checker, which we discuss [here](https://github.com/rubyatscale/packwerk-extensions/discussions/2) about changing violations so a package's `package_todo.yml` file shows *inbound* privacy violations rather than the default *outbound* violations.

I delegated this behavior to the `Checker` as a method with a default implementation (equal to today's implementation). This made sense to me since the idea is that a checker would decide this. However, we could also choose other implementations, such as having a checker specify in a method whether its violations are listed in the outbound package or the inbound, and then have the implementation of that be hidden.

Does this make sense as the API to specify location of violation?

## Type of Change

- [ ] Bugfix
- [x] New feature

## Checklist

- [x] I have updated the documentation accordingly => We probably want to have some general documentation on adding a new checker, although I think a user could always just look at the dependency checker
- [x] I have added tests to cover my changes => I believe everything is covered by the existing tests for the dependency checker

### DIFF
--- a/lib/packwerk/checker.rb
+++ b/lib/packwerk/checker.rb
@@ -6,10 +6,10 @@ module Packwerk
     extend T::Sig
     extend T::Helpers
 
-    class TodoLocation < T::Enum
+    class OffendingPackage < T::Enum
       enums do
-        PackageReferencingConstant = new
-        PackageDefiningConstant = new
+        ReferencePackage = new
+        DefinitionPackage = new
       end
     end
 
@@ -54,18 +54,18 @@ module Packwerk
     sig { abstract.params(reference: Reference).returns(String) }
     def message(reference); end
 
-    sig { returns(TodoLocation) }
+    sig { returns(OffendingPackage) }
     def todo_location
-      TodoLocation::PackageReferencingConstant
+      OffendingPackage::ReferencePackage
     end
 
     sig { params(reference: Reference).returns(Package) }
-    def todo_file_for(reference)
+    def offending_package_for(reference)
       location = todo_location
       case location
-      when TodoLocation::PackageDefiningConstant
+      when OffendingPackage::DefinitionPackage
         reference.constant.package
-      when TodoLocation::PackageReferencingConstant
+      when OffendingPackage::ReferencePackage
         reference.source_package
       else
         T.absurd(location)
--- a/lib/packwerk/checker.rb
+++ b/lib/packwerk/checker.rb
@@ -25,7 +25,16 @@ module Packwerk
 
       sig { params(violation_type: String).returns(Checker) }
       def find(violation_type)
-        T.must(Checker.all.find { |c| c.violation_type == violation_type })
+        checker_by_violation_type(violation_type)
+      end
+
+      private
+
+      sig { params(name: String).returns(Checker) }
+      def checker_by_violation_type(name)
+        @checker_by_violation_type ||= T.let(@checker_by_violation_type, T.nilable(T::Hash[String, T.nilable(Checker)]))
+        @checker_by_violation_type ||= Checker.all.map { |c| [c.violation_type, c] }.to_h
+        T.must(@checker_by_violation_type[name])
       end
     end
 
--- a/lib/packwerk/checker.rb
+++ b/lib/packwerk/checker.rb
@@ -55,21 +55,8 @@ module Packwerk
     def message(reference); end
 
     sig { returns(OffendingPackage) }
-    def todo_location
+    def offending_package
       OffendingPackage::ReferencePackage
-    end
-
-    sig { params(reference: Reference).returns(Package) }
-    def offending_package_for(reference)
-      location = todo_location
-      case location
-      when OffendingPackage::DefinitionPackage
-        reference.constant.package
-      when OffendingPackage::ReferencePackage
-        reference.source_package
-      else
-        T.absurd(location)
-      end
     end
   end
 end
--- a/lib/packwerk/checker.rb
+++ b/lib/packwerk/checker.rb
@@ -6,6 +6,13 @@ module Packwerk
     extend T::Sig
     extend T::Helpers
 
+    class TodoLocation < T::Enum
+      enums do
+        PackageReferencingConstant = new
+        PackageDefiningConstant = new
+      end
+    end
+
     abstract!
 
     class << self
@@ -47,9 +54,22 @@ module Packwerk
     sig { abstract.params(reference: Reference).returns(String) }
     def message(reference); end
 
+    sig { returns(TodoLocation) }
+    def todo_location
+      TodoLocation::PackageReferencingConstant
+    end
+
     sig { params(reference: Reference).returns(Package) }
     def todo_file_for(reference)
-      reference.source_package
+      location = todo_location
+      case location
+      when TodoLocation::PackageDefiningConstant
+        reference.constant.package
+      when TodoLocation::PackageReferencingConstant
+        reference.source_package
+      else
+        T.absurd(location)
+      end
     end
   end
 end
--- a/lib/packwerk/checker.rb
+++ b/lib/packwerk/checker.rb
@@ -27,6 +27,7 @@ module Packwerk
 
       sig { returns(T::Array[Checker]) }
       def all
+        require "packwerk/reference_checking/checkers/dependency_checker"
         T.unsafe(@checkers).map(&:new)
       end
 
--- a/lib/packwerk/checker.rb
+++ b/lib/packwerk/checker.rb
@@ -22,6 +22,11 @@ module Packwerk
       def all
         T.unsafe(@checkers).map(&:new)
       end
+
+      sig { params(violation_type: String).returns(Checker) }
+      def find(violation_type)
+        T.must(Checker.all.find { |c| c.violation_type == violation_type })
+      end
     end
 
     sig { abstract.returns(String) }
@@ -32,5 +37,10 @@ module Packwerk
 
     sig { abstract.params(reference: Reference).returns(String) }
     def message(reference); end
+
+    sig { params(reference: Reference).returns(Package) }
+    def todo_file_for(reference)
+      reference.source_package
+    end
   end
 end
--- a/lib/packwerk/offense_collection.rb
+++ b/lib/packwerk/offense_collection.rb
@@ -76,7 +76,7 @@ module Packwerk
     sig { params(offense: ReferenceOffense).returns(Packwerk::PackageTodo) }
     def package_todo_for_offense(offense)
       checker = Checker.find(offense.violation_type)
-      package_name_where_violation_should_live = checker.offending_package_for(offense.reference)
+      package_name_where_violation_should_live = offending_package_for(checker, offense.reference)
       package_todo_for(package_name_where_violation_should_live)
     end
 
@@ -89,6 +89,19 @@ module Packwerk
           package,
           package_todo_file_for(package),
         ).delete_if_exists
+      end
+    end
+
+    sig { params(checker: Checker, reference: Reference).returns(Package) }
+    def offending_package_for(checker, reference)
+      location = checker.offending_package
+      case location
+      when Checker::OffendingPackage::DefinitionPackage
+        reference.constant.package
+      when Checker::OffendingPackage::ReferencePackage
+        reference.source_package
+      else
+        T.absurd(location)
       end
     end
 
--- a/lib/packwerk/offense_collection.rb
+++ b/lib/packwerk/offense_collection.rb
@@ -76,7 +76,7 @@ module Packwerk
     sig { params(offense: ReferenceOffense).returns(Packwerk::PackageTodo) }
     def package_todo_for_offense(offense)
       checker = Checker.find(offense.violation_type)
-      package_name_where_violation_should_live = checker.todo_file_for(offense.reference)
+      package_name_where_violation_should_live = checker.offending_package_for(offense.reference)
       package_todo_for(package_name_where_violation_should_live)
     end
 
--- a/lib/packwerk/offense_collection.rb
+++ b/lib/packwerk/offense_collection.rb
@@ -35,9 +35,7 @@ module Packwerk
       return false unless offense.is_a?(ReferenceOffense)
 
       reference = offense.reference
-      checker = Checker.find(offense.violation_type)
-      package_name_where_violation_should_live = checker.todo_file_for(reference)
-      package_todo_for(package_name_where_violation_should_live).listed?(reference,
+      package_todo_for_offense(offense).listed?(reference,
         violation_type: offense.violation_type)
     end
 
@@ -49,10 +47,8 @@ module Packwerk
         @errors << offense
         return
       end
-      checker = Checker.find(offense.violation_type)
-      package_name_where_violation_should_live = checker.todo_file_for(offense.reference)
-      package_todo = package_todo_for(package_name_where_violation_should_live)
-      unless package_todo.add_entries(offense.reference, offense.violation_type)
+
+      unless package_todo_for_offense(offense).add_entries(offense.reference, offense.violation_type)
         new_violations << offense
       end
     end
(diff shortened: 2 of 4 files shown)


mclark:
@alexevanczuk instead of searching for the checker corresponding to a given violation in potentially tight loops, what do you think about allowing checkers to initialize themselves (the target todo for their violation type in this case) when they are added to `@checkers`? This way, Packwerk could know the target todo for each violation type and it would allow a simple hash look up when reporting or building offenses.

In this case, we might want a few class methods that each Checker implementation could provide. For starters we might just have something like this (but with an enum instead of a string probably):

```
sig { returns(String) }
def self.offense_receiver
  "source"
end
```

I'm not sure this is the best way to declare this sort of configuration information but I think it's probably best to avoid looping through the checkers. We could create a `CheckerConfiguration` property and class as well, but at this point that feels like it could be a bit premature.

The specific implementation I'm less settled on, but I think the idea of checker initialization is something worthwhile figuring out prior to this change to allow configurable target violation todo files.

alexevanczuk:
@mclark Would it change anything if instead of looping through the checkers, we build a memoized hash of the checkers that looks like `{ 'privacy' => privacy_checker, 'dependency' => dependency_checker}` that we can index into instead of looping over? I think the API to specify the direction (i.e. `offense_receiver` vs `todo_file_for`) feels separate from how do we get the checker given a violation type. That being said, open to a different API for sure.

lib/packwerk/offense_collection.rb
```diff
@@ -46,7 +49,9 @@ def add_offense(offense)
         @errors << offense
         return
       end
-      package_todo = package_todo_for(offense.reference.source_package)
+      checker = Checker.find(offense.violation_type)
+      package_name_where_violation_should_live = checker.todo_file_for(offense.reference)
+      package_todo = package_todo_for(package_name_where_violation_should_live)
```
mclark, line 54: Thoughts on a new method `package_todo_for_offense` as follows:

```ruby
sig { params(ReferenceOffense: offense).returns(Packwerk::PackageTodo) }
def package_todo_for_offense(offense)
  checker = Checker.find(offense.violation_type)
  package_name_where_violation_should_live = checker.todo_file_for(offense.reference)
  package_todo_for(package_name_where_violation_should_live)
end
```

lib/packwerk/checker.rb
```diff
@@ -32,5 +53,23 @@ def invalid_reference?(reference); end
...
+    sig { params(reference: Reference).returns(Package) }
+    def todo_file_for(reference)
+      location = todo_location
+      case location
+      when TodoLocation::PackageDefiningConstant
+        reference.constant.package
+      when TodoLocation::PackageReferencingConstant
+        reference.source_package
+      else
+        T.absurd(location)
```
mclark, line 71: I like this and I think it makes sense 👍

lib/packwerk/checker.rb
```diff
@@ -6,6 +6,13 @@ module Checker
     extend T::Sig
     extend T::Helpers
 
+    class TodoLocation < T::Enum
+      enums do
+        PackageReferencingConstant = new
+        PackageDefiningConstant = new
+      end
+    end
```
mclark, line 14: I like this idea but I feel like we might want to bikeshed on the names a bit. On a higher level this is indicating the package which has committed the offense, we don't even need to consider the "todo" file at all. It's basically answering the question of "who's violation is this?" Maybe something like this:

```ruby
class OffendingPackage < T::Enum
  enums do
    ReferencePackage = new
    DefinitionPackage = new
  end
end
```

lib/packwerk/checker.rb
```diff
@@ -32,5 +53,23 @@ def invalid_reference?(reference); end
...
     sig { abstract.params(reference: Reference).returns(String) }
     def message(reference); end
+
+    sig { returns(TodoLocation) }
+    def todo_location
+      TodoLocation::PackageReferencingConstant
+    end
+
+    sig { params(reference: Reference).returns(Package) }
+    def todo_file_for(reference)
```
mclark, line 63:
```suggestion
    def package_for(reference)
```

alexevanczuk: Incorporated this as `offending_package_for`

alexevanczuk:
> > Is the idea to have a sort of `Packwerk::Checker::Configuration`
> 
> It could be, but I wouldn't jump to that. Initially, it could be just a matter of making `todo_location` a class method and calling it when `Packwerk::Checker` is included by a subclass and registering the location for the checker's violation type at that time. Then the checker finding logic could be dropped entirely.
> 
> The only question is: where do we register it? That I don't know unfortunately. Initially the `Checker` class could have a method that takes a violation type and returns a "todo location" with data populated from this registration.
> 
> Regardless, this can all wait for a follow up, I wouldn't block on this discussion.

Gotcha, that sounds good. Since this should all be private implementation (well – technically `Checker.find` is public, but it's intended to be a packwerk internal method), we can always refactor it as we add more checker-specific API.

lib/packwerk/checker.rb
```diff
@@ -32,5 +53,23 @@ def invalid_reference?(reference); end
 
     sig { abstract.params(reference: Reference).returns(String) }
     def message(reference); end
+
+    sig { returns(OffendingPackage) }
+    def todo_location
```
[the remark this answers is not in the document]
alexevanczuk: Changed!

lib/packwerk/checker.rb
```diff
@@ -22,6 +29,20 @@ def included(base)
       def all
         T.unsafe(@checkers).map(&:new)
       end
+
+      sig { params(violation_type: String).returns(Checker) }
+      def find(violation_type)
```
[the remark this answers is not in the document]
alexevanczuk: Added! I figured this was mostly tested via the existing integration tests with the dependency checker, but it can't hurt to have a standalone test.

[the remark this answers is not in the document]
mclark: Agreed, even if it's only for documentation purposes it's worth having I think.

alexevanczuk:
Thanks @mclark 👍 

@gmcgibbon Let me know if you have any feedback on this PR! When/if we feel comfortable with this change, I'll squash the commits and merge.

rafaelfranca:
I'm not sure this is the right direction to go. The yml files could be see as a database and what you want with inverting the report is a different view in the database as the default one. Should we inversions of the violation being done by the checker when writing the data or by another layer by asking a different different query to the database?

I think things would be way more simple if packwerk could write the a SQLite database which you could ask any questions you like instead of having different YML formatter/paths per checker.

alexevanczuk:
Hey Rafael I’m not sure either what the long-term best strategy is for writing and reading violation data. At least in the interim, this feels like a huge net positive to support better alignment of responsibilities for privacy violations into a pack.

One really nice thing about “package_todo.yml” files, and being able to control which one stores a violation, is that this is todays de-facto API for end users to understand and respond to violations. It’s true that tooling can later replace this, but if we’re able to punt on managing the complexity of a database or more sophisticated tooling, that feels like a win.

What do you think?

alexevanczuk:
Hey Rafael responding to your comments here, but also open to chatting more over zoom/slack if you'd prefer that.

> I prefer to not release half-backed solutions. If the best long term solution isn't allowing checks to change location of todo files, so why would we release that? I don't think having different todo files in different places in the codebase based on which checks you use isn't good usability.

I don't think this is half-baked – we've put a **lot** of thought into this based on using privacy checking extensively at Gusto and have determined it will lead to users more easily understanding what the scope of the work is to building a public API. Also I'm not suggesting having todo files in different places, just having checkers be able to decide which (referencing or defining) pack should be responsible for the TODO.

> I'm also not convinced that the inversion of the privacy check is a good thing, since even if a team expose a public API, people can chose to use a private API or maybe they don't know there is a public API which would add a violation to the "provider" side, not to the "user" side and there isn't anything the provider should do in that case (they already provided the API).

This is also something we've thought about a lot. I know that Shopify didn't find the privacy checker very valuable, but at Gusto we found that the happy path is that a team wants to start building public API, so it will start enforcing privacy and use the TODO list (which to the point of this PR, would be better in one place, rather than distributed across the codebase), and then work their way to an API. Then, one everything is migrated, they can make private API hard-private with `private_constant`. We also like the idea of "strict mode" [discussed here](https://github.com/Shopify/packwerk/discussions/241) to support no new violations on a pack. There's a tradeoff here, but I think we're on the wrong side off it.

That being said, this is more about a general capability for checkers to say that either the defining pack or the referencing pack is responsible.

> But let's say it makes sense to invert that. Can't this be done already by reading all ymls, putting in a database and querying that database? Do we really need to store the data diferently?

I don't think that a database is a better solution here. I think there's huge value in being able to see a very simple YML output of violations that:
- Shows up in search
- Doesn't require complex dependencies (maintaining a database)
- Doesn't require additional tooling: right now the user can understand packwerk very intuitively and visually. Requiring a CLI tool for this feels like a big regression in user experience

This also feels slightly orthogonal to the API for storing the data. Even if we had a database to back this, we'd still need this PR to support an API such as `violations_for_pack(package)` that would query the database. This would be essential complexity that would have to live somewhere, and we'd need some way for the checker to indicate which pack is responsible for resolving a violation.

mclark:
> I prefer to not release half-backed solutions.

This comes across as unnecessarily harsh.

> I don't think having different todo files in different places in the codebase based on which checks you use isn't good usability.

Having a bit of trouble parsing the double negative, I'll assume it was unintentional. We currently have different todos at different places in the codebase so this is not a change.

> I'm also not convinced that the inversion of the privacy check is a good thing

That's fair. Maybe a configurable option on the checker could resolve this as we obviously disagree?

> But let's say it makes sense to invert that. Can't this be done already by reading all ymls, putting in a database and querying that database?

Good idea, maybe that is something we should do as a future feature? I don't see how a database is a prerequisite for this change.

---

Essentially, this PR is introducing the notion of a "responsible package" for each violation here. The referencing package and the definition package are still the same as always, it's just saying that the assumption that the referencing package is always the one responsible for the violation and tech debt is not always true. In the database you speak of, it might mean a new column to indicate this responsibility.

rafaelfranca:
Sorry, not my intention to be hash, but at the same time I felt that my concerns are not being taken in consideration in favor of some kind of rush to ship this as-is, which is what I'm trying to avoid. We should not need to rush any decision.

Not saying the solution of inverting the violation is half-backed, I'm sure you have thought a lot about this problem, and I can see how it makes sense in your context. I do want to support it. What I don't want is to support it in detriment to making the best decision for the open source project.

My response was to "I’m not sure either what the long-term best strategy is for writing and reading violation data. At least in the interim, this feels like a huge net positive to support better alignment of responsibilities for privacy violations into a pack.". He should have enough time to find what is the best long-term strategy here, without having to rush an interim solution that might put us in a corner later. It is possible that your proposal is the best long-term solution, so let's make sure of that.

> We currently have different todos at different places in the codebase so this is not a change.

What I mean is that today all todos are in the "users" package. With this change they could be in the "users" or in the "providers" (maybe in another place in the future) and in my opinion, that is confusing.

I think the ownership of the violation is always on the "users" side, but I can totally get that they can't do much with that information in terms of privacy.

Given control to the checker to were put the violation data to me can create confusion since you need to know what is expected from the checker. Those yml files will have different meaning if we go to that direction, with packwerk not being a simple tool anymore, but something so complicated that feels like we have many tools in the same library.

My counter proposal is to keep the yml files where they are, and in packwerk itself provide a way to output a sqlite3 database that allow teams to build different dashboards at their choice, inverting (or doing other things) to the violations. Things will still show up in search, and it will not require external tools to build the database (maybe to create the reports we would need external tools).

It also allow way more use cases than this. For example allowing us to "Give me all packages that are violating package 'X' privacy", or "Give me all packages that depend on 'X'" that today isn't possible at all.

alexevanczuk:
Thanks for sharing more Rafael. I appreciate the discussion and I don't want to rush anything either. I'm also of course happy to close this PR if this is not the approach we want to take.

> What I mean is that today all todos are in the "users" package. With this change they could be in the "users" or in the "providers" (maybe in another place in the future) and in my opinion, that is confusing.

I think this is part of the shift we want to make. What we've found is if the "users" package violates the "providers," package, right now that's a "providers" TODO that's in the "users" package. That is – dependency violations are TODOs of the *referencing*, but privacy violations are conceptually TODOs for the publishing/defining package. Packwerk today implicitly says "a TODO belongs to the package where the referencing file resides," but that doesn't line up with the way users think about and use the tool (hence why getting privacy right is so hard).

> provide a way to output a sqlite3 database that allow teams to build different dashboards at their choice

I agree there's a lot of value in a sqlite3 database too (and wonder if we can make the persistence layer an interface that can be implemented via `YmlPersistence` or `SqlitePersistence`). It certainly makes certain kinds of analysis simpler and opens up a lot of doors to interesting new tooling opportunities. I'd be interested in helping build that (or at least creating the ability to swap implementations for it), and it seems like would still want a DB driven solution to have a generic way that a checker can indicate which package is responsible for a violation. This PR creates both the interface to do this (as a checker implementation) and how the persistence layer (YML files) interact with it. It sounds like you're saying you want this to 100% be a client side filtering/organizing concept, but then we're back to square one where the generic tool wouldn't be capable of providing this nuance in how violations from generic checkers are reported per package.

baweaver:
I like the general idea of introducing a database here as many of my workflows involve reorienting Packwerk data to gain novel insights that may inform new approaches.

That said I have something which may be of interest, but will require a bit of lead in to understand my general philosophy here on violations.

### Consumer Orientation

As Packwerk exists today all violations are oriented to the consumers of private APIs. If a consumer needs a new API they have to reach out to a producing team, which may or may not respond, causing a lack of empowerment of the consumer. When implementing Packwerk chances are API coverage is near 0%, meaning paying down any violation as a consumer will be difficult to do or prioritize.

In the case where public APIs exist? This consumer orientation makes more sense as there's an immediate answer of what to use. Granted, getting that answer of what to use instead is its own unsolved problem around documentation or potential automation.

Consumer orientation is the most useful when public APIs already exist.

### Producer Orientation

When trying to consider how to go from 0 to reasonable API coverage we discovered that reorienting violations to the producing pack gave us novel insights into what APIs were most commonly used. Given our positions as owners of these packs we could create APIs which served our consumers, and would be empowered to do so given our domain knowledge of our own packs.

There are of course weaknesses here of over-designing and not listening to consumers here, but that's a side topic.

Producer orientation is the most useful when public APIs _do not_ exist yet.

### Get to the Point

The reason I bring this up is that as a pack evolves and gradually becomes "API complete" it will make sense for it to flip from "Producer Orientation" to "Consumer Orientation", switching the impetus of enforcement onto consumers to not violate what's been made.

Why is that important? Because I believe this should be a top-level configurable option depending on the state of your packs. Some will evolve public APIs much faster than others, and while they are doing so the impetus should be on them to create APIs rather than inflict that burden on anyone who needs their domain concerns.

------

I bring all of this up to say that the top level YAML files will still be valuable, even if we _do_ create the database around it, to drive conversations and roadmapping around modularization. The concept of "Producer Orientation" will be exceptionally powerful towards driving creation and adoption of public APIs, and making it top-level would be very useful for teams wanting to figure out where to start.

rafaelfranca:
I think @baweaver points only support more my view that changing the place were the database (YAML) is stored based on the checker makes no sense. If we do that, and the privacy checks use it, when a package is API complete and we want to swap the place where the YAML is stored we can't, or it will depend on how the package is configured. All the affordance of knowing where the TODO files lives will be lost.

I believe the YAML files should stay where they are, exactly to keep the conversations flowing, and reviews easier to be made (I don't want to review a binary file inside the repository to see if new violations were added) and the SQL based database is built using the YAML as source of true.

> but then we're back to square one where the generic tool wouldn't be capable of providing this nuance in how violations from generic checkers are reported per package.

That is true, but remember that the generic tooling only cares about dependencies checks, that I don't think makes much sense to have the perspective flipped. Having this in a SQL database can allow the privacy plugin to implement new commands like `packwerk report-missing-public-api my-package` that would read read the yaml files, generate a SQL database, query the database, and them create the report in the format you want. Having this intermediary SQL database can allow plugins to do many other things. Instead of making impossible to plugins to do what they want, I think it allow more flexibility to plugins.

alexevanczuk:
Thanks for sharing your perspective and all of the great discussion! Those points make sense, and the idea of keeping the YML alongside the binary file could definitely let us "have our cake and eat it too."

I'll close this for now! Looking forward to Packwerk 3.x!

PR #264, opened 2022-11-17
Allow extension system to load in custom OffensesFormatter

alexevanczuk:
Today, `packwerk` supports a configurable `OffensesFormatter`. However, the only way to configure it is to load the CLI itself. This means that this feature is not usable if a user has run `bundle binstubs packwerk`. Instead, the user needs to copy `exe/packwerk` into their local `bin/packwerk` and pass in a different formatter to `Packwerk::Cli.new`.

This PR makes this more user-friendly by leveraging the extension system to load in a new formatter. This way, a user can use the vanilla `packwerk` binstub and load in their own offenses formatter more easily.

I chose to use the `require` directive in `packwerk.yml` to load in a formatter via a setting on `Packwerk::Configuration`. The system first prioritizes the local input variable, then the configuration, then the default formatter.

Is there a more ideal API to set the custom formatter? Is there any reason we wouldn't want to encourage this via the extension system?

## Type of Change

- [ ] Bugfix
- [x] New feature

## Checklist

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -249,6 +249,15 @@ You can prefix local files with a dot to define them relative to `packwerk.yml`,
 You can also reference the name of a gem.
 
 ## Examples
+### Custom `OffensesFormatter`
+While `packwerk` ships with its own `OffensesFormatter`, the extension system provides a mechanism to configure your own to be used when `bin/packwerk check` is run.
+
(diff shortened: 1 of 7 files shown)


alexevanczuk:
I thought of an alternate implementation I think may be better. A --format option that matches on the name of a formatter. Then packwerk yml can specify the default. I'll put up a different PR later.

PR #265, opened 2022-11-17
Improve performance of bin/packwerk validate

alexevanczuk:
I did some performance analysis of `bin/packwerk validate` and found that the vast majority of the time was in checking acyclicity of the package graph. In Gusto's monolith, this PR represents an 80% performance improvement in `bin/packwerk validate`.

# Performance Benchmarking
I ran `time bin/packwerk validate` 5 times. Here are the results:

## BEFORE: 5x Runs Performance Benchmark on Gusto's monolith
```
real    0m30.920s
real    0m31.423s
real    0m31.106s
real    0m30.494s
real    0m31.340s
```

## AFTER: 5x Runs Performance Benchmark on Gusto's monolith
```
real    0m6.874s
real    0m6.716s
real    0m6.730s
real    0m6.737s
real    0m6.808s
```

I was originally looking into some other approaches, including:
1) Is it necessary to find *all* cycles? Or is it sufficient to find the first cycle that each node is in? I nixed this approach so as to not change behavior.
2) Is there a faster cycle-detection algorithm?

What it came down to is that we didn't need the full on `Packwerk::Package` – the `Graph` abstraction works on `nodes` that are just a `String`. I found that by passing in just the unique package name it massively reduced the computation required, creating a large performance improvement without a more involved rethink of cycle detection. 

- Is there any reason we *would not* want the `Graph` abstraction to take in a plain String and instead we would want a more sophisticated reprepresentation?
- Should we pursue further improvements?

## Type of Change

- [ ] Bugfix
- [ ] New feature

## Checklist

### DIFF
--- a/lib/packwerk/graph.rb
+++ b/lib/packwerk/graph.rb
@@ -4,8 +4,14 @@
 module Packwerk
   # A general implementation of a graph data structure with the ability to check for - and list - cycles.
   class Graph
-    # @param [Array<Array>] edges The edges of the graph; An edge being represented as an Array of two nodes.
-    def initialize(*edges)
+    extend T::Sig
+    sig do
+      params(
+        # The edges of the graph; An edge being represented as an Array of two nodes.
+        edges: T::Array[T::Array[T.any(String, Integer)]]
+      ).void
+    end
+    def initialize(edges)
       @edges = edges.uniq
       @cycles = Set.new
       process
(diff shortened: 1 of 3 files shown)


lib/packwerk/graph.rb
```diff
@@ -4,8 +4,14 @@
 module Packwerk
   # A general implementation of a graph data structure with the ability to check for - and list - cycles.
   class Graph
-    # @param [Array<Array>] edges The edges of the graph; An edge being represented as an Array of two nodes.
-    def initialize(*edges)
```
alexevanczuk, line 8: Splats are poorly supported by sorbet and I didn't see a great reason to keep this as a splat, but let me know if I'm missing something.

rafaelfranca (approved):
Great work!

PR #266, opened 2022-11-20
Proof of concept for configurable formatters

alexevanczuk:
- wip
- wip
- Use standard ApplicationFixtureHelper so we can more easily switch between templates
- Proof of concept for configurable formatters

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist


USAGE.md
```diff
@@ -249,6 +249,32 @@ You can prefix local files with a dot to define them relative to `packwerk.yml`,
...
+
+  def identifier
+    'my_offenses_formatter'
+  end
+end
+```
+
+Then in `packwerk.yml`, you can set the `formatter`:
+```yml
+formatter: my_offenses_formatter
```
alexevanczuk, line 270: Should this be `offenses_formatter`? Or just `format` to match the CLI flag?

lib/packwerk/cli.rb
```diff
@@ -5,6 +5,10 @@
 
 module Packwerk
   # A command-line interface to Packwerk.
+  # Would it be better to refactor this to use `thor`, which is already in the Gemfile.lock
+  # via railties and other gems?
+  # That would allow the `format` option to be simpler and would help us not need to reinvent the wheel
+  # submit a PR for this?
```
alexevanczuk, line 11: Will do this in a separate PR.

Note the `format` flag is not implemented in this PR.

lib/packwerk/formatters/offenses_formatter.rb
```diff
@@ -6,11 +6,13 @@ module Formatters
...
 
+      IDENTIFIER = T.let("default", String)
+
       extend T::Sig
 
-      sig { params(style: OutputStyle).void }
-      def initialize(style: OutputStyles::Plain.new)
-        @style = style
+      sig { returns(OutputStyle) }
+      def style
```
alexevanczuk, line 14: The `OffensesFormatter` interface calls `new` with these.

Using a YML file as the vehicle for configuration makes it hard to allow the client to specify their own `initialize` method.

Note that if we had a configuration option that looked like:
```ruby
Packwerk.configure do |config|
  config.formatters = [ MyFormatter.new(i_can_pass_in_whatever_here) ]
end
```

This would not be as much of a problem. However, since our implementation for loading in extensions is YML, we need to make the classes that implement the interfaces very simple. Therefore, I split this up into two classes: one that is the "default" that is colored, and one that right now is only used in tests which is the "plain" formatter. This way the class can be initialized and registered by the interface without an issue.

lib/packwerk/formatters/offenses_formatter_plain.rb
```diff
@@ -0,0 +1,23 @@
+# typed: strict
+# frozen_string_literal: true
+
+module Packwerk
+  module Formatters
+    class OffensesFormatterPlain < OffensesFormatter
```
alexevanczuk, line 6: This could also just be called `Plain` and the other one just called `Default`. Let me know what you think is best.

lib/packwerk/offenses_formatter.rb
```diff
@@ -6,7 +6,39 @@ module OffensesFormatter
...
+    abstract!
+
+    class << self
+      extend T::Sig
+
+      sig { params(base: Class).void }
+      def included(base)
+        @formatters ||= T.let(@formatters, T.nilable(T::Array[Class]))
+        @formatters ||= []
+        @formatter_by_identifier = nil
```
alexevanczuk, line 18: Note that we `nil` out the hash lookup when we add a new formatter. In practice, this should not be necessary, and we can remove if folks want. What's going on here is that a CLI was initialized in test, which included some formatters but not all, and then a later test added new formatters, but they were not properly registered due to the previously memoized lookup hash.

We could just change the tests to only load the CLI after we've loaded configuration, but this felt like an appropriate measure too. Open to changing!

lib/packwerk/formatters/offenses_formatter.rb
gmcgibbon: Using an abstract class instead where we control the initializer might help this situation. However, since people control the implementation of their formatter, we can assume they will not need to pass in anything special as they can memoize or inline it as needed.

lib/packwerk/formatters/offenses_formatter_plain.rb
alexevanczuk: Yep great idea, moved.

lib/packwerk/offenses_formatter.rb
```diff
@@ -6,7 +6,39 @@ module OffensesFormatter
...
+        formatter_by_identifier(identifier)
+      end
+
+      private
+
+      sig { params(name: String).returns(OffensesFormatter) }
+      def formatter_by_identifier(name)
+        @formatter_by_identifier ||= T.let(@formatter_by_identifier,
+          T.nilable(T::Hash[String, T.nilable(OffensesFormatter)]))
+        @formatter_by_identifier ||= OffensesFormatter.all.map { |c| [c.identifier, c] }.to_h
```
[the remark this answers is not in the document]
alexevanczuk: yep

alexevanczuk:
@gmcgibbon I think I've addressed/responded to all of your feedback.

lib/packwerk/formatters/offenses_formatter.rb
gmcgibbon: If we re-define the style in the formatter, does it make sense to pass the style into the CLI object anymore? Maybe we should remove that.

I'm also not 100% sure this should be public. It seems optional.

lib/packwerk/offenses_formatter.rb
```diff
@@ -1,12 +1,53 @@
...
+
+      sig { params(identifier: String).returns(OffensesFormatter) }
+      def find(identifier)
+        formatter_by_identifier(identifier)
+      end
+
+      private
+
+      sig { void }
+      def load_default_formatter
```
gmcgibbon, line 37: There isn't a need to referencer the constant if we require the offense formatter at the top of the file. If the formatter is only used when checking, maybe the best thing to do would be to do a runtime `require "packwerk/formatters/offenses_formatter"` in this function?

test/support/application_fixture_helper.rb
```diff
@@ -13,6 +13,11 @@ def setup_application_fixture
     @old_working_dir = Dir.pwd
   end
 
+  def remove_extensions
+    Object.send(:remove_const, :MyLocalExtension) if defined?(MyLocalExtension)
```
gmcgibbon, line 17:
```suggestion
    Object.send(:remove_const, :MyLocalExtension)
```
If we expect it to be there, it should be there. Otherwise our test is wrong for calling this method.

USAGE.md
alexevanczuk: Either the config nor CLI flag currently exist to my knowledge so shouldn't be breaking, so down to call it whatever!

USAGE.md
```diff
@@ -249,6 +249,27 @@ You can prefix local files with a dot to define them relative to `packwerk.yml`,
 You can also reference the name of a gem.
 
 ## Examples
+### Custom `OffensesFormatter`
```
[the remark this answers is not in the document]
alexevanczuk: Sure, will make that change

lib/packwerk/offenses_formatter.rb
alexevanczuk: Sure, I like that! Definitely clearer that we're explicitly loading rather than loading through invoking auto load.

test/support/application_fixture_helper.rb
alexevanczuk: Sure makes sense! Will change.

lib/packwerk/formatters/offenses_formatter.rb
alexevanczuk: Yeah, we still need style as a CLI arg for now for the progress formatter. However, you're right we don't need `style` as a public method of this formatter – made it private!

USAGE.md
gmcgibbon: Ah, wrote this with renaming the constant in mind, but if we go with the identifier approach you are correct 👍 

Let's go with `offenses_formatter` like you did with the CLI arg and we can change it before release if need be.

lib/packwerk/cli.rb
```diff
@@ -195,11 +196,20 @@ def parse_run(params)
         ignore_nested_packages = false
       end
 
+      if params.any? { |p| p.include?("--offenses_formatter") }
```
gmcgibbon, line 199:
```suggestion
      if params.any? { |p| p.include?("--offenses-formatter") }
```
Flags should alway be dasherized (or support both dashes and underscores)

test/test_helper.rb
```diff
@@ -19,6 +19,7 @@
 require "support/test_assertions"
 require "support/yaml_file"
 require "support/typed_mock"
+require "support/offenses_formatter_plain.rb"
```
gmcgibbon, line 22:
```suggestion
require "support/offenses_formatter_plain"
```

test/support/application_fixture_helper.rb
```diff
@@ -13,6 +13,18 @@ def setup_application_fixture
...
+  def remove_extensions
+    Object.send(:remove_const, :MyLocalExtension)
+    reset_formatters
+  end
+
+  def reset_formatters
+    Packwerk::OffensesFormatter.instance_variable_set(:@formatter_by_identifier, nil)
+    current_formatters = Packwerk::OffensesFormatter.instance_variable_get(:@offenses_formatters)
+    new_formatters = current_formatters.delete_if { |f| f.new.identifier == "my_offenses_formatter" }
+    Packwerk::OffensesFormatter.instance_variable_set(:@offenses_formatters, new_formatters)
```
gmcgibbon, line 25: We could just leave it at setting `nil`, right?

lib/packwerk/offenses_formatter.rb
```diff
@@ -15,5 +47,9 @@ def show_offenses(offenses)
     sig { abstract.params(offense_collection: Packwerk::OffenseCollection, for_files: T::Set[String]).returns(String) }
     def show_stale_violations(offense_collection, for_files)
     end
+
+    sig { abstract.returns(String) }
+    def identifier
+    end
```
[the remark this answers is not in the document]
gmcgibbon: Ok, it could be trying to build for the future too soon. Let's axe it for now and we can revisit it before release if need be.

USAGE.md
```diff
@@ -249,6 +250,34 @@ You can prefix local files with a dot to define them relative to `packwerk.yml`,
...
+```
+
+Then in `packwerk.yml`, you can set the `formatter` to the identifier for your class:
+```yml
+formatter: my_offenses_formatter
+```
+
+You can also pass in a formatter on the command line:
+```
+bin/packwerk check --format=my_offenses_formatter
```
gmcgibbon, line 278: Make sure this is using the correct flag before merging.

gmcgibbon (approved):
LGTM after some small changes.

PR #271, opened 2022-11-29
Allow strict mode enforcement of violations

alexevanczuk:
More info here: https://github.com/Shopify/packwerk/discussions/241
TLDR: This enables `strict` mode for checkers so that packages can disallow new violations, even if they're listed in `package_todo.yml` files.

I chose to delegate the question to the checker as it's closely coupled to *where* a violation is being stored. That is, if package A has a new dependency violation on package B, we need to check package A `enforce_dependencies` to determine if the violation when listed violates strict mode. If package A has a new *privacy* violation on package B, we need to check the `enforce_privacy` property for package B. Since this is determined by the checker, it needs to be a checker interface method.

Let me know if there is a better way to implement this, or a better UX (e.g. the output, the API to turn it on, etc.). Also let me know if I'm missing any edge cases!

## Type of Change

- [ ] Bugfix
- [x] New feature

## Checklist

### DIFF
--- a/lib/packwerk/checker.rb
+++ b/lib/packwerk/checker.rb
@@ -22,10 +22,28 @@ module Packwerk
       def all
         T.unsafe(@checkers).map(&:new)
       end
+
+      sig { params(violation_type: String).returns(Checker) }
+      def find(violation_type)
+        checker_by_violation_type(violation_type)
+      end
+
+      private
+
+      sig { params(name: String).returns(Checker) }
+      def checker_by_violation_type(name)
+        @checker_by_violation_type ||= T.let(Checker.all.to_h do |checker|
+                                               [checker.violation_type, checker]
+                                             end, T.nilable(T::Hash[String, Checker]))
+        @checker_by_violation_type.fetch(name)
+      end
     end
 
     sig { abstract.returns(String) }
     def violation_type; end
+
+    sig { abstract.params(listed_offense: ReferenceOffense).returns(T::Boolean) }
+    def strict_mode_violation?(listed_offense); end
 
     sig { abstract.params(reference: Reference).returns(T::Boolean) }
     def invalid_reference?(reference); end
--- a/lib/packwerk/offense_collection.rb
+++ b/lib/packwerk/offense_collection.rb
@@ -18,6 +18,7 @@ module Packwerk
       @root_path = root_path
       @package_todo = T.let(package_todo, T::Hash[Packwerk::Package, Packwerk::PackageTodo])
       @new_violations = T.let([], T::Array[Packwerk::ReferenceOffense])
+      @strict_mode_violations = T.let([], T::Array[Packwerk::ReferenceOffense])
       @errors = T.let([], T::Array[Packwerk::Offense])
     end
 
@@ -26,6 +27,9 @@ module Packwerk
 
     sig { returns(T::Array[Packwerk::Offense]) }
     attr_reader :errors
+
+    sig { returns(T::Array[Packwerk::ReferenceOffense]) }
+    attr_reader :strict_mode_violations
 
     sig do
       params(offense: Packwerk::Offense)
@@ -46,9 +50,11 @@ module Packwerk
         @errors << offense
         return
       end
-      package_todo = package_todo_for(offense.reference.package)
-      unless package_todo.add_entries(offense.reference, offense.violation_type)
+
+      if !already_listed?(offense)
         new_violations << offense
+      elsif strict_mode_violation?(offense)
+        strict_mode_violations << offense
       end
     end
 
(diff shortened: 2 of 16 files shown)


lib/packwerk/checker.rb
```diff
@@ -22,11 +22,28 @@ def included(base)
...
+      def find(violation_type)
+        checker_by_violation_type(violation_type)
+      end
+
+      private
+
+      sig { params(name: String).returns(Checker) }
+      def checker_by_violation_type(name)
+        @checker_by_violation_type ||= T.let(@checker_by_violation_type, T.nilable(T::Hash[String, T.nilable(Checker)]))
+        @checker_by_violation_type ||= Checker.all.map { |c| [c.violation_type, c] }.to_h
```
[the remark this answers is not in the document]
alexevanczuk: Sounds good

lib/packwerk/offense_collection.rb
```diff
@@ -46,9 +50,15 @@ def add_offense(offense)
         @errors << offense
         return
       end
-      package_todo = package_todo_for(offense.reference.source_package)
-      unless package_todo.add_entries(offense.reference, offense.violation_type)
+
+      # require 'pry'
+      # binding.pry if offense.reference.constant.name.include?('Squid')
```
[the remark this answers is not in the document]
alexevanczuk: Whoops, yep!

lib/packwerk/checker.rb
```diff
@@ -22,11 +22,29 @@ def included(base)
...
+      sig { params(violation_type: String).returns(Checker) }
+      def find(violation_type)
+        checker_by_violation_type(violation_type)
+      end
+
+      private
+
+      sig { params(name: String).returns(Checker) }
+      def checker_by_violation_type(name)
+        @checker_by_violation_type ||= T.let(Checker.all.map do |c|
```
gmcgibbon, line 35:
```suggestion
        @checker_by_violation_type ||= T.let(Checker.all.to_h do |checker|
```
You can use Array#to_h to map and cast in one call.

test/unit/cli_test.rb
```diff
@@ -143,6 +143,10 @@ def show_stale_violations(_offense_collection, _fileset)
         def identifier
           "custom"
         end
+
+        def show_strict_mode_violations(offenses)
```
gmcgibbon, line 147:
```suggestion
        def show_strict_mode_violations(_offenses)
```

lib/packwerk/package.rb
```diff
@@ -30,7 +30,12 @@ def initialize(name:, config:)
 
     sig { returns(T::Boolean) }
     def enforce_dependencies?
-      @config["enforce_dependencies"] == true
+      [true, "strict"].include?(@config["enforce_dependencies"])
+    end
+
+    sig { returns(T::Boolean) }
+    def strictly_enforce_dependencies?
```
gmcgibbon, line 37: Is this used?

lib/packwerk/offense_collection.rb
```diff
@@ -46,9 +50,13 @@ def add_offense(offense)
         @errors << offense
         return
       end
-      package_todo = package_todo_for(offense.reference.package)
-      unless package_todo.add_entries(offense.reference, offense.violation_type)
+
+      already_listed = package_todo_for(offense.reference.package).add_entries(offense.reference,
+        offense.violation_type)
+      if !already_listed
```
gmcgibbon, line 56: Let's make this a private method too for readability.

test/unit/checker_test.rb
```diff
@@ -0,0 +1,17 @@
+# typed: true
+# frozen_string_literal: true
+
+require "test_helper"
+
+module Packwerk
+  class CheckerTest < Minitest::Test
+    include FactoryHelper
```
gmcgibbon, line 8:
```suggestion
```

test/unit/checker_test.rb
```diff
@@ -0,0 +1,17 @@
...
+
+    test "#find is correctly able to find the right checker" do
+      # Load in the checker so its registered
+      require "packwerk/reference_checking/checkers/dependency_checker"
```
gmcgibbon, line 12: I assume https://github.com/Shopify/packwerk/blob/72aa3cf9dd1c4035356e5474e115583bada071da/lib/packwerk/run_context.rb#L12 is implicitly loading this in non-test runs. Should we just add a require to the main entrypoint like we did with the offense formatter?

alexevanczuk: Yep, makes a lot of sense. This was also able to remove the unused `DEFAULT_CHECKERS` param which was just implicitly used to register the constant.

lib/packwerk/package.rb
alexevanczuk: Nope! An artifact of a spike that I didn't end up using, removing.

lib/packwerk/offense_collection.rb
alexevanczuk: sounds good

alexevanczuk:
Incorporated the newest round of feedback @gmcgibbon!

gmcgibbon (approved):

PR #272, opened 2022-11-30
Don't allow update on specific files

gmcgibbon:

When updating the todo, Packwerk should know about all packages in an application. Most of the time, running update-todo with arguments doesn't make sense and is prone to errors. So, let's stop this usage and make sure developers are always updating todos without arguments.

Pass the `FilesForProcessing` object instead of the file set specifically to provide more info about the file-set, and to lazily fetch / filter the files.

Does this make sense?

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist


alexevanczuk (approved):
Seems good to me, thanks for doing this!

Issue #275, opened 2022-12-06
[Packwerk 3.0] Ensure API boundaries are up-to-date


alexevanczuk:
We want to use this major version release to take the opportunity to mark undocumented API as explicitly private, using any combination of:
- `private`
- `private_class_method`
- `private_constant`
- `# :node` or whatever YARD uses

Similarly, we'll want to make sure all public API has clear, intentional, and documented interfaces.

PR #276, opened 2022-12-06
Mark non-documented API as explicitly private

alexevanczuk:
# What are you trying to accomplish?
This PR addresses a portion of #275.

In this PR, I mark many constants as explicitly private API.

PR is best reviewed commit by commit with hide whitespace changes set to true.

Alternate implementation: https://github.com/Shopify/packwerk/pull/282/commits

I used `private_constant` as the vanilla Ruby way to mark something as private. I did this instead of something like `:nodoc:`, which requires folks to either read the source code or look at code with YARD. `private_constant` gives the user unavoidable feedback when their code is executed.

I separated out the main autoloads into two lists – one to autoload public API and one to autoload private API. This is mostly for the benefit of maintainers to easily see what APIs need to stay supported and what can change.

As a result of this, I needed to change references to these constants to omit the leading `Packwerk` namespace, which would constitute a private (and therefore not allowed) reference.

I had an alternate idea, which was to create a module `Packwerk::Private` where we could move all of our constants. With this approach, we could have a single `private_constant :Private` call that would mark everything as private. I like this pattern because it would mean all public API would be in one place that can easily be seen, and private things would be in their own explicit location. I'd be happy to apply this pattern instead if other maintainers prefer it.

Here is that implementation: https://github.com/Shopify/packwerk/pull/282/commits

- Would folks prefer to move all private implementation into one place, as described above?
- Is there any private API that should be public?
- Is there any public API that should be private?

## Type of Change

- [x] Bugfix
- [ ] New feature

### Additional Release Notes

This is a breaking change because we've marked a lot of things as private. I've left many things as public where we know we are exposing a public API, but its possible some folks are using some of these constants that are currently private. My hope was that they could release an issue against packwerk 3.0 if they have a use-case for these private constants.

## Checklist

### DIFF
--- a/lib/packwerk.rb
+++ b/lib/packwerk.rb
@@ -13,44 +13,58 @@ require "active_support/core_ext/object/json"
 module Packwerk
   extend ActiveSupport::Autoload
 
-  autoload :ApplicationLoadPaths
-  autoload :ApplicationValidator
-  autoload :AssociationInspector
-  autoload :OffenseCollection
-  autoload :Cache
+  # Public APIs
   autoload :Checker
   autoload :Cli
-  autoload :Configuration
-  autoload :ConstantDiscovery
-  autoload :ConstantNameInspector
-  autoload :ConstNodeInspector
-  autoload :PackageTodo
-  autoload :ExtensionLoader
-  autoload :FileProcessor
-  autoload :FilesForProcessing
-  autoload :Graph
-  autoload :Loader
-  autoload :Node
-  autoload :NodeHelpers
-  autoload :NodeProcessor
-  autoload :NodeProcessorFactory
-  autoload :NodeVisitor
   autoload :Offense
   autoload :OffensesFormatter
   autoload :OutputStyle
   autoload :Package
   autoload :PackageSet
-  autoload :ParsedConstantDefinitions
+  autoload :PackageTodo
   autoload :Parsers
-  autoload :ParseRun
-  autoload :UnresolvedReference
   autoload :Reference
-  autoload :ReferenceExtractor
   autoload :ReferenceOffense
-  autoload :Result
-  autoload :RunContext
   autoload :Validator
   autoload :Version
+
+  # Private APIs
+  # Please submit an issue if you have a use-case for these
+  autoload :ApplicationLoadPaths
+  autoload :ApplicationValidator
+  autoload :AssociationInspector
+  autoload :Cache
+  autoload :Configuration
+  autoload :ConstantDiscovery
+  autoload :ConstantNameInspector
+  autoload :ConstNodeInspector
+  autoload :ExtensionLoader
+  autoload :FileProcessor
+  autoload :FilesForProcessing
+  autoload :Graph
+  autoload :Node
+  autoload :NodeHelpers
+  autoload :NodeProcessor
+  autoload :NodeProcessorFactory
+  autoload :NodeVisitor
+  autoload :OffenseCollection
+  autoload :ParsedConstantDefinitions
+  autoload :ParseRun
+  autoload :ReferenceExtractor
+  autoload :RunContext
+  autoload :UnresolvedReference
+
+  module Validator
+    extend ActiveSupport::Autoload
+
+    autoload :Result
+  end
+
+  class Cli
+    extend ActiveSupport::Autoload
+
+    autoload :Result
+  end
 
   module OutputStyles
     extend ActiveSupport::Autoload
@@ -69,12 +83,16 @@ module Packwerk
     autoload :ProgressFormatter
   end
 
+  private_constant :Formatters
+
   module Generators
     extend ActiveSupport::Autoload
 
     autoload :ConfigurationFile
     autoload :RootPackage
   end
+
+  private_constant :Generators
 
   module ReferenceChecking
     extend ActiveSupport::Autoload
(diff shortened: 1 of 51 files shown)


lib/packwerk.rb
```diff
@@ -13,44 +13,74 @@
 module Packwerk
   extend ActiveSupport::Autoload
 
+  # Public APIs
```
alexevanczuk, line 16: I chose these APIs with these rationales:

### Checker
API for extending checkers

### Cli
Main API for calling packwerk right now. Eventually we might want API like `Packwerk.check` instead of using the CLI, but this is the de-facto API.

### Offense
Part of the interface for `OffensesFormatter`

### OffensesFormatter
API for extending offenses formatter

### OutputStyle
Implementation of this needs to be passed into CLI

### Package, PackageSet, PackageTodo
API for getting information about YML files

### Parsers
Contains documented API for `Packwerk::Parsers::Factory.erb_parser_class=(...)`, which is the current API for substituting the ERB parser.
The hope is that https://github.com/Shopify/packwerk/pull/243 will allow us to mark this API as private.

This module also contains `ParserInterface` as well. With the above PR, we'd allow the interface to be public but the rest to be private.

### Reference, ReferenceOffense
Part of the API to `Checker`

### Result
Return value of CLI

gmcgibbon (commented):
I agree with the shape of this API, but I think we need to label each class as `@api private` and make sure the public ones are documented (at least at the class/module level). Though, maybe `private_constant` does this for us. @rafaelfranca do you have an opinion on using `private_constant` here?

Also, can we confirm `private_constant` works on autoloaded constants without loading them?

lib/packwerk.rb
gmcgibbon: Result is overloaded, I vote to nest the toplevel result in CLI. We should also nest the app validator result in Validator instead, especially if we use it in the interface.

alexevanczuk:
@gmcgibbon 

I created the alternate implementation where I nest everything under `Private` here: https://github.com/Shopify/packwerk/pull/282

Let me know what you think! The diff is a bit messy when not viewed commit by commit since it looks like a bunch of files are deleted/renamed, so its easier to just review the last four commits (since that PR is based off of this one) one by one.

test/unit/application_validator_test.rb
```diff
@@ -73,9 +73,9 @@ class ApplicationValidatorTest < Minitest::Test
 
     private
 
-    sig { returns(Packwerk::ApplicationValidator) }
+    sig { returns(ApplicationValidator) }
```
gmcgibbon, line 76: It looks like unit tests also are missing the toplevel packwerk folder to match the module nesting. We aren't autoloading our tests, but this would make it easier if we ever chose to.

lib/packwerk.rb
```diff
@@ -13,44 +13,84 @@
...
+  private_constant :NodeVisitor
+  autoload :OffenseCollection
+  private_constant :OffenseCollection
   autoload :ParsedConstantDefinitions
-  autoload :Parsers
+  private_constant :ParsedConstantDefinitions
   autoload :ParseRun
-  autoload :UnresolvedReference
-  autoload :Reference
+  private_constant :ParseRun
```
gmcgibbon, line 73: I've always wondered why this is called `ParseRun`. Should we just rename it to `Run`? It is private API so we can do this for free later if need be.

USAGE.md
```diff
@@ -265,7 +265,7 @@ You can also reference the name of a gem.
 
 While `packwerk` ships with its own offense formatter, you may specify a custom one in your configuration file via the `offenses_formatter:` key.  Your custom formatter will be used when `bin/packwerk check` is run.
 
-Firstly, you'll need to create an `OffensesFormatter` class that includes `Packwerk::OffensesFormatter`. You can use [`Packwerk::Formatters::OffensesFormatter`](lib/packwerk/formatters/offenses_formatter.rb) as a point of reference for this. Then, in the `require` directive described above, you'll want to tell `packwerk` about it:
+Firstly, you'll need to create an `OffensesFormatter` class that includes `Packwerk::OffensesFormatter`. You can use [`Formatters::OffensesFormatter`](lib/packwerk/formatters/offenses_formatter.rb) as a point of reference for this. Then, in the `require` directive described above, you'll want to tell `packwerk` about it:
```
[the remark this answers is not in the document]
alexevanczuk: For sure, I think this just came from an errant find and replace.

lib/packwerk.rb
```diff
@@ -13,44 +13,84 @@
...
+  autoload :PackageTodo
+  autoload :Parsers
+  autoload :Reference
+  autoload :ReferenceOffense
+  autoload :Validator
+
+  # Private APIs
+  # Please submit an issue if you have a use-case for these
   autoload :ApplicationLoadPaths
+  private_constant :ApplicationLoadPaths
```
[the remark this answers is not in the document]
alexevanczuk: Sure thing

test/unit/application_validator_test.rb
alexevanczuk: Same comment – let me know if you're okay with us moving test files in another PR.

lib/packwerk.rb
alexevanczuk: Not sure about that one either. Personally, my preference would be to just have API on `packwerk` directly, e.g.
```
Packwerk.check
Packwerk.update_todo
```

This feels like the best UX to me.

test/parser_test_helper.rb
```diff
@@ -1,10 +1,12 @@
 # typed: true
 # frozen_string_literal: true
 
-module ParserTestHelper
-  class << self
-    def parse(source)
-      Packwerk::Parsers::Ruby.new.call(io: StringIO.new(source))
+module Packwerk
```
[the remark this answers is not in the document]
gmcgibbon: Yes, this is noisy enough, we can do it later.

gmcgibbon (approved):

PR #280, opened 2022-12-08
Better validations

gmcgibbon:

Various fixes with validation.

See commits. Some refactoring to parse run and progress formatter. Also some fixing of validation cases where steps of the app validator would trip over each other (circular dependency checks didn't work with nonexistent packages and dependency checks wouldn't work with non-array values).

Do these changes make sense?

## Type of Change

- [x] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist


alexevanczuk (approved):
Thanks!!

PR #283, opened 2022-12-13
More strict typing

gmcgibbon:

Convert more files to typed: strict. Also fix a bug in `Cli#validate` unveiled by new sigs where `validation_started` is returning early (before the block starts). This adds extra text to the validation step, which I can remove if we don't like it. It seems we always wanted it to act like this, but never tested it.

Convert more files to `typed: strict`. I found dead code (an unreachable IF statement) and investigated an interesting regression in `T::Struct` in the last strong typing PR, so I think this work is worth doing. 

Does the parser approach make sense? I've never used generics in Sorbet before.

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist

### DIFF
--- a/lib/packwerk/parsers/erb.rb
+++ b/lib/packwerk/parsers/erb.rb
@@ -1,4 +1,4 @@
-# typed: true
+# typed: strict
 # frozen_string_literal: true
 
 require "ast/node"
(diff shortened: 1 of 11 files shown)


lib/packwerk/parsers/erb.rb
```diff
@@ -35,12 +40,18 @@ def parse_buffer(buffer, file_path:)
 
       private
 
+      sig do
+        type_parameters(:T).params(
+          erb_ast: T.all(T.type_parameter(:T), Object),
+          file_path: String
+        ).returns(T.type_parameter(:T))
+      end
       def to_ruby_ast(erb_ast, file_path)
```
alexevanczuk, line 49: Can the `erb_ast` be many things? Could/should we just enumerate what it can be? Why generic?

alexevanczuk (approved):
Thank you!

lib/packwerk/parsers/erb.rb
gmcgibbon: So it turns out the AST value can either be a node, a string, and nil. Using these types in a `T.any` clause works.

PR #284, opened 2022-12-14
Allow ERB parser to_ruby_ast to return nil

alexevanczuk:
After we added types, I found this method returned `nil` on our monolith sometimes, causing `bin/packwerk update-todo` unable to execute properly.

I updated the sig to ensure it continues to work as expected. We might want to come back later and...
1) figure out when/how its `nil`
2) Add a test for when its `nil` (or possible fix it so its never `nil` if that turns out to be a better approach)

## Type of Change

- [x] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist

### DIFF
--- a/lib/packwerk/parsers/erb.rb
+++ b/lib/packwerk/parsers/erb.rb
@@ -15,7 +15,7 @@ module Packwerk
 
       sig { params(parser_class: T.untyped, ruby_parser: Ruby).void }
       def initialize(parser_class: BetterHtml::Parser, ruby_parser: Ruby.new)
-        @parser_class = parser_class
+        @parser_class = T.let(parser_class, T.class_of(BetterHtml::Parser))
         @ruby_parser = ruby_parser
       end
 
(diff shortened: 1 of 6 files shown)


lib/packwerk/parsers/erb.rb
```diff
@@ -44,7 +44,7 @@ def parse_buffer(buffer, file_path:)
         params(
           erb_ast: T.all(::AST::Node, Object),
           file_path: String
-        ).returns(::AST::Node)
+        ).returns(T.nilable(::AST::Node))
```
gmcgibbon, line 47: Is it possible to add coverage? If it isn't covered in our test suite I can see this being removed again.

alexevanczuk: Yep, I can look more into it.

alexevanczuk: It looks like ERB can contain javascript which returns `nil` rather than a node. I added a simple test to demonstrate this. I also added more types to the parser class so that the nil type is also supported by the return value of the parser classes.

PR #285, opened 2022-12-14
Remove private_constant from some API with valid use cases

alexevanczuk:
Allow clients to use API that was marked `private_constant` in a recent PR.

I removed `private_constant`.

Any reason not to make these publics?

## Type of Change

- [x] Bugfix
- [ ] New feature

## Checklist

### DIFF
--- a/lib/packwerk.rb
+++ b/lib/packwerk.rb
@@ -14,8 +14,11 @@ module Packwerk
   extend ActiveSupport::Autoload
 
   # Public APIs
+  autoload :ApplicationLoadPaths
   autoload :Checker
   autoload :Cli
+  autoload :ConstantContext
+  autoload :Node
   autoload :Offense
   autoload :OffensesFormatter
   autoload :OutputStyle
@@ -27,38 +30,6 @@ module Packwerk
   autoload :ReferenceOffense
   autoload :Validator
   autoload :Version
-
-  # Private APIs
-  # Please submit an issue if you have a use-case for these
-  autoload :ApplicationLoadPaths
-  autoload :ApplicationValidator
-  autoload :AssociationInspector
-  autoload :Cache
-  autoload :Configuration
-  autoload :ConstantDiscovery
-  autoload :ConstantNameInspector
-  autoload :ConstNodeInspector
-  autoload :ExtensionLoader
-  autoload :FileProcessor
-  autoload :FilesForProcessing
-  autoload :Graph
-  autoload :Node
-  autoload :NodeHelpers
-  autoload :NodeProcessor
-  autoload :NodeProcessorFactory
-  autoload :NodeVisitor
-  autoload :OffenseCollection
-  autoload :ParsedConstantDefinitions
-  autoload :ParseRun
-  autoload :ReferenceExtractor
-  autoload :RunContext
-  autoload :UnresolvedReference
-
-  module Validator
-    extend ActiveSupport::Autoload
-
-    autoload :Result
-  end
 
   class Cli
     extend ActiveSupport::Autoload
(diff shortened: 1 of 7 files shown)


lib/packwerk.rb
```diff
@@ -14,8 +14,11 @@ module Packwerk
   extend ActiveSupport::Autoload
 
   # Public APIs
+  autoload :ApplicationLoadPaths
```
alexevanczuk, line 17: I stub `ApplicationLoadPaths` so that I can run `packwerk` in some integration tests for some things in the broader packwerk ecosystem.

lib/packwerk.rb
```diff
@@ -14,8 +14,11 @@ module Packwerk
...
   autoload :Checker
   autoload :Cli
+  autoload :ConstantDiscovery
+  autoload :Node
```
alexevanczuk, line 21: Needed to build a `Packwerk::Reference`, existing public API.

lib/packwerk.rb
```diff
@@ -14,8 +14,11 @@ module Packwerk
   extend ActiveSupport::Autoload
 
   # Public APIs
+  autoload :ApplicationLoadPaths
   autoload :Checker
   autoload :Cli
+  autoload :ConstantDiscovery
```
alexevanczuk, line 20: Needed to build a `Packwerk::Reference`, existing public API.

lib/packwerk.rb
gmcgibbon: Can't you just stub the caller method in configuration? Why specifically this module?

lib/packwerk.rb
gmcgibbon: Side-note, why is `Packwerk::Node` a class when it only nests `Location`? Can we make it a module?

lib/packwerk.rb
gmcgibbon: I assume that's because we need `ConstantContext`? Should we just move this toplevel? We don't have a valid use-case of using it in an extension, do we?

lib/packwerk.rb
alexevanczuk: I probably can -- I'll look into this!

alexevanczuk: So... I use `ApplicationLoadPaths` in `danger-packwerk` to initialize a `ConstantResolver`: https://github.com/rubyatscale/danger-packwerk/blob/cbcd07cf42f7f555e02e0b1ac36552142a4262d4/lib/danger-packwerk/private.rb#L17, which we use to find where a constant is defined: https://github.com/rubyatscale/danger-packwerk/blob/cbcd07cf42f7f555e02e0b1ac36552142a4262d4/lib/danger-packwerk/danger_package_todo_yml_changes.rb#L121 for the purpose of being able to link to constant definitions when we leave GitHub comments through our Danger integration.

I understand it's not exactly the intended public API of packwerk, and I always could just copy and paste this code into `danger-packwerk` for it to get the load paths itself, but wondering if this would be alright to use until we find a better solution. Also open to other suggestions you have!

lib/packwerk.rb
alexevanczuk: Yes, we only use `ConstantContext`. That makes sense – I've moved it to the top-level so it can be public without all of `ConstantDiscovery` being public.

lib/packwerk.rb
```diff
@@ -83,8 +83,6 @@ module Formatters
     autoload :ProgressFormatter
   end
 
-  private_constant :Formatters
```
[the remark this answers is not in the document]
alexevanczuk: Changed!

lib/packwerk.rb
gmcgibbon: I have an issue making `ApplicationLoadPaths` public for 3 reasons:

- I think it is better to name it `LoadPaths.for(root, evironment:)` or similar.
- It looks like we are exposing public methods we don't need to in it. `extract_relevant_paths` should be the only public method.
- We use a railties as a peer dependency (we reference `Rails` but don't depend on railties or require it anywhere). This access should be safer and tolerant of it not being there. Or, we should use proper peer dependency syntax (see [example](https://github.com/Shopify/rails/blob/656118bc17791291db7c10075b489dfc6f93fce3/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L3-L4)).

alexevanczuk: These are valid concerns. How do you feel about us merging this PR as is and I do a fast follow with these updates? Happy to do first if you'd prefer.

gmcgibbon: Approved, we can do this next.

PR #287, opened 2022-12-15
Move test files so that file structure matches constant nesting structure

alexevanczuk:
Follow-up to https://github.com/Shopify/packwerk/pull/276#discussion_r1047741870

In this PR, we're moving test files into folder structures that matches their naming convention. Although we are not using `zeitwerk` to load test files, we thought this is a best practice that improves navigatability of the codebase.

I chose `test/*` as a sort of "autoload" path in zeitwerk parlance, with anything that follows that to constitute a portion of the constant namespacing.

That is, `test/support` and `test/integration` are organizational units to organize test files into general buckets. Everything that follows represents a portion of the namespace.

Does the nesting make sense?
Should we enforce this by actually loading test constants with zeitwerk?

## Type of Change

- [ ] Bugfix
- [ ] New feature

## Checklist


gmcgibbon (approved):
Thanks!

PR #288, opened 2022-12-15
Update ApplicationLoadPaths to make more things private

alexevanczuk:
Follow up to https://github.com/Shopify/packwerk/pull/285#discussion_r1050106654

We wanted to update `ApplicationLoadPaths` to make more things private.

1) I changed the name as suggested
2) I didn't update the API to `LoadPaths.for(...)`, as I didn't see why we needed to create an instance and hold state here. Happy to change this if it's still desired.
3) I updated the tests to still test the same things as before, but through the public API rather than the private API.

Let me know if what is public and its interface + tests make sense.

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

We are changing `ApplicationLoadPaths` to `LoadPaths`, so clients will have to update their callsites accordingly.

## Checklist

We'll still need to add some more documentation on this.


lib/packwerk/load_paths.rb
```diff
@@ -5,7 +5,7 @@
 
 module Packwerk
   # Extracts the load paths from the analyzed application so that we can map constant names to paths.
-  module ApplicationLoadPaths
+  module LoadPaths
```
gmcgibbon, line 8: Hm, now that I read this more closely, I think we should name this `RailsLoadPaths` and go the peer dependency route. We can make this interface generalized and switch the module out for something else later as needed (if we support other frameworks, for example).

lib/packwerk/configuration.rb
```diff
@@ -61,7 +61,7 @@ def initialize(configs = {}, config_path: nil)
     end
 
     def load_paths
-      @load_paths ||= ApplicationLoadPaths.extract_relevant_paths(@root_path, "test")
+      @load_paths ||= LoadPaths.extract_relevant_paths(@root_path, "test")
```
gmcgibbon, line 64:
```suggestion
      @load_paths ||= RailsLoadPaths.for(@root_path, environment: "test")
```

> I didn't update the API to LoadPaths.for(...), as I didn't see why we needed to create an instance and hold state here. Happy to change this if it's still desired.

I'm not sure I follow this comment. I meant for this line to work, so the signature would still be on the module and the environment would be a keyword. Does that make sense to you?

gmcgibbon (commented):
>

alexevanczuk:
> Looks good to me, but we need to make `railties` a peer dep still so non-rails usage will crash elegantly.

Oh right! I knew I was forgetting something. Will update that tomorrow.

lib/packwerk/configuration.rb
alexevanczuk: Yes that makes sense to me! [This comment](https://github.com/Shopify/packwerk/pull/285#discussion_r1050106654) had me confused 
> I think it is better to name it LoadPaths.for(root, evironment:) or similar.
> It looks like we are exposing public methods we don't need to in it. extract_relevant_paths should be the only public method.

I thought you wanted the method to still be `extract_relevant_paths`, e.g. `LoadPaths.for(...).extract_relevant_paths`

I see what you're saying now though.

packwerk.gemspec
```diff
@@ -47,6 +47,9 @@ Gem::Specification.new do |spec|
   spec.add_dependency("sorbet-runtime", ">= 0.5.9914")
   spec.add_dependency("zeitwerk", ">= 2.6.1")
 
+  # For `RailsLoadPaths`
+  spec.add_dependency("railties")
```
gmcgibbon, line 51: I think we use a peer dependency instead because I still think there's a use-case for non-rails now that frameworks like hanami are using Zeitwerk. See example in https://github.com/Shopify/packwerk/pull/285#discussion_r1050106654.

alexevanczuk: Interesting okay.

alexevanczuk: Okay I tried, let me know what you think. I also removed the `rails` dependency.

lib/packwerk/rails_load_paths.rb
```diff
@@ -2,22 +2,26 @@
 # frozen_string_literal: true
 
 require "bundler"
+gem "railties"
+require "rails/railtie"
```
gmcgibbon, line 6:
```suggestion
require "rails"
```

packwerk.gemspec
```diff
@@ -53,4 +53,6 @@ Gem::Specification.new do |spec|
 
   # For ERB parsing
   spec.add_dependency("better_html")
+
+  spec.add_development_dependency("rails")
```
gmcgibbon, line 57:
```suggestion
  spec.add_development_dependency("railties")
```

gmcgibbon (approved):

lib/packwerk/rails_load_paths.rb
```diff
@@ -2,22 +2,26 @@
 # frozen_string_literal: true
 
 require "bundler"
+gem "railties"
```
gmcgibbon, line 5: We should constrain this to the version that adds `Rails.autoloaders` with Zeitwerk by default.

gmcgibbon: I think this should be the version where Rails removed the classic autoloader. So that's 7.0: https://guides.rubyonrails.org/classic_to_zeitwerk_howto.html#applications-running-rails-7

PR #290, opened 2022-12-16
Fix bug with dependency validator strict mode

alexevanczuk:
Allow strict mode to work for dependency validator.

Compare value (instead of the class) against the string `strict`.

n/a

## Type of Change

- [x] Bugfix
- [ ] New feature

## Checklist

### DIFF
--- a/lib/packwerk/validators/dependency_validator.rb
+++ b/lib/packwerk/validators/dependency_validator.rb
@@ -34,10 +34,11 @@ module Packwerk
       def check_package_manifest_syntax(configuration)
         errors = []
 
+        valid_settings = [true, false, "strict"]
         package_manifests_settings_for(configuration, "enforce_dependencies").each do |config, setting|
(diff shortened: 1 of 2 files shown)


lib/packwerk/validators/dependency_validator.rb
```diff
@@ -37,7 +37,7 @@ def check_package_manifest_syntax(configuration)
         package_manifests_settings_for(configuration, "enforce_dependencies").each do |config, setting|
           next if setting.nil?
 
-          unless [TrueClass, FalseClass, "strict"].include?(setting.class)
+          unless [TrueClass, FalseClass].include?(setting.class) || setting == "strict"
```
gmcgibbon, line 40:
```suggestion
          unless [true, false, "strict"].include?(setting)
```
Can't we just do this instead?

alexevanczuk:
Updated @gmcgibbon !

PR #291, opened 2022-12-16
Even more strict typing

gmcgibbon:

Strict type remaining lib files, except for structs.

Better type coverage. Found some redundant code from this change.

We should find a way to type structs, but I don't see an easy way to do it right now without re-defining methods.

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist


lib/packwerk/configuration.rb
```diff
@@ -32,26 +36,51 @@ def from_packwerk_config(path)
...
-      @root_path = File.expand_path(root)
-      @package_paths = configs["package_paths"] || "**/"
-      @custom_associations = configs["custom_associations"] || []
-      @parallel = configs.key?("parallel") ? configs["parallel"] : true
-      @cache_enabled = configs.key?("cache") ? configs["cache"] : false
-      @cache_directory = Pathname.new(configs["cache_directory"] || "tmp/cache/packwerk")
+      @root_path = T.let(File.expand_path(root), String)
+      @package_paths = T.let(configs["package_paths"] || "**/", T.untyped)
+      @custom_associations = T.let(configs["custom_associations"] || [], T.untyped)
+      @parallel = T.let(configs.key?("parallel") ? configs["parallel"] : true, T.untyped)
```
alexevanczuk, line 76: Could be `T::Boolean`

lib/packwerk/configuration.rb
```diff
@@ -60,10 +89,15 @@ def initialize(configs = {}, config_path: nil)
       end
     end
 
+    sig { returns(T::Hash[String, Module]) }
     def load_paths
-      @load_paths ||= ApplicationLoadPaths.extract_relevant_paths(@root_path, "test")
+      @load_paths ||= T.let(
+        ApplicationLoadPaths.extract_relevant_paths(@root_path, "test"),
```
alexevanczuk, line 95: Will be some merge conflicts on this now that its `RailsLoadPaths.for`

lib/packwerk/configuration.rb
gmcgibbon: I didn't want to strongly type some of the config values in the constructor because I think sorbet would complain before validation errors surface. I'll look into it before shipping because I don't remember after vacation.

gmcgibbon: It just fails type checks somewhere else. I'll cast them to what they're supposed to be. If it isn't covered by our validation tests, we can fix it later.

lib/packwerk/configuration.rb
```diff
@@ -9,6 +9,9 @@ class Configuration
     extend T::Sig
 
     class << self
+      extend T::Sig
```
davidstosik, line 12: @gmcgibbon I'm not sure what I'm talking about, but it looks like these changes needed a new `require`.

Because of this:

https://github.com/Shopify/packwerk/blob/0d44cc3779968989e5cf617b40ef245f617b8a0d/USAGE.md?plain=1#L68

It looks like my application is not able to start anymore:

```
$ bin/rails c
/home/spin/src/github.com/Shopify/packwerk/lib/packwerk/spring_command.rb:9:in `<class:SpringCommand>': uninitialized constant Packwerk::SpringCommand::T (NameError)

    extend T::Sig
            ^^^^^
        from /home/spin/src/github.com/Shopify/packwerk/lib/packwerk/spring_command.rb:8:in `<module:Packwerk>'
        from /home/spin/src/github.com/Shopify/packwerk/lib/packwerk/spring_command.rb:7:in `<top (required)>'
        from /home/spin/src/github.com/Shopify/shopify/config/spring.rb:4:in `require'
```

Should we add `require "sorbet-runtime"` at the top of this file?

davidstosik: Update: I was able to reproduce the problem in a fresh Rails minimal application after making these changes to it:

```diff
diff --git c/Gemfile w/Gemfile
index a329896..acca730 100644
--- c/Gemfile
+++ w/Gemfile
@@ -25,6 +25,6 @@ end
 
 group :development do
   # Speed up commands on slow machines / big apps [https://github.com/rails/spring]
-  # gem "spring"
+  gem "spring"
+  gem "packwerk", github: "Shopify/packwerk", branch: "main"
 end
-
diff --git c/bin/rails w/bin/rails
index efc0377..c8b5338 100755
--- c/bin/rails
+++ w/bin/rails
@@ -1,4 +1,5 @@
 #!/usr/bin/env ruby
+load File.expand_path("spring", __dir__)
 APP_PATH = File.expand_path("../config/application", __dir__)
 require_relative "../config/boot"
 require "rails/commands"
diff --git c/bin/rake w/bin/rake
index 4fbf10b..7327f47 100755
--- c/bin/rake
+++ w/bin/rake
@@ -1,4 +1,5 @@
 #!/usr/bin/env ruby
+load File.expand_path("spring", __dir__)
 require_relative "../config/boot"
 require "rake"
 Rake.application.run
diff --git c/bin/spring w/bin/spring
new file mode 100755
index 0000000..37f429b
--- /dev/null
+++ w/bin/spring
@@ -0,0 +1,14 @@
+#!/usr/bin/env ruby
+
+# This file loads Spring without using loading other gems in the Gemfile, in order to be fast.
+# It gets overwritten when you run the `spring binstub` command.
+
+if !defined?(Spring) && [nil, "development", "test"].include?(ENV["RAILS_ENV"])
+  require "bundler"
+
+  Bundler.locked_gems.specs.find { |spec| spec.name == "spring" }&.tap do |spring|
+    Gem.use_paths Gem.dir, Bundler.bundle_path.to_s, *Gem.path
+    gem "spring", spring.version
+    require "spring/binstub"
+  end
+end
diff --git c/config/spring.rb w/config/spring.rb
new file mode 100644
index 0000000..a485fb4
--- /dev/null
+++ w/config/spring.rb
@@ -0,0 +1 @@
+require "packwerk/spring_command"
```

PR #292, opened 2022-12-19
Fix error message for strict mode

alexevanczuk:
Since a checker can produce a violation if the referencing OR the defining package declares a violation AND we have no way of knowing which (referencing or defining) package is the one that produces the violation, the error message on strict mode cannot suggest exactly which `package.yml` file is at cause.

I chose the route of just making the reference to the `package.yml` file ambiguous and leaving the reader to determine which `package.yml` to look at if they want to see where `strict` mode is set.

An alternative to this would be to look at https://github.com/Shopify/packwerk/pull/261, which allows the checker to define which package is the one that enforces violations and then to use that package in the error message. Since that change has less general agreement, I thought this would be a simpler stopgap.

Let me know if there's anything I'm missing.

## Type of Change

- [x] Bugfix
- [ ] New feature

## Checklist

### DIFF
--- a/test/unit/packwerk/parse_run_test.rb
+++ b/test/unit/packwerk/parse_run_test.rb
@@ -376,7 +376,7 @@ module Packwerk
       expected_message = <<~EOS
         No offenses detected
         No stale violations detected
-        components/source cannot have dependency violations on components/destination because strict mode is enabled for dependency violations in components/source/package.yml
+        components/source cannot have dependency violations on components/destination because strict mode is enabled for dependency violations in the enforcing package's package.yml
       EOS
       assert_equal expected_message, result.message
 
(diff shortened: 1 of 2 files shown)


gmcgibbon (approved):

test/unit/packwerk/parse_run_test.rb
```diff
@@ -376,7 +376,7 @@ def something
       expected_message = <<~EOS
         No offenses detected
         No stale violations detected
-        components/source cannot have dependency violations on components/destination because strict mode is enabled for dependency violations in components/source/package.yml
+        components/source cannot have dependency violations on components/destination because strict mode is enabled for dependency violations in the enforcing package's package.yml
```
gmcgibbon, line 379: Should we add a test for the other case?

alexevanczuk: I started adding a test but I think it's okay without, since the other case would just involve other test setup, but the same assertions and code paths. If there was a regression, it would show up equally in another test vs. the current one, since whether a checker's enforcement is on the referencing or consuming package is something that isn't part of the interface quite yet (just a conceptual consideration).

PR #293, opened 2022-12-19
Stop aborting when no files are input into bin/packwerk check

alexevanczuk:
When using `bin/packwerk check` as a pre-commit hook, packwerk will error if only excluded files are passed in.

Instead, I think it makes more sense to simply succeed gracefully if only excluded files are passed in, in which case `bin/packwerk check` should return no errors.
This is more like rubocop's behavior.

I think the intent of the check was to be defensive in case a user is doing something that they maybe shouldn't be doing (testing excluded files). I think it actually ends up producing more confusion than its worth.

Let me know if there are other valid use cases for this check that I should be aware of and if we should solve this another way.

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

I think this is technically a breaking change if someone depends on packwerk aborting in this case.

## Checklist

### DIFF
--- a/lib/packwerk/cli.rb
+++ b/lib/packwerk/cli.rb
@@ -138,7 +138,7 @@ module Packwerk
         ignore_nested_packages: ignore_nested_packages,
         configuration: @configuration
       )
-      abort(<<~MSG.squish) if files_for_processing.files.empty?
+      @out.puts(<<~MSG.squish) if files_for_processing.files.empty?
         No files found or given.
         Specify files or check the include and exclude glob in the config file.
       MSG


lib/packwerk/cli.rb
```diff
@@ -133,17 +133,11 @@ def output_result(result)
       ).returns(FilesForProcessing)
     end
     def fetch_files_to_process(relative_file_paths, ignore_nested_packages)
-      files_for_processing = FilesForProcessing.fetch(
+      FilesForProcessing.fetch(
```
gmcgibbon, line 136: I feel like this is only needed if ARGV is empty. If we filter everything out and silently fail I think that's unhelpful to the developer. Can we change it to only abort when we provide files, but none were found.

alexevanczuk: Good call, I changed this to just warn with a `puts`.

lib/packwerk/cli.rb
```diff
@@ -138,7 +138,7 @@ def fetch_files_to_process(relative_file_paths, ignore_nested_packages)
         ignore_nested_packages: ignore_nested_packages,
         configuration: @configuration
       )
-      abort(<<~MSG.squish) if files_for_processing.files.empty?
+      puts(<<~MSG.squish) if files_for_processing.files.empty?
```
gmcgibbon, line 141:
```suggestion
      @out.puts(<<~MSG.squish) if files_for_processing.files.empty?
```

gmcgibbon (approved):

PR #300, opened 2023-02-09
Move sorbet disable require

gmcgibbon:

Make sorbet spring command work when using a binscript (eg. `bin/packwerk`).

Disable sorbet at the top of the spring command file so that sigs evaluated in the before the disable are ignored.

This isn't easy to test since the suite running tests has type checking enabled already, so I went without.

## Type of Change

- [x] Bugfix
- [ ] New feature

## Checklist


lib/packwerk/disable_sorbet.rb
```diff
@@ -0,0 +1,38 @@
...
+        value
+      end
+
+      def bind(value, type, checked: true)
+        value
+      end
+    end
+  )
+rescue RuntimeError => error
+  if /Set the default checked level earlier./.match?(error.message)
```
iwz, line 33: It's not clear where this string comes from. Maybe extract it to a nicely named variable?

iwz (approved):
LGTM, left a minor bit of feedback, totally fine to ignore it

lib/packwerk/disable_sorbet.rb
gmcgibbon: I'll add a comment with the error's source and it happens. Basically Sorbet throws a RuntimeError if it evaluates signatures before you try to disable typechecking. We match based on the message here because it isn't a custom error type we can account for.

PR #302, opened 2023-02-09
Remove empty_commit file, as it seems unnecessary

davidstosik:

Remove an empty file in the repository's root named `empty_commit`.
The file was added in #276, where I could not find an explanation for its presence.

`git rm`

My assumption is that this file was added in order to create an "empty commit" that would trigger CI.
I'd like to confirm that.

For reference, the following command creates an empty commit without the need to commit any file.

```
git commit --allow-empty --message="Empty commit"
```

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist


gmcgibbon (approved):
🔥

PR #304, opened 2023-02-10
Don't require Packwerk

gmcgibbon:

Don't recommend requiring packwerk.

`gem "packwerk", require: false`

Like other static analysis tools, it is not needed when your app boots, so there's no need to require it.

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Watch a [1-minute video demo](https://www.youtube.com/watch?v=NwqlyBAxVpQ&featur
 1. Add this line to your application's Gemfile:
 
 ```ruby
-gem 'packwerk'
+gem "packwerk", require: false
 ```
 
 2. Install the gem


iwz (approved):
:shipit:

rafaelfranca (changes requested):
Explained in the comment above

gmcgibbon:
> Why?

There is no reason to require a library that is required through an executable script. Rubocop, for example, recommends this in their README so I don't see why Packwerk should be different. Packwerk doesn't slow down boot, but there is zero reason to do it.

andrewn617 (approved):

rafaelfranca:
I feel like expecting people to pass the `require: false` flag in their Gemfile is a failure in UX of the library. If rubocop does that, I'm sorry for then, but we should not be doing this.

exterm:
Historical context - we actually _needed_ to require packwerk at some point because we injected a custom inflector, so that packwerk would have access to the same inflections that the app uses. I've lost track of whether that's still true.

gmcgibbon:
> I feel like the approach we took with tapioca might be better here.

Packwerk won't inherently disable sorbet with the default require, so there is no action to be done. I'll close since you seem to be opposed to this change. We can monitor it in profiling to see if it really actually needs to happen. If for some reason in the future it does, and it does and we can't optimized away, we can revisit this.

PR #310, opened 2023-02-22
chore: add version command

silasb:
I was debugging issues with the newest 2.3.0 release and was having difficulties knowing what version I was running.

I noticed an error (I can pull this out into another PR if needed) while running `rake`, it failed to run specs because a missing `require "pathname"`.

I choose to add a `version` command in to make it easy for anyone running `bundle exec packwerk` to get the version. It feels like best practice to always provide a `version` command or flag for a CLI application.

The output.  I'm not sure if we'd want to use 'v' prefix in-front of the version number.

## Type of Change

- [ ] Bugfix
- [ ] New feature

## Checklist

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -3,6 +3,8 @@
 
 ENV["RAILS_ENV"] = "test"
 
+require "pathname"
+
 $LOAD_PATH.unshift(File.expand_path("../lib", __dir__))
 ROOT = Pathname.new(__dir__).join("..").expand_path
 
--- a/lib/packwerk.rb
+++ b/lib/packwerk.rb
@@ -31,7 +31,6 @@ module Packwerk
   autoload :Reference
   autoload :ReferenceOffense
   autoload :Validator
-  autoload :Version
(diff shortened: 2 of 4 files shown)


test/test_helper.rb
```diff
@@ -3,6 +3,8 @@
 
 ENV["RAILS_ENV"] = "test"
 
+require "pathname"
```
gmcgibbon, line 6: This isn't needed for our setup, but I assume you ran tests without rake to get the error (via `ruby -I`). In that case, we can keep it 👍

gmcgibbon (approved):
LGTM, but this made me realize we have a dead autoload entry [here](https://github.com/Shopify/packwerk/blob/6e57928e1508afd0f651ab68682c06b3a361d399/lib/packwerk.rb#L34). Can you remove it and squash your commits? Thanks

gmcgibbon:
Thanks! For anyone else having issues with this on version of packwerk, you can use `bundle info packwerk` in the interim.

PR #312, opened 2023-02-24
remove leftover inflections yml

exterm:
Since https://github.com/Shopify/packwerk/pull/157, packwerk doesn't use YAML files for inflections anymore since we're now always using the application's inflector. So we don't need this file in the test skeleton anymore either.

## Type of Change

- [ ] Bugfix
- [ ] New feature

## Checklist


gmcgibbon (approved):
Thanks!

PR #316, opened 2023-03-01
Remove more mentions of privacy checking

gmcgibbon:

Remove remaining mentions of privacy checking.

Remove references because they are no longer needed.

Does this make sense?

## Type of Change

- [x] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist

### DIFF
--- a/RESOLVING_VIOLATIONS.md
+++ b/RESOLVING_VIOLATIONS.md
@@ -3,7 +3,7 @@
 Violations can be [recorded as a deprecation](#recording-violations) or (better!) [eliminated](#eliminating-violations).
 
 ## Recording Violations
-💡 New privacy and dependency violations are never hard-blocked. There are many very valid reasons to run `bin/packwerk update-todo`, adding new violations to `package_todo.yml` files. Even if you feel your reason might not be "valid," if your judgement says adding the violation and shipping your change will produce positive impact, trust your gut.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -61,7 +61,6 @@ Various third parties have built tooling on top of packwerk. Here's a selection
 - https://github.com/Gusto/packwerk-vscode integrates packwerk into Visual Studio Code so you can see violations right in your editor
 - https://github.com/Gusto/stimpack sets up Rails autoloading, as well as `rspec` and `FactoryBot` integration, for packages arranged in a flat list. Stimpack is quite convenient, but for autoloading we recommend to use `Rails::Engine`s instead.
 - https://github.com/rubyatscale/danger-packwerk integrates packwerk with [danger.systems](https://danger.systems) to provide packwerk feedback as Github inline PR comments
-- https://github.com/rubyatscale/packwerk-extensions contains extensions for packwerk, including a checker for packwerk that allows you to enforce public API boundaries. This was originally extracted from `packwerk` itself.
```
alexevanczuk, line 64: This still seems like it would be helpful to link as this is a section for surrounding ecosystem gems.

gmcgibbon: Good point. It is under the ecosystem of gems, which doesn't necessarily mean people have to use it. I think the broader change is of removing it from usage is more important. Your gem can own the documentation for privacy checking since it is now responsible for it.

alexevanczuk: `packwerk-extensions` owning documentation for privacy checking makes perfect sense to me 👍

rafaelfranca: Yeah, I forgot we had this list already. If we have all those gems here, it is fine to have the privacy checking gem as well. Sorry for the confusion.

davidstosik (approved):

PR #318, opened 2023-03-02
✏️ including new update-todo command in CLI help

marcreynolds:

This PR updates the 2.3.x CLI `packwerk help` documentation to mention the new `packwerk update-todo` command and adds a deprecation note to the help text for `packwerk update-deprecations`.

If we're going to maintain 2.3.x for a bit, it's useful for devs to know about the new command via the `packwerk help` command and to know about the command deprecation (although executing the command shows the deprecation).

It's probably best to cut a new 2.3.x release with this in it... no need to merge this into `main` or include it in the 3.0.x line.

Grammar, punctuation, and spelling of the text change.

## Type of Change

- [ ] Bugfix
- [ ] New feature

## Checklist

- [ ] ~I have added tests to cover my changes.~

### DIFF
--- a/lib/packwerk/cli.rb
+++ b/lib/packwerk/cli.rb
@@ -125,7 +125,7 @@ module Packwerk
           init - set up packwerk
           check - run all checks
           update-todo - update package_todo.yml files
-          update-deprecations - (deprecated) update package_todo.yml files
+          update-deprecations - update package_todo.yml files (deprecated, use update-todo instead)
           validate - verify integrity of packwerk and package configuration
           help  - display help information about packwerk
       USAGE


marcreynolds:
_Changed merge target from `main` to `2-stable'_

lib/packwerk/cli.rb
```diff
@@ -124,7 +124,8 @@ def usage
         Subcommands:
           init - set up packwerk
           check - run all checks
-          update-deprecations - update deprecated_references.yml files
+          update-todo - update package_todo.yml files
+          update-deprecations - (deprecated) update package_todo.yml files
```
gmcgibbon, line 128:
```suggestion
          update-deprecations - update package_todo.yml files (deprecated, use update-todo instead)
```

gmcgibbon (commented):
Thanks! I'm open to accepting this patch, but I think if people want to use this they can source directly from `2-stable`. It is unlikely we will be releasing another 2.x version.

Issue #327, opened 2023-03-07
[Bug Report] Disable sorbet when using spring


gmcgibbon:
**Description**
Packwerk does not disable sorbet when using Spring. This slows down checks. 

**To Reproduce**

- `bin/packwerk check` says `Packwerk couldn't disable Sorbet. Please ensure it isn't being used before Packwerk is loaded.`

**Expected Behaviour**

Sorbet is successfully disabled, but only for packwerk.

`bin/rails runner "puts ::T::Private::RuntimeLevels.default_checked_level"` should be "always", but packwerk should not print the error.

**Screenshots**

**Version Information**
 - Packwerk: 3.0
 - Ruby 3.2.1

**Additional Context**

https://github.com/Shopify/packwerk/pull/326

PR #328, opened 2023-03-08
Silence sorbet warning when booting with spring

gmcgibbon:

Silence warning when booting with Sorbet.

Before:
```
% bin/packwerk check
Running via Spring preloader in process 904969
Packwerk couldn't disable Sorbet. Please ensure it isn't being used before Packwerk is loaded.
📦 Packwerk is inspecting <n> files
```

After:
```
% bin/packwerk check
Running via Spring preloader in process 904969
📦 Packwerk is inspecting <n> files
```

use `defined?(Spring)`

When booting the Rails app with Spring and the app uses Sorbet, Sorbet cannot be disabled because the app has likely already invoked a typed method.

For now, this silences the error when disabling sorbet while using a spring booted server. It would be nice if we could look at disabling at at the right time in a safer way, but I have no idea how to do that yet reading spring. Maybe we need to boot our own version of a test mode spring server with typechecking disabled. See https://github.com/Shopify/packwerk/issues/327

## Type of Change

- [x] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist

### DIFF
--- a/exe/packwerk
+++ b/exe/packwerk
@@ -1,7 +1,10 @@
 #!/usr/bin/env ruby
 # frozen_string_literal: true
 
-require "packwerk/disable_sorbet"
+unless defined?(Spring)
+  require "packwerk/disable_sorbet"
+end
+
 require "packwerk"
 
 # Needs to be run in test environment in order to have test helper paths available in the autoload paths


shioyama (approved):

gmcgibbon:
I'll run a release if we agree this an acceptable measure to take in the interim.

PR #329, opened 2023-03-08
Use a single `OptionParser` instance to parse both possible options

davidstosik:

Currently, it is not possible to pass both `--packages` and `--offenses-formatter` options to a single call to the Packwerk CLI:

```
packwerk check --packages ./components/platform --offenses-formatter=default

/home/spin/.bundle/shopify/gems/packwerk-2.3.0/lib/packwerk/cli.rb:199:in `parse_run': invalid option: --offenses-formatter=plain (OptionParser::InvalidOption)
        from /home/spin/.bundle/shopify/gems/sorbet-runtime-0.5.10676/lib/types/private/methods/call_validation.rb:256:in `bind_call'
        from /home/spin/.bundle/shopify/gems/sorbet-runtime-0.5.10676/lib/types/private/methods/call_validation.rb:256:in `validate_call'
        from /home/spin/.bundle/shopify/gems/sorbet-runtime-0.5.10676/lib/types/private/methods/_methods.rb:275:in `block in _on_method_added'
        from /home/spin/.bundle/shopify/gems/packwerk-2.3.0/lib/packwerk/cli.rb:54:in `execute_command'
        from /home/spin/.bundle/shopify/gems/sorbet-runtime-0.5.10676/lib/types/private/methods/call_validation.rb:256:in `bind_call'
        from /home/spin/.bundle/shopify/gems/sorbet-runtime-0.5.10676/lib/types/private/methods/call_validation.rb:256:in `validate_call'
        from /home/spin/.bundle/shopify/gems/sorbet-runtime-0.5.10676/lib/types/private/methods/_methods.rb:275:in `block in _on_method_added'
        from /home/spin/.bundle/shopify/gems/packwerk-2.3.0/lib/packwerk/cli.rb:43:in `run'
```

This is because the first `OptionParser#parse!` will fail if it sees an option it does not know about.

https://github.com/Shopify/packwerk/blob/43ca76f57c5f73aa38a34f3f8d36607e3966edab/lib/packwerk/cli.rb#L197-L216

This PR is trying to fix that.

Parsing both options together in a single `OptionsParser` is, I think the way `OptionsParser` is meant to be used.

Not super happy with the stubbing in the test. Do you have suggestions on how we could improve that?

## Type of Change

- [x] Bugfix
- [ ] New feature

### Additional Release Notes

None.

## Checklist

### DIFF
--- a/lib/packwerk/cli.rb
+++ b/lib/packwerk/cli.rb
@@ -194,25 +194,21 @@ module Packwerk
       ignore_nested_packages = nil
       formatter = @offenses_formatter
 
-      if args.any? { |arg| arg.include?("--packages") }
-        OptionParser.new do |parser|
-          parser.on("--packages=PACKAGESLIST", Array, "package names, comma separated") do |p|
-            relative_file_paths = p
-          end
-        end.parse!(args)
-        ignore_nested_packages = true
-      else
+      OptionParser.new do |parser|
+        parser.on("--packages=PACKAGESLIST", Array, "package names, comma separated") do |p|
+          relative_file_paths = p
+          ignore_nested_packages = true
+        end
+
+        parser.on("--offenses-formatter=FORMATTER", String,
+          "identifier of offenses formatter to use") do |formatter_identifier|
+          formatter = OffensesFormatter.find(formatter_identifier)
+        end
+      end.parse!(args)
+
+      if relative_file_paths.empty?
         relative_file_paths = args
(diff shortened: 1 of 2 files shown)


shioyama (approved):
👍 much better

gmcgibbon (approved):
We should eventually move creation of the option parser to a separate file or method, but this is a good fix. Thank you!

Please consider squashing your commits before merging.

PR #334, opened 2023-03-14
Fix exception when running sig blocks

### LINKED ISSUE #333: [Bug Report] Exception when loading signatures
**Description**

Packwerk crashes on uninitialized constant `Parser::AST` when running running Sorbet sig blocks:

```
/Users/jbourassa/3.2.1/gems/packwerk-3.0.1/lib/packwerk/file_processor.rb:61:in `block in <class:FileProcessor>': uninitialized constant Parser::AST (NameError)

      params(node: Parser::AST::Node, relative_file: String).returns(T::Array[UnresolvedReference])
                              ^^^^^^
	from /Users/jbourassa/.gem/ruby/3.2.1/gems/sorbet-runtime-0.5.10712/lib/types/private/methods/_methods.rb:358:in `instance_exec'
	from /Users/jbourassa/.gem/ruby/3.2.1/gems/sorbet-runtime-0.5.10712/lib/types/private/methods/_methods.rb:358:in `run_builder'
	from /Users/jbourassa/.gem/ruby/3.2.1/gems/sorbet-runtime-0.5.10712/lib/types/private/methods/_methods.rb:336:in `run_sig'
	from /Users/jbourassa/.gem/ruby/3.2.1/gems/sorbet-runtime-0.5.10712/lib/types/private/methods/_methods.rb:246:in `block in _on_method_added'
	from /Users/jbourassa/.gem/ruby/3.2.1/gems/sorbet-runtime-0.5.10712/lib/types/private/methods/_methods.rb:447:in `run_sig_block_for_key'
	from /Users/jbourassa/.gem/ruby/3.2.1/gems/sorbet-runtime-0.5.10712/lib/types/private/methods/_methods.rb:464:in `block in run_all_sig_blocks'
	from /Users/jbourassa/.gem/ruby/3.2.1/gems/sorbet-runtime-0.5.10712/lib/types/private/methods/_methods.rb:461:in `loop'
	from /Users/jbourassa/.gem/ruby/3.2.1/gems/sorbet-runtime-0.5.10712/lib/types/private/methods/_methods.rb:461:in `run_all_sig_blocks'
	from /Users/jbourassa/.gem/ruby/3.2.1/gems/sorbet-runtime-0.5.10712/lib/types/utils.rb:83:in `run_all_sig_blocks'
	from /Users/jbourassa/src/minimal-repo/config/initializers/run_all_sig_blocks.rb:1:in `<top (required)>'
[...]
```

**To Reproduce**

```
# Setup the Rails app
rails new --minimal packwerk_repro
cd packwerk_repro
bundle add packwerk sorbet
echo 'T::Utils.run_all_sig_blocks'  > config/initializers/run_all_sig_blocks.rb

# Run it
bin/rails runner "1"
```

**Version Information**
 - Packwerk: v3+
 - Ruby v3.2.1

**Additional Context**

This is reproducible on a typical Rails app generated from Shopify's template (e.g. `dev init`), and we hit this in production only when deploying our service. I can provide a bugsnag link of that's helpful, hit me up on Slack.

jbourassa:
Fixes #333 by adding the missing require.

Also add an integration test so that future similar issues are caught.

## Type of Change

- [x] Bugfix
- [ ] New feature

## Checklist

### DIFF
--- a/lib/packwerk/file_processor.rb
+++ b/lib/packwerk/file_processor.rb
@@ -1,7 +1,7 @@
 # typed: strict
 # frozen_string_literal: true
 
-require "ast/node"
+require "parser/ast/node"
 
 module Packwerk
   class FileProcessor
--- a/test/loading/packwerk/autoload_test.rb
+++ b/test/loading/packwerk/autoload_test.rb
@@ -17,6 +17,15 @@ module Packwerk
       )
     end
 
+    test "running sig blocks does not load extra files" do
+      require "packwerk"
+      before_sig_files = loaded_packwerk_files
+
+      T::Utils.run_all_sig_blocks
+
+      assert_equal(before_sig_files, loaded_packwerk_files)
+    end
+
     private
 
     def loaded_packwerk_files


lib/packwerk/file_processor.rb
```diff
@@ -1,6 +1,7 @@
 # typed: strict
 # frozen_string_literal: true
 
+require "parser/ast/node"
```
gmcgibbon, line 4: I don't see usage of `AST::Node`. Can we just replace the other require in this file with this one?

test/integration/packwerk/signatures_load_test.rb
```diff
@@ -0,0 +1,13 @@
+# typed: true
+# frozen_string_literal: true
+
+require "test_helper"
+require "support/rails_test_helper"
+
+module Packwerk
+  class SignatureLoadTest < Minitest::Test
+    test "run all sig blocks without crashing" do
+      %x(bundle exec ruby -Ilib -rbundler/setup -rpackwerk -e "T::Utils.run_all_sig_blocks")
```
gmcgibbon, line 10: Instead, can we use `ActiveSupport::Testing::Isolation`?

jbourassa: TIL about `Testing::Isolation`, neat! I don't think it works here though because it uses forking, which already loaded the required code, unless I'm doing it wrong? Here's what I have:

```ruby
include ActiveSupport::Testing::Isolation`

test "run all sig blocks without crashing" do
  run do
    T::Utils.run_all_sig_blocks
  end
end
```

Commenting out the fix, this test still passes.

lib/packwerk/file_processor.rb
jbourassa: Makes sense, done!

gmcgibbon (approved):
Thanks! This is definitely a bug, but I'm very confused as to how this is happening since the files should be autoloaded. How does Sorbet know about the sig in this method? It must be loading early somehow.

gmcgibbon:
Ah, it is loaded by [another sig](https://github.com/Shopify/packwerk/blob/90a0a038083d6e3ace035197617daaf524ab22ff/lib/packwerk/run_context.rb#L96), which chain autoloads most of Packwerk. This might be a legitimate use-case for recommending `require: false` in the gem entry. If you find that loading all sigs slows down the application, I think that would help.

jbourassa:
> This might be a legitimate use-case for recommending `require: false` in the gem entry.

That sounds reasonable to me. I was _very_ surprised when a packwerk update broke our deploymeny 😅.

test/integration/packwerk/signatures_load_test.rb
jbourassa: Gotcha. Looking into this again though, `-rbundler/setup` and `bundle exec` should not be both required – only one. I'll fix that.

gmcgibbon:
I'll open an issue, I don't think this is correct behaviour on Packwerk's end. Thanks for reporting.

jbourassa:
> Can you add it to the [new test suite](https://github.com/Shopify/packwerk/tree/4e02bf1800ff8e653edb65ac1d47be72b0a27c96/test/loading/packwerk) before merging?

I tried something but not sure if this is exactly what you had in mind. The old test exercised running all the sigs, but this new one doesn't: the files haven't been loaded. It still maintains the idea though: running sigs do not crash, and do not require more packwerk files.

About merging: I assumed you wanted to wait before merging. I'd rather have you merge when / if you think this is fine, I wouldn't want to merge something you disagree with.

gmcgibbon (approved):

PR #336, opened 2023-03-15
Remove extraneous requires

### LINKED ISSUE #335: [Bug Report] Default require behaviour
**Description**
Requiring `packwerk` via `require "packwerk"` (the default require) sets autoloads and some upfront requires. Tools like `T::Utils.run_all_sig_blocks` triggers signatures to be evaluated, which expands Packwerk into memory as it loads constants in sigs.

**To Reproduce**
A scenario like https://github.com/Shopify/packwerk/pull/334 using `T::Utils.run_all_sig_blocks`.

**Expected Behaviour**
`T::Utils.run_all_sig_blocks` should not inflate Packwerk into memory.

**Screenshots**
N/A

**Version Information**
 - Packwerk: edge
 - Ruby 3.2.1

**Additional Context**
We don't seem to want to encourage developers reach for `require: false` as of #304. What we can do instead is remove the requires from the main entrypoint file and make sure no more are added (and thus no sigs evaluated). We don't seem to want to move autoloads out of the default require for API design reasons.

gmcgibbon:

Closes #335 

Remove extraneous requires from the main require file, and clean up early autoloads.

Lazily load checkers, validators, and offense formatters so that we don't need to add requires for them in the main file. I also moved the nested autoloads into nested files to make sure they don't load early.

This should make sure load times for Packwerk are minimal if the user chooses to add the gem without `require: false`.

## Type of Change

- [x] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist


test/integration/packwerk/load_test.rb
```diff
@@ -0,0 +1,34 @@
...
+      )
+    end
+
+    private
+
+    def loaded_packwerk_files
+      $LOADED_FEATURES.select { |path| path.starts_with?(GEM_LIB) }
+    end
+
+    def unload_packwerk
```
gmcgibbon, line 29: I'm not super happy about this hack. Is there a way to test in isolation without needing to undo what the test runner process has already done?

lib/packwerk.rb
```diff
@@ -101,26 +96,10 @@ module Checkers
...
 end
 
 require "packwerk/version"
-
-# Required to register the DefaultOffensesFormatter
-# We put this at the *end* of the file to specify all autoloads first
-require "packwerk/formatters/default_offenses_formatter"
-
-# Required to register the default DependencyChecker
-require "packwerk/reference_checking/checkers/dependency_checker"
```
alexevanczuk, line 124: Do the autoloads take care of this registering? I recall us adding these because without a reference to them in the codebase that will cause zeitwerk to require them, these don't get properly registered.

test/integration/packwerk/load_test.rb
gmcgibbon: Great idea! I'll give that a shot.

gmcgibbon: I initially had this in integration, but our integration tests are just interacting with Rails, and this seemed different, so I made a `test:loading` suite with a different helper file. Seems to work, and this would be a good place to put https://github.com/Shopify/packwerk/pull/334 too.

shioyama (approved):

PR #340, opened 2023-03-29
Implement a `count-offenses` command

davidstosik:

As we are planning to track the evolution of Packwerk over time, and to measure the impact of our work, having a Packwerk command that provides a count for offenses per package would be very useful. This PR implements it.

We had the choice to count offenses in the YAML file, or to find them all in the code and count that. Each option has its pros and cons:
- counting the values in `package_todo.yml` files is fast
- but files referenced in `package_todo.yml` get deduplicated, so the count is less accurate

We picked the second option, even though it is slower, in order to get a more accurate count and to be able to observe when the number of references grows or shrinks within a single file.

I needed to expose the `package_todos` attribute of the `OffenseCollection` class, in order to count offenses. I imagine I could instead define the method `OffenseCollection#count_package_todo` so that I don't have to expose `package_todo`. Is there a preferred approach?
(Note that I also renamed the `@package_todo` instance variable to make it plural, as it makes more sense.)

## Type of Change

- [ ] Bugfix
- [x] New feature

## Checklist

### DIFF
--- a/lib/packwerk/cli.rb
+++ b/lib/packwerk/cli.rb
@@ -57,6 +57,8 @@ module Packwerk
         output_result(parse_run(args).check)
       when "update-todo", "update"
         output_result(parse_run(args).update_todo)
+      when "show-offenses", "show"
+        output_result(parse_run(args).show_offenses)
       when "validate"
         validate(args)
       when "version"
@@ -118,6 +120,7 @@ module Packwerk
         Subcommands:
           init - set up packwerk
           check - run all checks
+          show-offenses - displays all offenses
           update-todo - update package_todo.yml files
           validate - verify integrity of packwerk and package configuration
           version - output packwerk version
--- a/lib/packwerk/parse_run.rb
+++ b/lib/packwerk/parse_run.rb
@@ -61,6 +61,16 @@ module Packwerk
     end
 
     sig { returns(Cli::Result) }
+    def show_offenses
+      run_context = RunContext.from_configuration(@configuration)
+      all_offenses = find_offenses(run_context)
+
+      message = @offenses_formatter.show_offenses(all_offenses)
+
+      Cli::Result.new(message: message, status: true)
+    end
+
+    sig { returns(Cli::Result) }
     def check
       run_context = RunContext.from_configuration(@configuration)
       offense_collection = OffenseCollection.new(@configuration.root_path)
@@ -93,8 +103,8 @@ module Packwerk
       ).returns(T::Array[Offense])
     end
     def find_offenses(run_context, &block)
-      offenses = T.let([], T::Array[Offense])
-      process_file = if block_given?
+      all_offenses = T.let([], T::Array[Offense])
+      process_file = if block
         T.let(proc do |relative_file|
           run_context.process_file(relative_file: relative_file).tap(&block)
         end, ProcessFileProc)
@@ -105,14 +115,14 @@ module Packwerk
       end
 
       @progress_formatter.started_inspection(@relative_file_set) do
-        offenses = if @configuration.parallel?
+        all_offenses = if @configuration.parallel?
           Parallel.flat_map(@relative_file_set, &process_file)
         else
           serial_find_offenses(&process_file)
         end
       end
 
-      offenses
+      all_offenses
     end
 
     sig { params(block: ProcessFileProc).returns(T::Array[Offense]) }


lib/packwerk/cli.rb
```diff
@@ -57,6 +57,8 @@ def execute_command(args)
         output_result(parse_run(args).check)
       when "update-todo", "update"
         output_result(parse_run(args).update_todo)
+      when "count-offenses"
```
gmcgibbon, line 60:
```suggestion
      when "count-offenses", "count"
```

USAGE.md
```diff
@@ -240,6 +240,15 @@ Above is an example of a constant violation entry in `package_todo.yml`.
 
 Violations exist within the package that makes a violating reference.
 
+
+# Counting Offenses
+
+You can generate a count of all offenses by running:
+
+    bin/packwerk count-offenses
```
gmcgibbon, line 248: We should use code blocks like
```
bin/packwerk count-offenses
```
similar to other examples in the docs

It would also be good to add an example of the output.

test/unit/packwerk/parse_run_test.rb
```diff
@@ -501,5 +501,23 @@ def something
...
+      )
+      RunContext.any_instance.stubs(:process_file).returns([offense])
+
+      parse_run = ParseRun.new(
+        relative_file_set: Set.new(["path/of/exile.rb"]),
+        configuration: Configuration.from_path
+      )
+      result = parse_run.count_offenses
+
+      assert_equal "components/source,1\n", result.message
```
gmcgibbon, line 520: I think it would be useful to test a setup with multiple packages (3?).

lib/packwerk/parse_run.rb
```diff
@@ -57,6 +58,20 @@ def update_todo
...
+      run_context = RunContext.from_configuration(@configuration)
+      offense_collection = find_offenses(run_context)
+
+      message = CSV.generate do |csv|
+        offense_collection.package_todos.each do |package, package_todo|
+          csv << [package.name, package_todo.count]
+        end
+      end
+
+      Cli::Result.new(message: message, status: true)
```
gmcgibbon, line 72: I think we talked about this a little, but do you think we could achieve the same thing with an offense formatter? Also do you think CSV is a readable default for this command? Maybe we should print them out by default, and then offer CSV as a flag.

USAGE.md
```diff
@@ -20,7 +20,6 @@
   * [Understanding how to respond to new violations](#understanding-how-to-respond-to-new-violations)
 * [Recording existing violations](#recording-existing-violations)
   * [Understanding the package todo file](#understanding-the-package-todo-file)
-  * [Understanding the list of deprecated references](#understanding-the-list-of-deprecated-references)
```
davidstosik, line 23: This points to a non-existent entry in the doc.

lib/packwerk/parse_run.rb
davidstosik: Somehow, GitHub managed to get this comment to point to something in the PR, but please note that the method is not `#count_offenses` anymore, but `#show_offenses`. It uses an offense formatter.

lib/packwerk/parse_run.rb
```diff
@@ -76,16 +94,25 @@ def check
...
-      end, ProcessFileProc)
+      process_file = if block
+        T.let(proc do |relative_file|
+          run_context.process_file(relative_file: relative_file).tap(&block)
+        end, ProcessFileProc)
+      else
+        T.let(proc do |relative_file|
+          run_context.process_file(relative_file: relative_file)
+        end, ProcessFileProc)
+      end
```
gmcgibbon, line 115: Can we move this to a private method? This is slightly repetitive, but I think it is justified with how many times this gets  called.

lib/packwerk/cli.rb
```diff
@@ -57,6 +57,8 @@ def execute_command(args)
         output_result(parse_run(args).check)
       when "update-todo", "update"
         output_result(parse_run(args).update_todo)
+      when "show-offenses", "show"
```
gmcgibbon, line 60: I wonder if violations or offense is the right wording here. We should be more consistent with out wording internally. Since we've already gotten things like "offense formatters", this is probably fine.

gmcgibbon (commented):
This is looking great! Just one comment about `process_file` creation.

gmcgibbon:
I'm really curious how this affects the performance of check and update. Can you provide some benchmarks if you have time?

lib/packwerk/parse_run.rb
davidstosik: Do you mean something like this?

```rb
def process_file_proc(&block)
  if block
    T.let(proc do |relative_file|
      run_context.process_file(relative_file: relative_file).tap(&block)
    end, ProcessFileProc)
  else
    T.let(proc do |relative_file|
      run_context.process_file(relative_file: relative_file)
    end, ProcessFileProc)
  end
end

# Then I can call it where it's used:
# ...
      @progress_formatter.started_inspection(@relative_file_set) do
        all_offenses = if @configuration.parallel?
          Parallel.flat_map(@relative_file_set, &process_file_proc(&block))
        else
          serial_find_offenses(&process_file_proc(&block))
        end
      end
# ...
# (Or I can keep a local variable if that seems better.)
```

davidstosik:
@gmcgibbon A very quick time check on the Shopify core code base does not seem to show noticeable changes (either good or bad):

```
spring stop && time bin/packwerk ${command}

Spring stopped.
...
📦 Finished in XXX
...
bin/packwerk ${command} ___s user ___s system __% cpu YYY total
```

XXX and YYY times below:

command | `main` | `sto/count-violations`
---|---|---
`check` | 16.12s / 45.424s | 15.76s / 45.336s
`update-todo` | 16.29s / 43.725s | 15.83s / 42.382s

Should I try to properly use benchmark with the gem's fixtures?

lib/packwerk/cli.rb
alexevanczuk: Big plus for this! We call these three things:
- todos (e.g. `update-todo`, `package_todo.yml`
- violations
- offenses

IMO violations and offenses are somewhat aggressive terms and TODO captures the desire to change/fix while being less aggressive and also shorter to write!

I'd be in support of changing all language across the board to `todos`! Some things could be changed right away (e.g. internal implementation variable names), others could probably be changed while not being considered a public API breakage (e.g. output of commands), others would need to be deprecated or have a switchover before a major version change (e.g. CLI commands, offenses formatter, the word violation in package_todo files).

What do you all think? (We could start a discussion too.)

davidstosik: Not sure I agree with using the word TODO.
To me, the word TODO in Packwerk has the same meaning as in Rubocop's `.rubocop_todo.yml`, which is there to disable _offenses_ "to be fixed later":

> The generated file `.rubocop_todo.yml` contains configuration to disable cops that currently detect an offense in the code by changing the configuration for the cop, excluding the offending files, or disabling the cop altogether once a file count limit has been reached. ([source](https://docs.rubocop.org/rubocop/configuration.html#:~:text=The%20generated%20file%20.rubocop_todo.yml%20contains%20configuration%20to%20disable%20cops%20that%20currently%20detect%20an%20offense%20in%20the%20code%20by%20changing%20the%20configuration%20for%20the%20cop%2C%20excluding%20the%20offending%20files%2C%20or%20disabling%20the%20cop%20altogether%20once%20a%20file%20count%20limit%20has%20been%20reached.))

I think we could use the same word for _violations_ and _offenses_, but would not conflate the concept with _todo_.
_Offense_ feels a bit more neutral to me (and I guess it also matches Rubocop's vocabulary). The word is also already used more than twice more than _violation_.

```
ag -i offense | wc -l
     799
ag -i violation | wc -l
     325
```

I have a hunch that there might be a will to give _violation_ and _offense_ slightly different meanings, but 1. I'm not sure I'm right and 2. this might not be necessary, or could be expressed more explicitly:
- _offense_: anything that's causing an error (config error, reference that violates declared dependencies, etc)
- _violation_: an _offense_ that is not recorded as a _todo_ (and that triggers an error on `check`).

alexevanczuk: Interestingly, even though offense is used more often in the source code, violation (at least at Gusto) is used much more often by the client. Not sure if it's the same at Shopify, but here everyone uses the word violation, and it's probably in part because the word `violation` is in `package_todo.yml` files explicitly. Also, the output of runs uses the word violation (e.g. "no stale violations detected"), although the word offenses is used too sometimes.

I can buy the different use cases for the words. Something is an offense for example in packwerk if there's a file that can't be parsed correctly. A violation is an offense, but is also what happens when everything works correctly in packwerk, but there's a "violation" of packwerk directives (e.g. dependency usage, private API usage, etc.). TODOs are recorded violations, perhaps short for "violations TODO."

I'm not sure if all of this distinction is useful to the client using this system though, but also I haven't really heard of many folks confused about these terms besides us so perhaps it's not a problem! I'm just glad we replaced `deprecated_references` with `todo` though – I've found that has resolved a lot of language confusion on our end.

davidstosik:
Hello @rafaelfranca, sorry I forgot to reply to your comment. 🙇🏻‍♂️ 
We're taking a different route trying to use output formatters instead, but for that I've been working on a [refactor](https://github.com/Shopify/packwerk/pull/351) of Packwerk's CLI code.

Setting this to draft for now.

Issue #299, opened 2023-02-02
[Bug Report]


huguesbr:
**Description**
When running `bin/packwerk check` I got the following error

```
bin/packwerk check
[TEST PROF INFO] Spring detected
📦 Packwerk is inspecting 4433 files
...........................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................Traceback (most recent call last):
	140: from bin/packwerk:27:in `<main>'
	139: from bin/packwerk:27:in `load'
	138: from /Users/hugues/.rbenv/versions/2.7.7/lib/ruby/gems/2.7.0/gems/packwerk-2.2.2/exe/packwerk:12:in `<top (required)>'
	137: from /Users/hugues/.rbenv/versions/2.7.7/lib/ruby/gems/2.7.0/gems/sorbet-runtime-0.5.10647/lib/types/private/methods/_methods.rb:275:in `block in _on_method_added'
	136: from /Users/hugues/.rbenv/versions/2.7.7/lib/ruby/gems/2.7.0/gems/sorbet-runtime-0.5.10647/lib/types/private/methods/call_validation.rb:256:in `validate_call'
	135: from /Users/hugues/.rbenv/versions/2.7.7/lib/ruby/gems/2.7.0/gems/sorbet-runtime-0.5.10647/lib/types/private/methods/call_validation.rb:256:in `bind_call'
	134: from /Users/hugues/.rbenv/versions/2.7.7/lib/ruby/gems/2.7.0/gems/packwerk-2.2.2/lib/packwerk/cli.rb:43:in `run'
	133: from /Users/hugues/.rbenv/versions/2.7.7/lib/ruby/gems/2.7.0/gems/sorbet-runtime-0.5.10647/lib/types/private/methods/_methods.rb:275:in `block in _on_method_added'
	132: from /Users/hugues/.rbenv/versions/2.7.7/lib/ruby/gems/2.7.0/gems/sorbet-runtime-0.5.10647/lib/types/private/methods/call_validation.rb:256:in `validate_call'
[...]

PR #343, opened 2023-04-18
Rescue StandardError when processing file

alexevanczuk:
This should help https://github.com/Shopify/packwerk/issues/299.

While it doesn't solve the parsing error, it should improve debug reports by printing out which files are erroring with the error message.

Right now, when there is a `NodeHelpers::TypeError` OR a Sorbet `TypeError` (the test case I added reproduces this by creating a false positive AR association), Packwerk just crashes with a messy message:

```
$ bin/packwerk check
Running via Spring preloader in process 8043
📦 Packwerk is inspecting 100 files
......................................................................................................................................................................................................................................................................................./Users/alex.evanczuk/.rbenv/versions/3.1.4/lib/ruby/gems/3.1.0/gems/sorbet-runtime-0.5.10773/lib/types/_types.rb:218:in `must': Passed `nil` into T.must (TypeError)
        from /Users/alex.evanczuk/.rbenv/versions/3.1.4/lib/ruby/gems/3.1.0/bundler/gems/packwerk-816c080d296e/lib/packwerk/association_inspector.rb:67:in `association_name'
        from /Users/alex.evanczuk/.rbenv/versions/3.1.4/lib/ruby/gems/3.1.0/gems/sorbet-runtime-0.5.10773/lib/types/private/methods/call_validation_2_7.rb:703:in `bind_call'
        from /Users/alex.evanczuk/.rbenv/versions/3.1.4/lib/ruby/gems/3.1.0/gems/sorbet-runtime-0.5.10773/lib/types/private/methods/call_validation_2_7.rb:703:in `block in create_validator_method_medium1'
        from /Users/alex.evanczuk/.rbenv/versions/3.1.4/lib/ruby/gems/3.1.0/bundler/gems/packwerk-816c080d296e/lib/packwerk/association_inspector.rb:38:in `constant_name_from_node'
        from /Users/alex.evanczuk/.rbenv/versions/3.1.4/lib/ruby/gems/3.1.0/gems/sorbet-runtime-0.5.10773/lib/types/private/methods/call_validation.rb:153:in `bind_call'
        from /Users/alex.evanczuk/.rbenv/versions/3.1.4/lib/ruby/gems/3.1.0/gems/sorbet-runtime-0.5.10773/lib/types/private/methods/call_validation.rb:153:in `validate_call_skip_block_type'
        from /Users/alex.evanczuk/.rbenv/versions/3.1.4/lib/ruby/gems/3.1.0/gems/sorbet-runtime-0.5.10773/lib/types/private/methods/call_validation.
```

It's not clear what file is causing `packwerk` to break. It might be better to list out the files that have issues to exclude them, fix the issues, or better report the issue to packwerk maintainers.

I decided to rescue all of `StandardError`, because there are a number of possibly unexpected things that can cause packwerk parsing to break. In theory we could target more specific errors, but then we wouldn't be catching all possible errors and ambiguous errors continue to be reported ambiguously.

I wasn't able to reproduce the specific case in the issue – any thoughts on what is going on there?
Let me know if I should approach this a different way.

## Type of Change

- [x] Bugfix
- [ ] New feature

## Checklist

### DIFF
--- a/lib/packwerk/file_processor.rb
+++ b/lib/packwerk/file_processor.rb
@@ -53,6 +53,17 @@ module Packwerk
       ProcessedFile.new(unresolved_references: unresolved_references)
     rescue Parsers::ParseError => e
       ProcessedFile.new(offenses: [e.result])
+    rescue StandardError => e
+      message = <<~MSG
+        Packwerk encountered an internal error.
+        For now, you can add this file to `packwerk.yml` `exclude` list.
+        Please file an issue and include this error message and stacktrace:
+
+        #{e.message} #{e.backtrace}"
+      MSG
+
+      offense = Parsers::ParseResult.new(file: relative_file, message: message)
+      ProcessedFile.new(offenses: [offense])
     end
 
     private
--- a/test/integration/packwerk/custom_executable_integration_test.rb
+++ b/test/integration/packwerk/custom_executable_integration_test.rb
@@ -128,6 +128,29 @@ module Packwerk
       end
     end
 
+    test "'packwerk check' does not blow up when parsing files with syntax issues from a false positive association" do
+      open_app_file(TIMELINE_PATH.join("app", "models", "bad_file.rb")) do |file|
+        # This is an example of a file that has an object called `belongs_to` that accepts methods
+        content = <<~CONTENT
+          belongs_to.some_method
+        CONTENT
+
+        file.write(content)
+        file.flush
+
+        refute_successful_run("check")
+
+        assert_match(/Packwerk is inspecting 13 files/, captured_output)
+        assert_match(%r{components/timeline/app/models/bad_file.rb}, captured_output)
+        assert_match(/Packwerk encountered an internal error/, captured_output)
+        assert_match(/For now, you can add this file to `packwerk.yml` `exclude` list./, captured_output)
+        assert_match(/Please file an issue and include this error message and stacktrace:/, captured_output)
+        assert_match(/Passed `nil` into T.must/, captured_output)
+        assert_match(/1 offense detected/, captured_output)
+        assert_match(/No stale violations detected/, captured_output)
+      end
+    end
+
     private
 
     def assert_successful_run(command)
--- a/test/unit/packwerk/file_processor_test.rb
+++ b/test/unit/packwerk/file_processor_test.rb
@@ -109,9 +109,14 @@ module Packwerk
         )
       )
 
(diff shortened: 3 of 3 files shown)


test/integration/packwerk/custom_executable_integration_test.rb
```diff
@@ -128,6 +128,33 @@ class CustomExecutableIntegrationTest < Minitest::Test
...
+        CONTENT
+
+        file.write(content)
+        file.flush
+
+        refute_successful_run("check")
+
+        assert_match(/Packwerk is inspecting 13 files/, captured_output)
+        assert_match(%r{components/timeline/app/models/bad_file.rb}, captured_output)
+        assert_match(/Passed `nil` into T.must/, captured_output)
```
gmcgibbon, line 152: Should we be making this code more fault tolerant? If this stack trace is within Packwerk, I imagine the end user should not be concerned with this and it is actually a bug.

lib/packwerk/file_processor.rb
```diff
@@ -53,6 +53,9 @@ def call(relative_file)
       ProcessedFile.new(unresolved_references: unresolved_references)
     rescue Parsers::ParseError => e
       ProcessedFile.new(offenses: [e.result])
+    rescue StandardError => e
```
rafaelfranca, line 56: What exactly is raising the error? The way this code is, if I write bad code in packwerk it is our users that will see the error, not us.

test/integration/packwerk/custom_executable_integration_test.rb
```diff
@@ -128,6 +128,33 @@ class CustomExecutableIntegrationTest < Minitest::Test
...
+    test "'packwerk check' does not blow up when parsing files with syntax issues from a false positive association" do
+      open_app_file(TIMELINE_PATH.join("app", "models", "bad_file.rb")) do |file|
+        # This is an example of a file that references `belongs_to`, but as an ActiveAdmin API
+        # rather than as a Rails Association concept
+        content = <<~CONTENT
+          require 'rails_helper'
+
+          RSpec.describe ActiveAdmin::Resource::BelongsTo do
+            it "should have an owner" do
+              expect(belongs_to.owner).to eq post_config
```
rafaelfranca, line 140: I'm not sure I get this example. It is failing because `belongs_to` doesn't exist? Is packwerk executing the Ruby code instead of just parsing it?

rafaelfranca (approved):

PR #345, opened 2023-04-19
Use Packs to find load paths instead of loading Rails

alexevanczuk:
# Summary
This PR uses `packs` to find the load paths for apps that use `packs-rails`. This allows folks to use packwerk to gain information about a Rails app without booting Rails, which is especially helpful for analyzing apps that are not yet setup.

# Commits
- use packs
- bundle lock --add-platform ruby
- bin/tapioca gem packs
- Do not use Rails – use autoload paths from Packs instead


alexevanczuk:
Sorry, I meant to open this as a PR against a fork!

PR #349, opened 2023-04-19
Add command line option to toggle parallel execution

meganemura:

Allows parallel configuration to be overridden by command line options.
This is useful if you want to switch parallel options between different execution environments, such as local and CI environments.

I added the --parallel / --no-parallel option to the process of parsing command-line options by OptionParse.

- Changes adding an attr_ accessor to `@parallel` in the Configuration class.
- ~~That the test is against private methods.~~ (fixed)

## Type of Change

- [ ] Bugfix
- [x] New feature

## Checklist

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -188,7 +188,7 @@ You can also specify packages for a shorter run time. When checking against pack
 
     bin/packwerk check --packages=components/your_package,components/your_other_package
 
-You cal also specify enable or disable parallel processing by a command line option that overrides the configuration file.
+Using the following command line option you can also enable or disable parallel processing. It is enabled by default.
 
     bin/packwerk check --[no-]-parallel
 
--- a/USAGE.md
+++ b/USAGE.md
@@ -188,6 +188,10 @@ You can also specify packages for a shorter run time. When checking against pack
 
     bin/packwerk check --packages=components/your_package,components/your_other_package
 
+You cal also specify enable or disable parallel processing by a command line option that overrides the configuration file.
+
+    bin/packwerk check --[no-]-parallel
+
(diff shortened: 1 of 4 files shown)


mclark (commented):
This change makes sense to me. Just a minor wording suggestion. 👍

USAGE.md
```diff
@@ -188,6 +188,10 @@ You can also specify packages for a shorter run time. When checking against pack
 
     bin/packwerk check --packages=components/your_package,components/your_other_package
 
+You cal also specify enable or disable parallel processing by a command line option that overrides the configuration file.
```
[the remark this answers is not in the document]
meganemura: Thank you. I appreciate it.

test/unit/packwerk/cli_test.rb
```diff
@@ -296,5 +296,17 @@ def show_strict_mode_violations(_offenses)
 
       cli.execute_command(["check", "--offenses-formatter=default", "--packages=components/platform"])
     end
+
+    test "#parse_run (private) parses parallel option and overrides the configuration" do
```
rafaelfranca, line 300: Can we avoid calling the private method by calling the public interface and asserting the configuration object that we are already passing changed its value?

rafaelfranca (changes requested):

rafaelfranca (approved):

rafaelfranca (approved):

test/unit/packwerk/cli_test.rb
davidstosik: Thank you for writing tests against the public interface! That means I can refactor without needing to change tests! 🙇🏻‍♂️

PR #350, opened 2023-04-19
Add 'ignored_dependencies' option for package.yml files to incrementally adopt dependency protection

alexevanczuk:
More info here: https://github.com/Shopify/packwerk/discussions/313#discussioncomment-5606593

TLDR: Someone can now turn on `enforce_dependencies` but focus on specific dependencies they want to enforce, while ignoring others. This may be useful if some, but not all, dependencies need to be considered thoughtfully for a package, but others do not matter. The user can then ignore them (explicitly) to help focus on specific dependencies.

I chose to add an `ignored_dependencies` key as a separate top-level package key.

Any reason why we'd not want this behavior?

## Type of Change

- [ ] Bugfix
- [x] New feature

## Checklist


USAGE.md
```diff
@@ -163,6 +163,19 @@ dependencies:
...
 It will be a dependency violation when `components/shop_identity` tries to reference a constant that is not within `components/platform` or itself.
 
+If you'd like to enforce some, but not all, dependencies, you can use the `dependencies` key. For example, if we want to enforce dependencies from `components/shop_identity` onto `components/platform`, but not onto `components/admin`, we can do this:
+
+```yaml
+# components/shop_identity/package.yml
+enforce_dependencies: true
+dependencies:
+  - components/platform
+ignored_dependencies:
```
rafaelfranca, line 173: I feel like `temporary_dependencies` dependencies would be a better name. Ignore seems like it is on to depend on this code and nothing needs to be done in the future, when in reality you want those to be fixed, but you can't right now.

```suggestion
temporary_dependencies:
```

rafaelfranca (commented):
I'm on the fence on this one

 What this feature is introducing, in my opinion is what the package_todo file is for. You want those "ignored" dependencies to be fixed some day, you just can't right now. So you should be adding to the todo.

But I totally get why you don't want to use the todo for this. People will just learn the behavior of adding everything to the todo list. They will likely add everything to the list of "ignored" dependencies as well.

So it is really hard to me to see a justification of why we can't use the todo file to record all the dependencies we can't fix right now, but we should be fixing.

If we move with this, I believe we should record the entire package in the todo as well like this:

```
# components/shop_identity/package_todo.yml
components/admin:
  disabled: true
```

This will make explicit dependencies to this package still need to fixed to whoever is reading the todo file to figure out what is waiting to be solved.

rafaelfranca:
> 1. Not all violations to `components/admin` are disabled – only dependency. Do you have a preference for how that's displayed? Maybe `disabled` is a list of the violation types, e.g. `disabled: ['dependency']`. However, that doesn't really map to other checkers. While `ignored_dependencies` is specific to dependency and not known about by other checkers, I'm not sure what interfaces makes sense for the `package_todo.yml` to be applicable generically for checkers, and we probably should make it a `Checker` Interface method if we want it in the `package_todo.yml`.

Good point. I think we should split violation by checker like we do on Rubocop. Instead of having the file format being:

```
component:
  class_name:
    checker:
```

Have it be:

```
checker:
  component:
    class_name:
```

> 2\. It feels redundant to repeat ourselves (and with different language) in the `package_todo.yml` if it's already listed as ignored in the main `package.yml`.

Yeah, I know it is redundant, but the todo file is what should be making things that need to be fixed. If the temporary dependencies aren't in that file, people will think it is ok to keep them forever without fixing it.

> 3\. If the dependency is ignored, it seems like what is being communicated is that there isn't actually anything to do until it is no longer ignored.

But that is the problem in my view. There is something to do, remove it from the list of ignored dependencies. After you do that, you will get other things to do, but I don't think it makes sense for the design of packwerk to allow dependencies to continue not being defined. If we do that we are rendering the tool unnecessary.

> 4\. In some other discussions/PRs you decided to not invest in allowing checkers to customize how they show up in `package_todo.yml` files, preferring instead to invest in a different/modular database implementation. Should that apply here too?

That discussion is about not allowing where the todo database show up depending on the check. About the format of the todo file, we can always discuss how to better allow checkers to describe what needed to be done.

> 5\. This is a breaking change to `package_todo.yml` file formats, so any tool or workflow that depends on them in their current format would no longer work, so I'd prefer to allow the value add portion (new feature) to show up in a minor release and any breaking changes in a major release.

That is an interesting point. We never made the package_todo file public API. Tools and workflows that depend on them are doing on their own risk. I warned this many time in the past to your team when you decided to build tools based on the current format. I'm totally fine to change an internal format of packwerk any time we want, that is why we never made it public in the first place. I'm totally open to a conversation where we end up with a stable format that we can make public, but the current one isn't it.

So I don't see this as a breaking change. Nothing in the packwerk public API is breaking. All commands will continue to work.

USAGE.md
shageman: Would `currently_ignored_dependencies` satisfy both your viewpoints? (And be a sufficiently obvious awkward and long name that folks get the message that they shouldn't want this?)

mclark: > Another possibility, if that we also record circular dependencies in the todo, instead of introducing a way to ignore dependencies.

🤔 This is a very compelling idea. `package_todo.yml` already tracks dependency violations, why not circular dependency violations as well? It could get noisy but surfacing circular dependencies in the PR diff could be very illuminating and force teams to acknowledge this risk/tech debt they are taking on by having them.

@alexevanczuk would this address your needs for ignored dependencies?

rafaelfranca:
Not sure what to make of the last comment. If you got what I mean, and you are "ok with either", what is the goal of the comment?

shageman:
> Not sure what to make of the last comment. If you got what I mean, and you are "ok with either", what is the goal of the comment?

To try and change your mind. You can say that it is not part of the public API, but that is not as important as realizing and acting with the knowledge that it is part of the public interface.

USAGE.md
alexevanczuk: It's pretty close, because it allows us to ignore dependencies by listing them once. I still really like the communication vehicle of it being "ignored" because it allows folks to express "these may or may not be our long-term dependencies: we don't want to consider them for violations" as a separate expressive statement compared to "these are our true dependencies." Having them combined lets folks ignore them more easily by we'd still end up with the conceptual issue of `dependencies` being overloaded.

gmcgibbon:
I'm also still slightly hesitant about this change. Here are some of my thoughts:

- The `ignored_dependencies` feature is encoding tech debt that is intended to be worked on later into the `package.yml`, which seems wrong to me. As Rafael mentions, this information should be recorded in the TODO file.
- The TODO fatigue is a legitimate concern, but it is the right place to record this information. I don't think it is clear how things needs to change to accommodate this concern yet.
- Regarding the TODO file being public API or not, I'm also of the opinion that it is not. While it can definitely be argued that it is, as maintainers, we need to draw the line somewhere, and this simply seems like a difference of opinion.

alexevanczuk:
Thank you everyone for the lively and passionate discussion! Since there are a lot of objections to this approach, I'll close this PR out now as we think more about it and explore changing the `.package_todo.yml` format, which will require a handful of structural changes first.

I'm fine with the idea of updating the TODO list format so each checker is fully responsible for recording its own TODOs. As discussed, this might also help resolve other issues & differences of opinion around where checkers record violations. It doesn't present much risk to our other tooling since none of our tooling reads the YML files directly except for a single library (`parse_packwerk`) which can easily be updated.

I opened up a new discussion to discuss cycles as a TODO item, which I find compelling after thinking about it more:
https://github.com/Shopify/packwerk/discussions/353

exterm:
I'm chiming in really late on this, but maybe my perspective can still be helpful.

I think there was a great and productive discussion above!

For context - early prototypes of `packwerk` did include two kinds of dependencies in `package.yml` files, for exactly the reason that Alex is proposing this here. We dropped the "feature" because it seemed confusing and we hadn't figured out the design and implications (this was really early in the process).

I agree the todo file was never intended to be public API and warned about it too, but on the other hand I can understand why people started building upon it. In this case though, it seems like a change to the format is warranted.

I like the idea of having `ignored_dependencies` and then represent them in the todo file. And I do think this is a much better fit with packwerk's existing design than what's proposed in https://github.com/Shopify/packwerk/discussions/353.

PR #351, opened 2023-04-20
Extract CLI commands to individual classes

davidstosik:

Packwerk's code around the CLI is currently hard to understand, mixes concepts together, and is overall very difficult to change. (Changing a seemingly simple thing in the CLI would sometimes requires editing many files).

This PR attempts a refactor of the code CLI in order to make it

1. easier to understand
2. easier to modify

The first approach that we tried with @gmcgibbon involved using Thor, but we realized that using Thor while not breaking public APIs would be overly complicated, and end up trading one type of complexity for another.

In this PR, my approach was mainly to:
1. separate each CLI command to its own class
2. avoid introducing behaviour changes in public APIs (the CLI should behave the same, return same results and output same text, public classes should keep the same public methods with the same signature and expected behaviour)
3. make the CLI open-closed, allowing an external actor (such as a plugin?) to declare new CLI commands without having to monkey-patch any code in Packwerk.

I know this looks rather big, but to this PR's defense, about 500 lines additions/deletions are test code lines that moved to a different file.
I left a few comments which you can check below about the things I'm unsure and would like a second opinion about.

If the approach sounds useful to you, you can also check the changes commit by commit, as I have made each commit a logical and justified change, ensuring that each commit also keeps the test suite green. 

## Type of Change

- [ ] Bugfix
- [ ] New feature

## Checklist


lib/packwerk.rb
```diff
@@ -39,10 +40,6 @@ module OutputStyles
     autoload :Plain
   end
 
-  autoload_under "commands" do
-    autoload :OffenseProgressMarker
-  end
-
```
davidstosik, line 45: This does not seem necessary, and was actually interfering with the `autoload :Commands` that we added above.

lib/packwerk/cli.rb
```diff
@@ -11,6 +9,20 @@ class Cli
...
     autoload :Result
 
+    class << self
+      extend T::Sig
+
+      sig { params(command: String).returns(T.nilable(T.class_of(Commands::BaseCommand))) }
+      def command_class_for(command)
+        command_class_name = command.sub(" ", "_").underscore.classify + "Command"
+        if Commands.const_defined?(command_class_name)
+          Commands.const_get(command_class_name) # rubocop:disable Sorbet/ConstantsFromStrings
```
davidstosik: I tried with the registry pattern first, but it required me to `require` all the files (with `Dir[]` and a glob for example) preemptively.
Using `const_get`, we can keep lazy loading.

lib/packwerk/commands/uses_parse_run.rb
davidstosik, line 1: This module is a temporary to keep the behaviour as it was previously, focusing this PR on a refactor that mostly does not change behaviour.

However, the respective behaviours of `check` and `update-todo` are expected to diverge:
- `--offenses-formatter` does not make much sense in the case of `update-todo`
- `update-todo` does not accept a list of files (either passed as extra arguments or with `--packages`

When we work on this separation of behaviours, this module will naturally go away.

lib/packwerk/cli/result.rb
davidstosik, line 1: I removed this because it didn't seem to make sense anymore:
- it's never returned by methods that are part of the public API
- it imposes the commands to return their last message wrapped in a `Result` object, instead of outputting it in the relevant IO object that they know about, and to which they keep outputting during their execution

lib/packwerk/parse_run.rb
davidstosik, line 1: I'm tempted to rename this file `offenses_finder.rb` now. 🤔

lib/packwerk/commands/check_command.rb
```diff
@@ -0,0 +1,48 @@
...
+
+      sig { override.returns(T::Boolean) }
+      def run
+        if @files_for_processing.files.empty?
+          @out.puts(<<~MSG.squish)
+            No files found or given.
+            Specify files or check the include and exclude glob in the config file.
+          MSG
+
+          true
```
davidstosik, line 18: It feels like it should return `false` if the above was an error, but this is the existing behaviour.

Also, we pass an `err_out` around but it feels like we're not using it as much as we should...

In the spirit of keeping this PR a low impact refactor, I decided not to change those.

lib/packwerk/commands/help_command.rb
```diff
@@ -0,0 +1,27 @@
...
+# frozen_string_literal: true
+
+module Packwerk
+  module Commands
+    class HelpCommand < BaseCommand
+      extend T::Sig
+
+      sig { override.returns(T::Boolean) }
+      def run
+        @err_out.puts(<<~USAGE)
```
davidstosik, line 11: Here too, I'm preserving existing behaviour: this help message was output to `err_out` before, still is.

lib/packwerk.rb
shioyama: 👍 

Seems it was left behind in https://github.com/Shopify/packwerk/pull/124

lib/packwerk/cli.rb
```diff
@@ -49,181 +45,23 @@ def run(args)
...
-        output_result(parse_run(args).update_todo)
-      when "validate"
-        validate(args)
-      when "version"
-        @out.puts(Packwerk::VERSION)
-        true
-      when nil, "help"
-        usage
+      command = args.shift || "help"
+      command_class = Commands.class_for(command)
```
shioyama, line 49: You can avoid the need for `class_for` by just `const_get`-ting the command (converted to class name), and rescuing the possible `NameError`, like this:

```ruby
      class_name = command.sub(" ", "_").underscore.classify + "Command"
      Commands.const_get(class_name).new(
        args,
        configuration: @configuration,
        out: @out,
        err_out: @err_out,
        progress_formatter: @progress_formatter,
        offenses_formatter: @offenses_formatter,
      ).run
    rescue NameError
      @err_out.puts("'#{command}' is not a packwerk command. See `packwerk help`.",)

      false
    end
```

lib/packwerk/commands.rb
```diff
@@ -0,0 +1,30 @@
...
+    autoload :VersionCommand
+
+    class << self
+      extend T::Sig
+
+      sig { params(command: String).returns(T.nilable(T.class_of(Commands::BaseCommand))) }
+      def class_for(command)
+        class_name = command.sub(" ", "_").underscore.classify + "Command"
+        if Commands.const_defined?(class_name)
+          Commands.const_get(class_name) # rubocop:disable Sorbet/ConstantsFromStrings
```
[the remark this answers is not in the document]
shioyama: I think `const_get` is ok, but we don't need this `class_for` method, just lean on autoloading to do this work for you. See my other comment. I ran tests and they all pass with the proposed change.

lib/packwerk/commands/help_command.rb
```diff
@@ -0,0 +1,25 @@
...
+        Cli::Result.new(status: true, print_as_error: true, message: <<~USAGE)
+          Usage: #{$PROGRAM_NAME} <subcommand>
+
+          Subcommands:
+            init - set up packwerk
+            check - run all checks
+            update-todo - update package_todo.yml files
+            validate - verify integrity of packwerk and package configuration
+            version - output packwerk version
+            help  - display help information about packwerk
```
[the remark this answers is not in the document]
shioyama: Maybe we could use the registry pattern just for help messages? So have some registry that commands can call to register their help message, but let the actual command class be autoloaded.

lib/packwerk/commands.rb
```diff
@@ -0,0 +1,30 @@
...
+
+module Packwerk
+  module Commands
+    extend ActiveSupport::Autoload
+
+    autoload :BaseCommand
+    autoload :CheckCommand
+    autoload :HelpCommand
+    autoload :InitCommand
+    autoload :UpdateCommand, "packwerk/commands/update_todo_command"
```
shioyama, line 12: Maybe we should add some aliasing functionality to make this more natural and explicit? i.e. here we're aliasing `update` to `update_todo`, so rather than adding an autoload we could have `execute_command` look at a list of aliases and "redirect" the command accordingly if an alias exists.

Or do we also need to support the existence of the `UpdateComand` constant as well? 🤔

shioyama (commented):
Still reading through changes, but made some initial comments. Really great work! ❤️ Way cleaner and more extensible.

lib/packwerk/commands.rb
davidstosik: > Or do we also need to support the existence of the `UpdateCommand` constant as well? 🤔

We don't. It's a class introduced in this PR, but doesn't have to.

My problem is as follows:
- we use autoloading, so the class is not loaded _until_ we have defined that it's necessary.
- if we implement a system to declare aliases, then we need it to be run during initialization
- I would like to avoid having to declare aliases _outside_ of the command class itself (open-closed principle)

I guess the fact that we need these `autoload` is already an infringement to the open-closed principle though... 🤔 

Check this commit to see an early idea I had, where each command was able to declare itself, and its aliases in the process. The caveat was that command classes needed to be eager loaded.

lib/packwerk/commands.rb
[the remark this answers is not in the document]
davidstosik: Yeah, so my idea was:
1. I wanted to extract the `const_get` + `rescue` to its own method to make the thing more readable (I didn't want to pollute `Packwerk::Cli#execute_command` with it). It first was a class method on `Packwerk::Cli`, and then naturally moved to `Packwerk::Commands`.
2. Once that was done, I noticed I was using exceptions as control flow whereas there is n appropriate method (`#const_defined?`) to check if the constant exists without raising `NameError`, so I used it instead.

I now realized that when I moved the method, I could have omitted the `Commands` here:

```diff
       sig { params(command: String).returns(T.nilable(T.class_of(Commands::BaseCommand))) }
       def class_for(command)
         class_name = command.sub(" ", "_").underscore.classify + "Command"
-        if Commands.const_defined?(class_name)
-          Commands.const_get(class_name) # rubocop:disable Sorbet/ConstantsFromStrings
+        if const_defined?(class_name)
+          const_get(class_name) # rubocop:disable Sorbet/ConstantsFromStrings
```

lib/packwerk/commands/help_command.rb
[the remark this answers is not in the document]
davidstosik: > registry pattern

Yup! See my other comment with a link to a specific commit. That's what I used first. But that would mean we need to eager load CLI commands.

lib/packwerk/commands.rb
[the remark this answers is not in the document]
shioyama: > I didn't want to pollute `Packwerk::Cli#execute_command` with it

Yeah but now we have a new public method to support, just for a single use case. Feels to me more appropriate just to hide the implementation in the only place that actually uses it (esp since it's just a couple lines).

> I noticed I was using exceptions as control flow

Calling an undefined command _is_ an "exception" though, so feels to me that rescuing here is fine (to avoid exposing an exception to the user). I would consider it "control flow" if we were then doing something other work in the `rescue`, but we're literally just cleaning up the output.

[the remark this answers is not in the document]
davidstosik: >  a new public method to support
That's a very fair point! Maybe I can move the method back to `Cli`, make it private, and switch back to rescuing `NameError`. I don't have a strong opinion on this bit.

lib/packwerk/commands/help_command.rb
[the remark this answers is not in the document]
shioyama: Not necessarily. There could be a registry that can be pre-populated for in-gem commands. For commands added externally, the process would be to register your extension (for the help message) and also create an autoloadable constant, which would not be loaded until the command was actually used.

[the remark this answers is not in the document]
davidstosik: Ahh, so instead of registering the commands from inside the command class, we could do it outside (in a place that we know gets loaded on init), like so?

```rb
Commands.register("update-todo", help: "update package_todo.yml files", aliases: ["update"])
```

For the gem's core commands, these would all be located in the same location, while plugin gems would have to do that somewhere, where they get loaded (eg. their own `lib/packwerk/plugin_name.rb` file).

`#register` could possibly also offer a way to have command and class name be different, like so:

```rb
Commands.register("something", help: "does something", classname: "MyCommandClass")
```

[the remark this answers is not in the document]
shioyama: Yeah right exactly 👍

lib/packwerk/cli.rb
```diff
@@ -1,15 +1,29 @@
...
+    autoload :UpdateTodoCommand
+    autoload :UsesParseRun
+    autoload :ValidateCommand
+    autoload :VersionCommand
+
+    class << self
+      extend T::Sig
+      sig { params(name: String, help: String, aliases: T::Array[String]).void }
+      def register_command(name, help:, aliases: [])
+        CommandRegistry.register(name, help: help, aliases: aliases)
```
davidstosik, line 24: This just delegates to `CommandRegistry`. It allows me to keep that class private, while it exposes public methods to be used within Packwerk.

lib/packwerk/commands/help_command.rb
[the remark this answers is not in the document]
davidstosik: Some autoload tests caught me loading more than I should:
image

(Didn't know about `bin/rake test:loading`.)

So I added one more commit: 721b197 (Move command registration to Cli in order to avoid early loading)

lib/packwerk/cli/command_registry.rb
```diff
@@ -0,0 +1,64 @@
...
+      def initialize(name, help:, aliases: [])
+        @name = name
+        @help = help
+        @aliases = aliases
+      end
+
+      sig { returns(T.class_of(Cli::BaseCommand)) }
+      def command_class
+        classname = @name.sub(" ", "_").underscore.classify + "Command"
+        Cli.const_get(classname) # rubocop:disable Sorbet/ConstantsFromStrings
```
gmcgibbon, line 53: Can you explain your motivation to not use a `Packwerk::Cli::Commands` module to const_get from? We could fit all command related code in one place (eg. `Commands::Registry`, `Commands::Base`, `Commands::HelpCommand`, etc. It seems safer and more organized.

lib/packwerk/cli/check_command.rb
```diff
@@ -0,0 +1,48 @@
...
+          @out.puts(<<~MSG.squish)
+            No files found or given.
+            Specify files or check the include and exclude glob in the config file.
+          MSG
+
+          true
+        end
+
+        run_context = RunContext.from_configuration(@configuration)
+        offense_collection = OffenseCollection.new(@configuration.root_path)
```
gmcgibbon, line 22: These could be private methods.

lib/packwerk/cli/uses_parse_run.rb
```diff
@@ -0,0 +1,91 @@
...
+      sig { returns(ParseRun) }
+      def parse_run
+        ParseRun.new(
+          relative_file_set: @files_for_processing.files,
+          parallel: @configuration.parallel?,
+        )
+      end
+
+      sig { returns(T::Hash[Symbol, T.untyped]) }
+      def parse_options
```
gmcgibbon, line 61: Thinking ahead slightly, it might make sense for the base class to implement `def options; raise NotImplementedError; end` and have this be a memoized public method for commands to implement.

lib/packwerk/cli/command_registry.rb
```diff
@@ -0,0 +1,64 @@
+# typed: strict
+# frozen_string_literal: true
+
+module Packwerk
+  class Cli
+    class CommandRegistry
```
gmcgibbon, line 6: The instance of this class is an entry, not a registry, so it doesn't really make sense.  I would say that a better api would be `Commands.all` / `Commands.register`, and then name this class `Commands::Info`, `Commands::Lookup`, or `Commands::Entry` (etc.) so the registry can live on `Commands`.

lib/packwerk/cli.rb
```diff
@@ -49,181 +63,30 @@ def run(args)
...
-        relative_file_set: files_for_processing.files,
-        file_set_specified: files_for_processing.files_specified?,
-        configuration: @configuration,
-        progress_formatter: @progress_formatter,
-        offenses_formatter: formatter
-      )
-    end
   end
+
+  Cli.register_command("init", help: "set up packwerk")
```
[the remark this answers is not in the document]
davidstosik: See previous conversation with @shioyama : if I do so (which I'm not against), I'll need to either eager load command classes or at least lazy load them all when the user called for help.

lib/packwerk/cli/command_registry.rb
davidstosik: Without the need to auto match a command name with a class name, and with instead an explicit registration of each command, I felt `Commands` had lost its main purpose, that triggered its creation: hosting all commands and nothing else so that `packwerk not-a-command-but-const-with-matching-name-exists` would fail beautifully and not weirdly.
With its main purpose lost, I went for the "less code is better" approach.

lib/packwerk/cli/command_registry.rb
davidstosik: Yeah, the naming didn't please me. I'll have a look.

lib/packwerk/cli/command_registry.rb
gmcgibbon: Personally, I think the organization of the `commands/` namespace makes command specific code easier to reason about, but I don't feel strongly about it. I'm slightly worried about this from a public API perspective.

davidstosik: `Cli` is public, sure, but if I make my constants nested under it private, it shouldn't make a difference?
I'm not antagonizing, just trying to understand where the problem would be exactly.

lib/packwerk/cli/command_registry.rb
davidstosik: I renamed `CommandRegistry` to `LazyLoadedCommand` (to make the distinction with `BaseCommand`). [d49a7ea](https://github.com/Shopify/packwerk/pull/351/commits/d49a7eaa4726d31a2013be5002b01d1ebf561444)

davidstosik: > and then name this class `Commands::Info`, `Commands::Lookup`

Just understood what you meant here. Could work too. 🤔

davidstosik: 670fc67

lib/packwerk/cli/command_registry.rb
davidstosik: https://github.com/Shopify/packwerk/commit/670fc67b7413c5a03cc6896193ca64754228c9bd

lib/packwerk/cli.rb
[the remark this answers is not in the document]
davidstosik: https://github.com/Shopify/packwerk/pull/351/commits/c127e099b850121e0bd0e7931fdffec9baf13720

lib/packwerk/commands/check_command.rb
```diff
@@ -0,0 +1,60 @@
...
+        end
+
+        all_offenses = T.let([], T::Array[Offense])
+        on_interrupt = T.let(-> { @progress_formatter.interrupted }, T.proc.void)
```
gmcgibbon, line 24:
```suggestion
        on_interrupt = T.let(proc { @progress_formatter.interrupted }, T.proc.void)
```
These could also be memoized private methods

lib/packwerk/commands/lazy_loaded_entry.rb
```diff
@@ -0,0 +1,37 @@
+# typed: strict
+# frozen_string_literal: true
+
+module Packwerk
+  module Commands
+    class LazyLoadedEntry
```
gmcgibbon, line 6:
```suggestion
    class Entry
```
Nit: I think `LazyLoadedEntry` is a little too verbose.

lib/packwerk/commands/uses_parse_run.rb
```diff
@@ -0,0 +1,91 @@
...
+          configuration: Configuration,
+          out: T.any(StringIO, IO),
+          err_out: T.any(StringIO, IO),
+          progress_formatter: Formatters::ProgressFormatter,
+          offenses_formatter: OffensesFormatter,
+        ).void
+      end
+      def initialize(args, configuration:, out:, err_out:, progress_formatter:, offenses_formatter:)
+        @args = args
+        @configuration = configuration
```
gmcgibbon, line 26: These should be assigned by super, right?

lib/packwerk/commands/base_command.rb
```diff
@@ -0,0 +1,49 @@
...
+          offenses_formatter: OffensesFormatter,
+        ).void
+      end
+      def initialize(args, configuration:, out:, err_out:, progress_formatter:, offenses_formatter:)
+        @args = args
+        @configuration = configuration
+        @out = out
+        @err_out = err_out
+        @progress_formatter = progress_formatter
+        @offenses_formatter = offenses_formatter
```
gmcgibbon, line 42: I'm interested in how you feel about using ivars for stdout/stderr/args/formatters? Should we create private/protected/public readers for these ivars? It seems like it would be a nicer public API to me.

davidstosik: Yeah, I like the idea. I tend to define accessors for everything usually. I did not do it here because Sorbet requires a level of verbosity I was not super comfortable with. I'll give it a try.

lib/packwerk/commands/uses_parse_run.rb
davidstosik: https://github.com/Shopify/packwerk/pull/351/commits/2d9c5c02d34ba7f90df9c65a2f4bf21637e70973#diff-7a35006173631f73820a5c2882b286cbe58356933b7e74a50c5446d1c1547ce9L25

lib/packwerk/commands/base_command.rb
davidstosik: 2d9c5c02d34ba7f90df9c65a2f4bf21637e70973

lib/packwerk/commands/uses_parse_run.rb
gmcgibbon: Hm, I wonder if this is a bug in Sorbet. If we require the class as an ancestor, it should always expect the initializer to run.

davidstosik: But it doesn't know what any other initializer might run in the meantime? (If for example more modules are included.)

gmcgibbon: True, I guess it can't know for sure. The workaround is good 👍

lib/packwerk/commands/uses_parse_run.rb
```diff
@@ -0,0 +1,91 @@
...
+          configuration: Configuration,
+          out: T.any(StringIO, IO),
+          err_out: T.any(StringIO, IO),
+          progress_formatter: Formatters::ProgressFormatter,
+          offenses_formatter: OffensesFormatter,
+        ).void
+      end
+      def initialize(args, configuration:, out:, err_out:, progress_formatter:, offenses_formatter:)
+        super
+        @_parsed_options = T.let(nil, T.nilable(T::Hash[Symbol, T.untyped]))
```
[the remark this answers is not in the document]
davidstosik: Ah, I used an underscore because it's a memoization ivar for the `parsed_options` method, and Sorbet seemed to require this `T.let` in the initializer. This might not be true though, I'll check.

lib/packwerk/formatters/progress_formatter.rb
```diff
@@ -44,6 +53,7 @@ def mark_as_failed
       def interrupted
         @out.puts
         @out.puts("Manually interrupted. Violations caught so far are listed below:")
+        @out.puts
```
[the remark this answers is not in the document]
davidstosik: They are not new, they were just moved from somewhere else to keep the tests happy.
If I recall correctly, it used to be here:

https://github.com/Shopify/packwerk/blob/f317dea925fc89fcc0a9ed32ef01074015fd8a5c/lib/packwerk/cli.rb#L131

The change happened in this commit: 1bed3e8f258fc812f319d7421515297fb44ccc0b

lib/packwerk/commands/uses_parse_run.rb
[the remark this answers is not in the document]
davidstosik: 4ccb564944e7c1a3ac0cddb413cd9969dfdc4bf1

PR #357, opened 2023-05-18
Do not run duplicate CI checks

alexevanczuk:
Previously, I merged https://github.com/Shopify/packwerk/pull/236 (has screenshots), since we see duplicate checks run on every build.

This had to be reverted here https://github.com/Shopify/packwerk/pull/240 because it ended up not running checks on push.

Searching github community, I found a suggested solution:
https://github.com/orgs/community/discussions/26940#discussioncomment-5686753

We'll want to test this solution and revert if it doesn't work in the following contexts:
- running checks for a maintainer PR
- running checks for an external contributor PR (from a fork)
- running tests on main
- running tests on a branch that is NOT a PR. Currently I'm confident this will **not** work – do folks use branches without PRs and expect/need tests to run?


gmcgibbon (commented):
> running tests on a branch that is NOT a PR. Currently I'm confident this will not work – do folks use branches without PRs and expect/need tests to run?

Usually this shouldn't be a problem, but if we want to release a maintenance version like we did with 2.x (other branch) before releasing 3.x (main). I think it might be better to just leave things as they are, unless you can make CI run on branch pushes.

PR #367, opened 2023-08-02
Requalify strict violations

alxckn:

This is a proposal to change how `strict_mode_violations` offenses are gathered and prevent the `update_todo` from updating a package_todo.yml file with new violations once the strict mode is activated for a package. 

Currently the OffenseCollection class will only check whether an offense qualifies as a strict mode violation if it was already present in the file package_todo.yml, I think it should be up to the checker to determine whether offenses, known or unknown, qualify as strict mode violations.

Down the line, the goal would be to enable checkers to apply a 'less strict' mode where new violations cannot be added to the package_todo.yml file but violations that were written to the file before do not cause an error. This would be for packages where it is difficult to reach an empty package_todo.yml but we want to stop the bleeding.

- Always ask the checker whether an offense is a strict mode violation and add it to the list if it is
- Update of package_todo.yml does not include new strict mode violations. Being able to update the list with strict mode violations may lead to errors where new violations are not considered new after the file is updated

- Current usage of the gem and whether it would harm some existing workflows (I am not sure it would as a zero strict mode violations policy is currently enforced by all checkers)

## Type of Change

- [ ] Bugfix
- [x] New feature

### Additional Release Notes

Update with the `update_todo` command of package_todo.yml would fail in cases where it would work today

## Checklist

### DIFF
--- a/lib/packwerk/commands/update_todo_command.rb
+++ b/lib/packwerk/commands/update_todo_command.rb
@@ -39,12 +39,21 @@ module Packwerk
         offense_collection.add_offenses(offenses)
         offense_collection.persist_package_todo_files(run_context.package_set)
 
-        out.puts(<<~EOS)
-          #{offenses_formatter.show_offenses(offense_collection.errors + offense_collection.strict_mode_violations)}
-          ✅ `package_todo.yml` has been updated.
-        EOS
+        unlisted_strict_mode_violations = offense_collection.unlisted_strict_mode_violations
 
-        offense_collection.strict_mode_violations.empty? && offense_collection.errors.empty?
+        messages = [
+          offenses_formatter.show_offenses(offense_collection.errors + unlisted_strict_mode_violations),
+        ]
+
+        messages << if unlisted_strict_mode_violations.any?
+          "⚠️ `package_todo.yml` has been updated, but unlisted strict mode violations were not added."
+        else
+          "✅ `package_todo.yml` has been updated."
+        end
+
+        out.puts(messages.select(&:present?).join("\n") + "\n")
+
+        unlisted_strict_mode_violations.empty? && offense_collection.errors.empty?
       end
     end
   end
--- a/test/unit/packwerk/commands/update_todo_command_test.rb
+++ b/test/unit/packwerk/commands/update_todo_command_test.rb
@@ -126,15 +126,12 @@ module Packwerk
 
           📦 Finished in \\d+\\.\\d+ seconds
 
-          components/source/some/path.rb
-          some message
-
           components/source/other/path.rb
           other message
 
-          2 offenses detected
+          1 offense detected
 
-          ✅ `package_todo.yml` has been updated.
+          ⚠️ `package_todo.yml` has been updated, but unlisted strict mode violations were not added.
         EOS
         assert_match(/#{expected_output}/, out.string)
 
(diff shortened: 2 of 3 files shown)


lib/packwerk/commands/update_todo_command.rb
```diff
@@ -37,14 +37,22 @@ def run
 
         offense_collection = OffenseCollection.new(configuration.root_path)
         offense_collection.add_offenses(offenses)
-        offense_collection.persist_package_todo_files(run_context.package_set)
 
-        out.puts(<<~EOS)
-          #{offenses_formatter.show_offenses(offense_collection.errors)}
-          ✅ `package_todo.yml` has been updated.
-        EOS
+        if offense_collection.strict_mode_violations.any?
```
alexevanczuk, line 41: It looks like if there are *any* strict mode violations, we fail out immediately and do not make any changes to TODO files.

What do you think of instead making all the changes we *can* make when updating, but then if there are any strict mode violations, we inform the user and exit with a failure?

alexevanczuk (commented):
I really like the idea here. The fact that strict mode will only fail *after* you run update and then run `check` makes it a lot less useful, because most folks will run `update` and assume everything is fine without running check, and then wait until CI fails to realize they are violating strict mode.

Failing during update makes a lot of sense to me, since it reduces the number and duration of feedback loops. I find the idea of a softer strict mode interesting. Allowing the violations to exist in the TODO file, but not permit new ones to be added with the tooling is definitely interesting. Technically this can be bypassed by updating TODO files by hand, but I think that's okay (since we're trying to create happy-path guard rails, rather than a 'barbed wire fence').

It feels like possibly an alternative to https://github.com/Shopify/packwerk/pull/365, cc @athieriot

alxckn:
> It feels like possibly an alternative to https://github.com/Shopify/packwerk/pull/365, cc @athieriot

I think what this PR proposes and what's in @athieriot's PR could be complementary.
At Doctolib we have some heterogeneous needs with packages where we'd like to use strict mode only for _new_ violations (this PR + changes in the way strict mode violations are computed at the checker level) and other packages where we'd like to enforce strict mode for 'application' code but we can't enforce it in test setups.

test/unit/packwerk/commands/update_todo_command_test.rb
```diff
@@ -77,6 +77,74 @@ class UpdateTodoCommandTest < Minitest::Test
...
+
+          📦 Finished in \\d+\\.\\d+ seconds
+
+          components/source/some/path.rb
+          some message
+
+          components/source/other/path.rb
+          other message
+
+          2 offenses detected
```
alexevanczuk, line 135: It feels like this message should say something about not being able to successfully add certain violations to the TODO list due to strict mode. Right now it's unclear why it wasn't added.

alxckn: You're right, I've amended the message to only include strict violations that are blocking for the `update_todo` command to complete successfully and changed the final message to explain that something went wrong.

alexevanczuk:
This looks good to me. @rafaelfranca do you have any feedback or are we good to merge and release?

rafaelfranca:
CLA needs to be signed before we merge this though.

rafaelfranca:
Ah, it was.

PR #368, opened 2023-08-08
strict_for_new mode: fail check only for violations unlisted in package_todo.yml

alxckn:

Following https://github.com/Shopify/packwerk/pull/367, this PR introduces a new mode for the dependency checker: 'strict_for_new'.
With this mode, only new violations, unlisted in the package_todo.yml file will result in an error when using the 'check' command.

Let the checker know whether an offense was listed or not in the package_todo.yml to let it decide whether it should be a strict violation.

## Type of Change

- [ ] Bugfix
- [x] New feature

### Additional Release Notes

While the change is not breaking for this gem alone, it is a breaking change for the API that's used by packwerk extensions. So a major version bump will be required.

## Checklist

- [x] It is safe to rollback this change. (any 'strict_for_new' configuration will need to be rolled back as well)

### DIFF
--- a/lib/packwerk/checker.rb
+++ b/lib/packwerk/checker.rb
@@ -51,8 +51,8 @@ module Packwerk
     sig { abstract.returns(String) }
     def violation_type; end
 
-    sig { abstract.params(listed_offense: ReferenceOffense).returns(T::Boolean) }
-    def strict_mode_violation?(listed_offense); end
+    sig { abstract.params(offense: ReferenceOffense, already_listed: T::Boolean).returns(T::Boolean) }
+    def strict_mode_violation?(offense, already_listed:); end
 
     sig { abstract.params(reference: Reference).returns(T::Boolean) }
     def invalid_reference?(reference); end
--- a/lib/packwerk/offense_collection.rb
+++ b/lib/packwerk/offense_collection.rb
@@ -59,7 +59,7 @@ module Packwerk
 
       new_violations << offense unless already_listed
 
-      if strict_mode_violation?(offense)
+      if strict_mode_violation?(offense, already_listed)
         add_to_package_todo(offense) if already_listed
         strict_mode_violations << offense
       else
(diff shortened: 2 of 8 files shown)


rafaelfranca:
I'm not sure having a configurable mode is necessary. Let's just make that the default behavior of strict. If you have recorded violations and you enable strict mode, those should be fine, but new violations aren't. If you don't have any recorded violation and you enable strict mode same thing apply.

alxckn:
@rafaelfranca ok. In your opinion, is it still the Checker's role to determine whether a violation is strict or not depending on whether it was listed or not in the package_todo.yml?

alexevanczuk:
@alxckn 

So the strict refers to the checker's level of enforcement, not the violation itself. 

When the level is "true," we fail check on unlisted violations already and also permit new violations to be added. 

When the level is "strict," we do not permit any violations ever.

It sounds like you want something in the middle -- allow listed violations, but no new ones.

I think @rafaelfranca 's suggestion makes sense. The only downside is now strict is less strict than it used to be -- a package with strict enforcement may have violations (or more easily regress to having them), and there's no mode to say "violation free." Upside is "strict" mode becomes a lot more useful to stop the bleeding. Curious what others think!

alxckn:
@rafaelfranca @alexevanczuk 

My proposal was to avoid having too big of a breaking change. If organizations rely on the previous definition of 'strict' to function then it might be a complicated upgrade.

Keeping two distinct modes could have upsides, for instance not all violations for a given package are recorded within its package_todo.yml (I'm thinking of the privacy checker). Maintaining a 'zero violation policy' (existing strict mode behavior) with packwerk only enforcing a 'no new violation' policy could become a challenge for package owners as they don't own the file where the violations are being written. This may be a separate issue though.

> strict refers to the checker's level of enforcement, not the violation itself
I believe this can be true if strict mode only applies to unlisted violations. If we wanted to support two different modes such as 'strict' and 'strict_for_new', then it's the violation that is either strict or not depending on the package configuration (source or destination) and the listed status of the violation.

In any case I've updated the PR to reflect this direction, i.e. the check command may only fail because of strict mode violations if they're coming from unlisted offenses.

alxckn:
@rafaelfranca @alexevanczuk I'm available to continue the discussion or amend the PR as needed :)

I did make the change to support the behavior suggested by Rafael, ie. change the existing strict mode to not account for violations that are already present in the package_todo.yml file.

alxckn:
@oboxodo I had initially proposed a `strict_for_new_mode` in this PR (see [first commit](https://github.com/Shopify/packwerk/pull/368/commits/4bd03b2bd5a79fcf6da16d3d88af230732fa7d94)). 
There are also good arguments for changing the behavior of the existing `strict` mode, among which:
- simplicity, avoid opening the door to new modes that bring new subtleties each time
- it does not change much for existing packages that are already using strict mode, no new violation is allowed

rafaelfranca (approved):

PR #371, opened 2023-09-21
Update stale violations error message to use bin/packwerk command

christhesoul:

A more correct error message for those of us who just blindly copy and paste things. :)

Replaced `packwerk` with `bin/packwerk`

🤷 

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist

### DIFF
--- a/lib/packwerk/formatters/default_offenses_formatter.rb
+++ b/lib/packwerk/formatters/default_offenses_formatter.rb
@@ -23,7 +23,7 @@ module Packwerk
       sig { override.params(offense_collection: OffenseCollection, file_set: T::Set[String]).returns(String) }
       def show_stale_violations(offense_collection, file_set)
         if offense_collection.stale_violations?(file_set)
-          "There were stale violations found, please run `packwerk update-todo`"
+          "There were stale violations found, please run `bin/packwerk update-todo`"
         else
           "No stale violations detected"
         end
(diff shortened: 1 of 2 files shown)


gmcgibbon:
Thank you!

rafaelfranca:
This PR is wrong. Reverting. It assumes you have a binstub that you might not have.

rafaelfranca:
See https://github.com/Shopify/packwerk/pull/138#issuecomment-910753855

rafaelfranca:
bundler actually requires you know now a lot about it. That is why you see a lot of places recommending `bundle exec` to be used to call commands. Binstubs (the `bin/` files) are the equivalent way to tell bundler to execute a command without having to use `bundle exec`. I don't think we can recommend either of those because some people might not have binstubs, and `bundle exec` is an anti-patter. The ideal would be for environment to add trusted `bin` folder to the shell path, so you don't need to remember to use `bin/packwerk` over `packwerk`.

christhesoul:
Why is `bundle exec` an anti-pattern, out of interest?

I always saw it as a way of saying "I want to run this command within the context of a bunch of dependencies that I have defined in the Gemfile", and over time just used it due to muscle memory.

> The ideal would be for environment to add trusted `bin` folder to the shell path

Yeah, that's a really nice solution. What prevents us doing this in our spin environments?

rafaelfranca:
> What prevents us doing this in our spin environments?

I think nobody got to it. Because we already know a directory is trusted on shadowenv

> Why is `bundle exec` an anti-pattern, out of interest?

Because for some commands, like Rails you activate bundler twice.

PR #378, opened 2023-11-08
Update README with automatic_namespaces

gap777:
Provide more visibility of a helpful pack tool.

Separate gem that parses packwerk.yml

Is this the appropriate place to communicate this option?

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist


gmcgibbon (changes requested):
Thanks for the PR, but the gem seems more keyed to Packs rather than Packwerk. It also seems to monkey patch Rails, which we want to promote. Sorry

PR #384, opened 2024-01-03
Don't report violations until told to do so

exterm:

Packwerk is not opinionated about where you draw boundaries and which direction your dependencies have.

It's a tool that helps you enforce or move towards an architecture that you come up with.

So it shouldn't return errors unless a constraint specified by the user is violated.

Right now we set `enforce_dependencies: true` in the root package in `packwerk init`, with no list of dependencies (defaults to empty). This means that any newly introduced package may cause packwerk to report violations, even before the user has specified how this new package fits in with the application's architecture.

For example, see [my blog post about enforcing `lib/` to not have dependencies](https://simplexity.quest/posts/2023-12-23-lib-is-for-libraries.html#making-it-so).

Do not enforce anything on the root package by default. This is the simplest solution. Packwerk doesn't enforce anything by default. That means only user-specified constraints are enforced.

Is there any documentation I need to change along with this?

## Type of Change

- [X] Bugfix
  - I would call this a bugfix to `packwerk init`, but at the very least it's a breaking change to it. 
- [ ] New feature

### Additional Release Notes

  - Not a breaking change for existing installs.

## Checklist

### DIFF
--- a/lib/packwerk/generators/templates/package.yml
+++ b/lib/packwerk/generators/templates/package.yml
@@ -2,8 +2,8 @@
 # Please validate the configuration using `packwerk validate` (for Rails applications) or running the auto generated
 # test case (for non-Rails projects). You can then use `packwerk check` to check your code.
 
-# Turn on dependency checks for this package
-enforce_dependencies: true
+# Change to `true` to turn on dependency checks for this package
+enforce_dependencies: false
 
 # A list of this package's dependencies
 # Note that packages in this list require their own `package.yml` file


gmcgibbon (commented):
Personally I think that Packwerk should be enforcing dependencies by default. Without doing this, Packwerk holds no value in the codebase. If the developer chooses to not enforce dependencies at this time, it is their decision, and it should be done explicitly IMO. It may be annoying to have violations in new packages as a result of this behaviour, but I think it is well worth it to have a stricter default.

What we may want to do here though is uncomment the dependencies key and make it an empty array. In theory, the application package should not have any dependencies, but that is up to the developer to decide.

cc @Shopify/packwerk if anyone else has opinions.

exterm:
Hi Gannon!

I appreciate your thoughts on this, but respectfully disagree.

> What we may want to do here though is uncomment the dependencies key and make it an empty array.

That would be a slight improvement IMO but doesn't really address my concern.

> should be enforcing dependencies by default. Without doing this, Packwerk holds no value in the codebase.

:shrug: Packwerk doesn't know anything about the structure of your codebase, so it can't enforce anything out of the box. You have to define packages. Who says you will want to enforce dependencies on the root package? You may want to enforce dependencies only on one or multiple of the smaller packages you create.

The design of this tool is at its core focused on avoiding false positives, and in many realistic workflows these violations on the root package that spring up when you start packaging parts of your application do share a lot of properties with the false positives we tried to avoid in other places. That's why I chose this specific title and description for the PR. Violations of rules specified by the user are useful output. Everything else is potentially confusing noise.

Rubocop comes with cops that are enabled by default, which makes sense because there are some statements that we can make about Ruby code that are universally true. Packwerk is concerned with the shape and architecture of applications though where we can make a lot less assumptions. If we do want to enforce things out of the box the one obvious property for all Rails apps seems to be the one I describe in [my blog post](https://simplexity.quest/posts/2023-12-23-lib-is-for-libraries.html) - libraries shouldn't depend on the application. But I don't think "the root package doesn't depend on anything" is a valid assumption to make for all Rails apps. 

Maybe the cleaner solution is not to define a root package out of the box in the first place? I'm not sure. I find the root `package.yml` to be a nice template, and having it also makes it so all code is in _some_ package by default.

IIRC the reason we introduced the root `package.yml` to the generator was so that we could have all code in a package without having to add an _implicit_ root package (which was the original solution).

exterm:
> Tools inspired by Packwerk (eg. packs) have generator-like CLI subcommands to create packages, which we could introduce to make package creation more intentional.

👍🏼 👍🏼

exterm:
Two ideas:

1. Make packwerk unopinionated (which it mostly already is), put opinions into a separate gem. That gem could then also contain generators to enable people to replicate Shopify's usage in Componentization (as asked for in https://github.com/Shopify/packwerk/discussions/361)

2. Make packwerk opinionated by default but let users opt out - e.g. `packwerk init --bare` or something like that. Arguably that isn't too different from `packwerk init` followed by editing the configuration, but it's a little nicer.

tylerwillingham:
> I also think that if we add a package generator in the future to the gem, we should enforce dependencies by default for those. That's not really a concern for this change though.

Just voicing support for this. My team has created some bespoke (but lean) generators that effectively generate a package.yml that has `enforce_dependencies: true` and `dependencies: []` plus some metadata that we care about and an opinionated folder structure - so we've seen the use in some default opinions

exterm:
>  I also think that if we add a package generator in the future to the gem, we should enforce dependencies by default for those.

Yep, full agreement from me too!

PR #385, opened 2024-01-16
Change recommended drawing library

euglena1215:

Change the drawing library introduced as the packwerk ecosystem from [graphwerk](https://github.com/samuelgiles/graphwerk) gem to [visualize_packs](https://github.com/rubyatscale/visualize_packs) gem.

The graphwerk gem is introduced as part of the ecosystem, but the graphwerk gem has not been maintained recently.

Since the visualize_packs gem is more actively maintained, I thought it would be appropriate to introduce the visualize_packs gem as part of the ecosystem.

Please determine if the visualize_packs gem is appropriate as a recommended drawing library.

## Type of Change

- [ ] Bugfix
- [ ] New feature

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ Read [USAGE.md](USAGE.md) for usage once Packwerk is installed on your project.
 
 Various third parties have built tooling on top of packwerk. Here's a selection of some that might prove useful:
 
-- https://github.com/bellroy/graphwerk draws a graph of your package dependencies
+- https://github.com/rubyatscale/visualize_packs draws a graph of your package dependencies
 - https://github.com/rubyatscale/packwerk-vscode integrates packwerk into Visual Studio Code so you can see violations right in your editor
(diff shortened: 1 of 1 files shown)


euglena1215:
I have signed the CLA!

gmcgibbon (changes requested):
Thank you for your contribution, but this cannot be accepted. It is unfortunate that graphwerk hasn't been maintained recently, but I don't want to imply that one needs to use "packs" for visualization. It adds many features on top of Packwerk, some of which the maintainers of Packwerk disagree with.

euglena1215:
@gmcgibbon 
Thanks for the review. I did not know that the visualize_packs gem depends on the directory name "packs".
So I can understand why it is not accepetable.

alexevanczuk:
Hey @euglena1215 just wanted to chime in to say that `packs-rails` and the `packs` spec in general defaults to `./packs` as the root dir, but the pack directories are totally configurable. Only saying that to clear up any misconceptions!

PR #388, opened 2024-02-01
Use prism parser via translator

exterm:

Proof of concept for discussion https://github.com/Shopify/packwerk/discussions/387. Packwerk spends about 2/3 of the time it runs on a large code base in parsing Ruby. Prism is ~5x faster than the current parser (`whitequark/parser`). So this should be a significant speed boost for packwerk.

It looks like this could be a simple, almost trivial change.

We could use Prism directly, but that would require significant changes to Packwerk. Luckily @kddnewton has included a translator with prism that provides compatibility with `whitequark/parser`, so that it should be a drop-in replacement. We won't get the same performance improvements, but we get [some](https://github.com/ruby/prism/pull/2359), and almost for free.

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist


kddnewton:
@exterm Can you use prism v0.20.0 itself? It has `Prism::Translation::Parser` which is effectively this. I'm going to deprecate parser-prism

exterm:
I'm running into lots of errors that look like I'm not requiring the right thing.

`require "prism/translation/parser"` => 

```
NameError: uninitialized constant Prism::Compiler

      class Compiler < ::Prism::Compiler
                              ^^^^^^^^^^
Did you mean?  Complex
    /home/[...]/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/prism-0.20.0/lib/prism/translation/parser/compiler.rb:8:in `<class:Parser>'
```

When I add a `require "prism/compiler"` I get 

```
NoMethodError: undefined method `parse' for Prism:Module

        result = unwrap(Prism.parse(source, filepath: source_buffer.name))
                             ^^^^^^
    /home/[...]/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/prism-0.20.0/lib/prism/translation/parser.rb:45:in `parse'
```

kddnewton:
Ahh sorry it's actually the opposite, you should only require `prism` and not require the nested one. Everything is autoloaded so it should work just fine with just the top-level require.

exterm:
Done. I had to change a test because the error message was different. Ideally it would be the same as from `Parser::CurrentRuby`, no?

exterm:
@gmcgibbon FYI it seems to "just work" according to the test suite. Real world testing is required.

exterm:
I'll wait for the next prism release with the bugfix before I make this "ready for review"

kddnewton:
@exterm I'll release a patch release on Monday first thing. It'll be v0.20.1.

Thank you very much for your work on this! I'm still very interested in replacing the parser with prism as a whole (that should be much better than using the translation layer) but I recognize that's quite a bit more involved. Is that something you're still interested in working on? I'd be happy to help support in any way I can.

kddnewton:
The last thing is that we've worked hard to provide a really good Ruby API. Instead of dealing with a single class with a type field, all of our nodes have their own classes. We have utility functions on all of the classes that should make it quite a bit easier to use. Also, if you need anything, you have a line directly to the main contributor and I'll add whatever functions you may need.

kddnewton:
@exterm v0.21.0 is out now with these bug fixes

shageman:
Testing this. I am atm stuck on prism 0.17 due to our version of rbi. Had to read through the discussion to see the version dependency.

exterm:
mmmh, interesting. That might make a release incorporating this a little harder.

kddnewton:
I just opened a PR to add support for Ruby 2.7, so this won’t be an issue
once I release that.

exterm:
This is ready to go

packwerk.gemspec
```diff
@@ -50,6 +50,7 @@ Gem::Specification.new do |spec|
   # For Ruby parsing
   spec.add_dependency("ast")
   spec.add_dependency("parser")
+  spec.add_dependency("prism", ">= 0.22.0") # 0.21.0 fixes numbered block params, 0.22.0 adds Ruby 2.7 compatibility
```
gmcgibbon, line 53: Do we need to use parser still or can we drop it as a dependency now that we're on prism?

exterm: Yeah besides, in this PR packwerk still uses `Parser::AST` and a few other things.

A "clean" migration to Prism without using the translation layer, and using prism's AST format instead, would let us get rid of `parser` AFAICT. But it's significantly more work.

I've looked into it a bit, and among other things `NodeHelpers` has to be completely rewritten. I'm thinking we may want to completely remove it though since prism's AST format is a lot more user friendly and may not need this kind of leaky wrapper around it.

(@kddnewton the only thing that's been a little annoying is that the types don't seem to represent shared interfaces - e.g. between symbol and string nodes, or between class and module nodes. What's a good place to discuss these things?)

gmcgibbon: Looks like on a small app, we get around the same to slightly better performance that takes roughly `0.9s` to complete:
Before:
```
📦 Finished in 0.9 seconds
```
After:
```
📦 Finished in 0.78 seconds
```

 On a larger app, we see a significant slowdown for cached runs:
Before:
```
📦 Finished in 18.31 seconds
```
After:
```
📦 Finished in 30.77 seconds
```

I can't merge this until we find out why this is happening.

gmcgibbon: Before:
```
📦 Finished in 16.88 seconds

bin/packwerk check  0.30s user 0.04s system 1% cpu 33.053 total
```
After:
```
📦 Finished in 17.98 seconds

bin/packwerk check  0.29s user 0.04s system 0% cpu 34.413 total
```

Looks like we're at the same-ish difference to me now. How many times faster do we expect Prism to be using the translation layer?

gmcgibbon: @exterm do you have an application that benefits from this change? If you do, please provide benchmarks.

exterm: @kddnewton parsing is about 2/3 of packwerk runtime on large apps. So I'd expect a big speedup and I did get a big speedup on the app I was working on (~1 million Ruby LOC). I'll do another benchmark after updating to prism 0.23.0

exterm:
Up to date measurements from a reasonably sized Rails app, using spring. Sadly I can only currently access it through a Github Codespace, so it's a highly virtualized environment. I did not get a lot of variation on repeated runs though.

### Packwerk 3.1.0
cached: ~2.6s
uncached: ~32s

### Packwerk 3.1.0 with `Prism::Translator::Parser` from Prism 0.24.0

cached: ~2.6s
uncached: ~20s

Cached run times are the same, as expected - no parsing happens on repeated cached runs.

exterm:
For a single 3k line file, the difference is about 0.8s vs 1.1s - it's not dramatic, but it's faster. Moving off the translator layer would be better but is more work. It could be done as a follow-up to this PR.

packwerk.gemspec
```diff
@@ -50,6 +50,7 @@ Gem::Specification.new do |spec|
   # For Ruby parsing
   spec.add_dependency("ast")
   spec.add_dependency("parser")
+  spec.add_dependency("prism", ">= 0.23.0") # 0.24.0 fixes a performance issue with the parser translator
```
gmcgibbon, line 53: Let's change this to 0.24.0

gmcgibbon (approved):

packwerk.gemspec
exterm: I was goign to but somehow I only changed the comment 🤦🏼

gmcgibbon:
Thank you @exterm, I'll run a release for this now.

tylerwillingham:
25% faster on the main Rails app I work in 👍

PR #394, opened 2024-03-07
Custom offenses formatter docs fix

timjnh:

Fix the docs!  More specifically a broken link in them

I updated the URL to what looked like the correct destination.  Why?  I just felt like it was the right thing to do.

The link to the DefaultOffensesFormatter in the Custom Offenses Formatter section

## Type of Change

- [ ] Bugfix
- [ ] New feature

## Checklist


gmcgibbon (approved):
Thanks!

PR #397, opened 2024-03-19
Allow fetching all references for a file, or all files, using public API

exterm:

For introspection (e.g. network graphs & graphs over time) as well as advanced enforcement regimes that may be difficult to implement in a packwerk checker extension it would be great to be able to get a list of all static constant references from a file or between all files.

Extracting constants references is where packwerk's core complexity lies and we should allow its reuse directly, without having to go through the more opinionated and specific logic built on top of it.

I listed a naive approach in the first commit that doesn't require any modifications to existing packwerk code. This is the approach I currently use to generate detailed statistics about a monolith that I am working on. However, it uses a slew of private APIs.

In the second commit I propose a new module that creates public API for this functionality.

- Are we exposing more surface area than we need to? Anything specific we wouldn't want to expose?
- Can you think of a cleaner way to implement this?

## Caveats
- I decided to raise on parser errors so we don't have to expose the Offense class and generally have a simpler interface. I'm happy to discuss other options.
- Because of the design of the RunContext class, which stems from a time when packwerk was a rubocop cop and we had to work around some of rubocop's limitations, is complex and entangled. Thus
  - it is difficult to pry apart, as you can see in the slightly awkward implementation of this feature
  - it is notoriously hard to test; indeed it doesn't have unit tests
  - Anything _relying on_ RunContext is also hard to test, and so far I haven't been able to come up with a good way to test this new functionality.

## Type of Change

- [ ] Bugfix
- [X] New feature

## Checklist

## Additional Notes

This work is sponsored by https://www.onemedical.com/


exterm:
@gmcgibbon any interest in merging this?

exterm:
@rafaelfranca any interest in merging this? It would make analysis of the (actual) dependency graph considerably easier.

lib/packwerk/run_context.rb
```diff
@@ -75,15 +75,29 @@ def initialize(
...
     def process_file(relative_file:)
+      reference_checker = ReferenceChecking::ReferenceChecker.new(@checkers)
+
+      references_result = references_from_file(relative_file: relative_file)
+
+      references_result.file_offenses +
+        references_result.references.flat_map { |reference| reference_checker.call(reference) }
+    end
+
+    class FileReferencesResult < T::Struct
```
gmcgibbon, line 86: I believe T::Struct is slower than a normal struct. Please provide a benchmark, or switch to a plain old Struct (I think we only use Struct in other files).

lib/packwerk/references_from_file.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+
+    sig { params(config: Packwerk::Configuration).void }
+    def initialize(config = Configuration.from_path)
+      @config = config
+      @run_context = T.let(RunContext.from_configuration(@config), RunContext)
+    end
+
+    sig { params(relative_file_paths: T::Array[String]).returns(T::Array[Packwerk::Reference]) }
+    def list_all(relative_file_paths: [])
```
gmcgibbon, line 25: So is this the main method you use? Do we need to make the others public?

lib/packwerk/run_context.rb
exterm: There are 4 other uses of `T::Struct` around the codebase and 4 uses of `Struct`.

I did not know about performance differences and chose `T::Struct` because it's typed, and I appreciate the added documentation and type safety, especially within this class that is a little messy and doesn't have great tests.

exterm: A quick benchmark shows that access is about the same, while instantiation is ~5x slower for `T::Struct`. I was still able to instantiate 380k `T::Struct` within a second on my laptop.

lib/packwerk/references_from_file.rb
exterm: I pushed a commit that gets rid of the public `files` method in favour of some more elaborate stubbing in the test. It also renames the two remaining public methods so that hopefully their purpose is clearer.

exterm: mmh, I wonder now whether it makes sense to inject a FilesForProcessing instance in the test instead.

Issue #297, opened 2023-01-24
Packwerk thinks that `app/helpers/foobar_fizz.rb` is in the package `app/helpers/foobar` and not `app/helpers`


andrewhamon:
**Description**
When a file has a prefix that matches an adjacent folder with a package in that folder, Packwerk thinks that the file belongs to the wrong package.

Directory layout:

```
app/helpers
├── package.yml
├── foobar_fizz.rb     # class FoobarFizz, depends on Meow::Woof. NOT part of the app/helpers/foobar package
├── foobar
│   ├── package.yml
│   └── subclass.rb    # class Foobar::Subclass
└── meow
    ├── package.yml
    └── woof.rb        # class Meow::Woof

```

`foobar_fizz.rb` is in the directory/package `app/helpers`. It depends on the package `app/helpers/meow`. Packwerk wants that dependency declared in `app/helpers/foobar` rather than `app/helpers`.

**To Reproduce**
Minimal reproduction: https://github.com/andrewhamon/packwerk-reproduction#readme

In the repo, run:
```
bundle install
bundle exec packwerk check
```

**Expected Behaviour**
There should be no error, as the dependency was declared in the correct package. If the package `app/helpers/foobar` (which is uninvolved) is removed, the errors go away, which is very surprising.

**Version Information**
 - Packwerk: 2.2.2
 - Ruby: 3.1.2

exterm:
yeah it is this https://github.com/Shopify/packwerk/blob/f52fb63cb42336e6b470c9079041b1e56e7b4775/lib/packwerk/package.rb#L45

PR #398, opened 2024-03-19
don't associate a file with a different package that matches a prefix

exterm:
Fix https://github.com/Shopify/packwerk/issues/297

I think technically we shouldn't do string comparison for this, but the fix was so easy, and swapping in Pathname everywhere quite a bit of work, so I opted for this variant

## Type of Change

- [X] Bugfix
- [ ] New feature

### Additional Release Notes

- [X] Breaking change (fix or feature that would cause existing functionality to change)
This may reattribute files to other packages (the correct packages) and thus add or remove violations.

## Checklist

## Additional Notes

This work is sponsored by https://www.onemedical.com/

### DIFF
--- a/lib/packwerk/package.rb
+++ b/lib/packwerk/package.rb
@@ -42,7 +42,7 @@ module Packwerk
     def package_path?(path)
       return true if root?
 
-      path.start_with?(@name)
+      path.start_with?(@name + "/")
     end
 
     sig { params(other: T.untyped).returns(T.nilable(Integer)) }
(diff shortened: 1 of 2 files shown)


exterm:
@rafaelfranca I see you've been active on this repo recently - can you give this small bugfix a review?

gmcgibbon (approved):

seanarnold:
@gmcgibbon do you mind cutting a release with this included?

Issue #406, opened 2024-06-17
[Bug Report] Dependency violation not flagged when Zeitwerk collapsing is used


pareeohnos:
**Description**
A package calling another package is not triggering a violation

**To Reproduce**
- Create `app/modules` folder
- Create two packages in the new folder, with `package.yml` files enforcing dependencies
- Add a class in each package. Make one call the other
- Run `bin/packwerk check`

**Expected Behaviour**
A violation error to be raised

**Screenshots**
I've created a reproduction of the issue here https://github.com/pareeohnos/packwerk-issue

**Version Information**
 - Packwerk: 3.2
 - Ruby 3.3.0

**Additional Context**
The project structure contains modules in the `app/modules` directory. Within each package, each sub-directory is collapsed in Zeitwerk, so for example the directory `app/modules/package_one/services/service_one.rb` is loaded its `PackageOne::ServiceOne` instead of `PackageOne::Services::ServiceOne`.

I've tried removing the dependency on `.` and it triggers violations, but this is because the `ApplicationController` and `ApplicationRecord` classes are both located in the root of the project - it still makes not mention of the cross package violation.

If I remove the collapsing altogether, it starts to work correctly but this causes issues in our application so we need the collapsing.

fxn:
Packwerk uses its own constants resolver.

The last version of Zeitwerk provides [Zeitwerk::Loader#all_expected_cpaths](https://github.com/fxn/zeitwerk?tab=readme-ov-file#zeitwerkloaderall_expected_cpaths) that could help Packwerk honor user configuration like that by delegating that resolution to the Rails autoloaders.

Catsuko:
This is acknowledged in the [README](https://github.com/Shopify/packwerk?tab=readme-ov-file#limitations):
>Support for custom Zeitwerk configuration is limited. If [custom ActiveSupport inflections](https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#customizing-inflections) are used, Packwerk will understand that and everything is fine. However, if Zeitwerk is directly configured with [custom Zeitwerk inflections](https://github.com/fxn/zeitwerk#inflection) or to [collapse directories](https://github.com/fxn/zeitwerk#collapsing-directories), Packwerk will get confused and produce false positives.

I also hit this while trying to get a similar namespace to the one you mentioned 😭 Although I ended up giving up, while investigating I came across [this gem](https://github.com/gap777/automatic_namespaces) which might be of interest to you.

[...]

PR #410, opened 2024-07-18
Proof of concept for removing ConstantResolver and using Zeitwerk for ConstantDiscovery

Catsuko:

I am interested in using Zeitwerk features such as [collapsing directories](https://github.com/fxn/zeitwerk?tab=readme-ov-file#collapsing-directories) however currently Packwerk doesn't support this.

Based on [this comment](https://github.com/Shopify/packwerk/issues/406#issuecomment-2224756953) in a [related issue](https://github.com/Shopify/packwerk/issues/406), this PR adds an implementation of `ConstantDiscovery` using Zeitwerk rather than `ConstantResolver` as a proof of concept.

First I refactored to encapsulate `ConstantResolver` so it is easier for me to play with an alternate `ConstantDiscovery` implementation.

Then I implemented a Zeitwerk based `ConstantDiscovery` and copied tests from `ConstantDiscovery` and the `ConstantResolver` gem to make sure it worked ok. Private methods for matching constants from the map of paths was taken from [ConstantResolver](https://github.com/Shopify/constant_resolver/blob/main/lib/constant_resolver.rb#L143-L164) as well as many tests checking edge cases.

Finally I replaced `ConstantDiscovery` with my version and removed the code related to the old implementation. A lot of test helpers needed to be updated here since now `ConstantDiscovery` depends on Zeitwerk Loaders 😭 

- The second commit contains the zeitwerk implementation as a separate class as well as extended test coverage. I think this should be the focus since the third commit just rigs it all up so I could test on my repo.
- Is it okay to depend on Zeitwerk in this way?

## Type of Change

- [ ] Bugfix
- [ ] New feature

## Checklist


Catsuko:
I have signed the CLA!

lib/packwerk/constant_discovery.rb
Catsuko, line 1: For an easier to understand diff, see [this commit](https://github.com/Shopify/packwerk/pull/410/commits/ff62ac18e90d50f4e63ea81027b0897f3aecc559#diff-8aa9e8fd4e164114f7d838b524d6e4edaf76a68f05dc4b9996fd5de8944a74e3).

jmcfarland-figma:
👋 We'd love if this got merged, it would give us a lot more flexibility in how we organize our files.

PR #423, opened 2025-01-17
Clarify dependency violation message

derekprior:

I was bothered by a couple of things in the dependency violation message and this change fixes them.

1. "Are we missing an abstraction?" is not helpful context.
2. "Is the code making the reference, and the referenced constant, in the right packages?" was not grammatically correct or clear.

I edited the copy and updated the tests.

Whether you think this error message is more clear or helpful

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist


gmcgibbon (approved):
This seems clearer to me. Thank you

PR #431, opened 2025-03-28
Handle upcoming prism deprecations

Earlopain:

Avoid prism deprecation warnings:
* https://github.com/ruby/prism/issues/3524
* https://github.com/ruby/prism/pull/3501

Use the non-deprecated classes. `Parser34` is available since 0.24.0 and can be used without concern. The builder is somewhat new, so I use `defined?` to check it and left a note for when it can be removed.

I left more context in the two commit messages.

## Type of Change

- [ ] Bugfix
- [ ] New feature

### Additional Release Notes

## Checklist

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -9,7 +9,7 @@ PATH
       constant_resolver (>= 0.3)
       parallel (< 2)
       parser
-      prism (>= 0.25.0)
+      prism (>= 1.4.0)
       sorbet-runtime (>= 0.5.9914)
       zeitwerk (>= 2.6.1)
 
--- a/packwerk.gemspec
+++ b/packwerk.gemspec
@@ -48,7 +48,7 @@ Gem::Specification.new do |spec|
   # For Ruby parsing
   spec.add_dependency("ast")
   spec.add_dependency("parser")
-  spec.add_dependency("prism", ">= 0.25.0")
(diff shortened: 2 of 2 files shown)


Earlopain:
Prism 1.5.0 has been released and will now warn when an incompatible builder class is used. `Prism::Translation::Parser` was a misunderstanding from me. It is supposed to always parse as the latest version that prism knowns about.

Probably it should use `Prism::Translation::ParserCurrent` but don't have enought context to tell if that is correct so I left that change out for now.

NielsKSchjoedt (approved):

NielsKSchjoedt:
Is this bound to go out soon?

hypermkt:
@gmcgibbon @rafaelfranca 
Hi maintainers,

Just a quick note that this PR has been approved and fixes the Prism deprecation warning we’re encountering.
```
warning: [deprecation]: The builder passed to `Prism::Translation::Parser.new` is not a `Prism::Translation::Parser::Builder` subclass. This will raise in the next major version.
```

When you have time, could you please merge it and include it in a future release?

Thanks you!

hmcguire-shopify (approved):

hmcguire-shopify:
Switched the builder in https://github.com/Shopify/packwerk/pull/451, but I think your change to update the minimum prism version makes sense as well 👍

PR #438, opened 2025-06-19
Remove unused attribute from `NodeProcessorFactory`

ryym:

The `context_provider` attribute of `NodeProcessorFactory` seems to be unused so I want to delete it.

I just deleted it. Both typecheck and test passed.

That attribute was initially added by https://github.com/Shopify/packwerk/pull/63. At that time, `ReferenceExtractor` requires `context_provider` . But in #169, `context_provider` was removed from the argument of `ReferenceExtractor.new` .

If `context_provider` is used in some way or will be used in the future, please close this PR.

## Type of Change

- [ ] Bugfix
- [ ] New feature

## Checklist

### DIFF
--- a/lib/packwerk/node_processor_factory.rb
+++ b/lib/packwerk/node_processor_factory.rb
@@ -6,7 +6,6 @@ module Packwerk
     extend T::Sig
 
     const :root_path, String
-    const :context_provider, ConstantDiscovery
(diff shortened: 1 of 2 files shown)


ryym:
I have signed the CLA!

gmcgibbon (approved):
Thank you!

PR #447, opened 2026-04-13
Replace core parsing/resolution engine with Rubydex

paracycle:
## Summary

Replace the per-file AST walking, constant extraction, and resolution pipeline with [Rubydex](https://github.com/Shopify/rubydex), a high-performance Ruby indexer written in Rust. Constants are resolved from actual parsed definitions rather than Zeitwerk path conventions, enabling more thorough analysis with comparable speed.

- Replaces `constant_resolver`, `parallel`, and direct `parser`/`ast` deps with `rubydex`
- Replaces `better_html` with [Herb](https://github.com/marcoroth/herb) for ERB parsing (preserves block-spanning tags and source positions)
- Drops the runtime Rails dependency entirely (used only for old `constant_resolver` autoload paths)
- Removes the MD5-based file cache (Rubydex is fast enough without it)
- Removes the `parallel` gem (fork overhead exceeded the work being done)

## Architecture

**Before:** for each file → parse with Prism → walk AST → extract constant references → resolve via `constant_resolver` (Zeitwerk paths) → check violations

**After:** index all workspace files in one `Rubydex::Graph` batch call → resolve all constants → iterate `Declaration#references` per declaration → check violations. Association detection (`has_many`, `belongs_to`, etc.) runs as a supplementary Prism-based pass since Rubydex treats those as method calls, not constant references — and is targeted to only files that contain association method references via `graph.method_references` (~1% of files on Shopify Core).

## Dependency changes

| Removed | Added |
|---|---|
| `constant_resolver` | `rubydex` |
| `parallel` | `herb` |
| `better_html` | |
| `ast` (direct) | |
| `parser` (direct) | |

Kept: `prism` (for association detection), `activesupport`, `bundler`, `benchmark`.

## Key design decisions

1. **Full workspace indexing:** `index_and_resolve` indexes ALL Ruby files for cross-package resolution, even when a scoped check (`packwerk check components/foo`) is requested. Only the scoped files are checked for violations.
2. **Per-declaration iteration:** the hot loop walks `graph.declarations` and uses `Declaration#references` rather than iterating all 9M+ constant references globally. Skips declarations with no in-scope references entirely.
3. **URI-keyed lookups:** the hot loop compares URIs as opaque strings via precomputed hash maps (`@checked_uris`, `@uri_to_package`, `@uri_to_relative_path`). No URI parsing in the inner loop.
4. **Singleton class deduplication:** `Foo.bar` produces references to both `Foo` and `Foo::<Foo>` (the singleton class). Skipping `Rubydex::SingletonClass` declarations gives each user-visible reference once.
5. **Zeitwerk-canonical target paths:** when a constant has multiple definitions (e.g. canonical file + class reopenings), prefer the file matching the Zeitwerk naming convention (`ApiClient` → `app/models/api_client.rb`) for the violation message's "defined in" hint.
6. **Shared namespace handling:** if any definition of a constant is in the source file's package, the reference is treated as local (not a violation). Correctly handles cases like `module GraphApi` reopened across many packages.
7. **Targeted ERB and association parsing:** ERB files are extracted via Herb's `extract_ruby` which preserves line numbers, then fed into `graph.index_source`. Association detection only re-parses files that `graph.method_references` shows contain association method calls.
8. **0/1-based location conversion:** Rubydex uses 0-based line/column, Prism uses 1-based line + 0-based column, Packwerk displays 1-based for both. Conversions happen at the boundary, with regression tests asserting reported locations match the 1-based display convention for both source paths.

## Performance on Shopify Core (~60k files)

We're within 3-4s of released speed while doing significantly more analysis. The 31k extra offenses come from references the old `ConstNodeInspector`/`constant_resolver` pipeline missed — primarily ERB files (where the old concatenated-Ruby approach often produced unparseable output) and shared namespace references. After running `packwerk update-todo`, only ~620 strict-mode violations remain (which need manual resolution since strict packages reject new entries in `package_todo.yml`).

## Deleted code

16 source files, 11 test files, 3 RBI files removed:

- AST walking pipeline: `file_processor`, `node_processor`, `node_processor_factory`, `node_visitor`, `node_helpers`
- Constant extraction: `const_node_inspector`, `constant_name_inspector`, `constant_discovery`, `parsed_constant_definitions`, `reference_extractor`, `unresolved_reference`, `association_inspector`
- Per-file caching: `cache`
- Parsers: `parsers/ruby`, `parsers/factory`, `parsers/parser_interface`
- Rails integration: `rails_load_paths`

## Tests

134 tests pass (was 132 + 2 new integration tests for line/column locations). Sorbet typecheck clean. Rubocop clean.

The new line/column integration tests assert reported locations for both code paths (Rubydex source for plain constant references, Prism source for ActiveRecord associations) and catch off-by-one regressions.

## Commits

- Replace core parsing/resolution engine with Rubydex
- Optimize Rubydex pipeline for correctness and performance
- Refactor \`extract_refs_by_file\` into focused helpers
- Convert Rubydex/Prism 0-based locations to Packwerk's 1-based convention
- Refactor: URI-keyed lookups, named types, simpler helpers
- Replace better_html with Herb for ERB parsing
- Bump gem RBI files


exterm:
Cool stuff, Ufuk. Would this mean packwerk doesn't depend on the interrogated codebase using zeitwerk anymore?

paracycle:
> Cool stuff, Ufuk. Would this mean packwerk doesn't depend on the interrogated codebase using zeitwerk anymore?

That's correct. Rubydex can do proper Ruby constant resolution, so we don't need to use Zeitwerk heuristics to figure out what constant references resolve to based on their filename. At least, we shouldn't, and, if there are any problems with the resolution, then we should fix them in Rubydex.

exterm:
The [blog announcement](https://railsatscale.com/2026-05-12-one-engine-many-tools/) mentions `about 3x more constant references resolved in our monolith`. I don't know how the codebase has developed over the last 4 years but I used `packwerk` pretty heavily on it before that and I'm sure we were not accidentally missing over 60% of constant references. Can you expand on what references this new implementation finds that the old one didn't?

I'm worried there may be a change in semantics here.

exterm:
Thank you @paracycle , that explains it.

paracycle:
> The [blog announcement](https://railsatscale.com/2026-05-12-one-engine-many-tools/) mentions `about 3x more constant references resolved in our monolith`. I don't know how the codebase has developed over the last 4 years but I used `packwerk` pretty heavily on it before that and I'm sure we were not accidentally missing over 60% of constant references. Can you expand on what references this new implementation finds that the old one didn't?
> 
> I'm worried there may be a change in semantics here.

Valid concern, thanks for asking me to dig deeper here.

I got an LLM to analyze the new offenses flagged on with this branch and made it add 3 new tests in [this commit](https://github.com/Shopify/packwerk/pull/447/commits/1082bf17e61d0ed5be271db468425ce548f38436). From what I understand, most of the new violations are from us flagging nested references as separate violations, which are much more prevalent in ERB files.

This does seem to be a change in behaviour, but I am not sure if this is better or worse.

exterm:
From my perspective it is worse (steep increase in number of violations without adding a lot of signal, as these are the same violation at a different level of abstraction) but I am only one user in the end

paracycle:
@exterm There seems to be another category of violations that account for about 9000 new violations in Core:

> So `Shopify::Foo` is defined in `lib/shopify/foo.rb` which lives in the root package `.` Any reference to it from components/X becomes a violation against `.`
>
> The old pipeline:
> - `Shopify::Foo.foo` → extract just Shopify
> - `constant_resolver` looks for `shopify.rb` in autoload paths
> - Probably doesn't find a definition (because `lib/` isn't typically in autoload paths) → silently skipped
> - Result: zero violation
> 
> The new pipeline:
> - `Shopify::Foo.foo` → extract `Shopify::Foo` (after collapse, the leaf)
> - Rubydex finds `Shopify::Foo` defined in `lib/shopify/foo.rb`
> - Source package is `components/baz`, target package is `.`
> - Result: dependency violation against `.`
> 
> This is a real cross-package reference that the old pipeline missed because it didn't know `lib/shopify/` defines `Shopify::Foo`.

What do you think about this one?

exterm:
I think there is something wrong with that analysis (I don't see why packwerk on current `main` would extract only `Shopify` from `Shopify::Foo.foo`; it should extract `Shopify::Foo`).

However there is also something very valid in there - Packwerk with `constant_resolver` can only resolve constants defined in autoloaded code. That was an implementation limitation, not the preferred behavior. So, barring a misunderstanding on my side, these 9000 new violations are probably valid.

I believe modern Rails applications autoload `lib` by default, in which case it will make much less of a difference. Still, I see it as a considerable improvement if packwerk can now trace constants to non-autoloaded code.

swiknaba:
closes #467

PR #448, opened 2026-04-21
Fix Prism 1.9.0 deprecation: use Prism::Translation::Parser::Builder as builder superclass

dduugg:

Fix a `Prism 1.9.0` deprecation warning that causes `packwerk` to fail with an internal error when used alongside Rails with `ActiveSupport` deprecation notifications enabled, fix all resulting CI failures, restore `bundle exec tapioca gems --all` to a working state on Ruby 3.4, fix Sorbet type-checking errors introduced by the sorbet 0.6.x upgrade, and update `packwerk.gemspec` dependency bounds to accurately reflect the minimum prism and sorbet-runtime versions required by the current code.

Prism 1.9.0 added a check in `Prism::Translation::Parser#initialize` ([source](https://github.com/ruby/prism/blob/v1.9.0/lib/prism/translation/parser.rb#L74-L82)) that emits a `warn()` when the builder is not a `Prism::Translation::Parser::Builder` subclass:

```
DEPRECATION WARNING: The builder passed to `Prism::Translation::Parser.new` is not a
`Prism::Translation::Parser::Builder` subclass. This will raise in the next major version.
```

When Rails is configured with `config.active_support.deprecation = :raise` (common in CI environments), this `warn` is converted to an `ActiveSupport::DeprecationException`, which packwerk's `rescue StandardError` in `FileProcessor#call` catches and surfaces as "Packwerk encountered an internal error."

### Core fix (`lib/packwerk/parsers/ruby.rb`, `packwerk.gemspec`)

Change `TolerateInvalidUtf8Builder` to inherit from `Prism::Translation::Parser::Builder` instead of `Parser::Builders::Default` directly. `Prism::Translation::Parser::Builder` is itself a subclass of `Parser::Builders::Default`, so all existing behavior (including the `string_value` override) is fully preserved. This satisfies Prism's new requirement.

Bump the minimum Prism version in `packwerk.gemspec` from `>= 0.25.0` to `>= 1.4.0`. `Prism::Translation::Parser::Builder` was [introduced in prism 1.4.0](https://github.com/ruby/prism/commit/b080e608a8de3b30095a61bb54c0d3e747aa0619), so this is the accurate minimum that makes `TolerateInvalidUtf8Builder`'s new superclass available.

### CI fixes (four independent root causes)

**1. Lint / Type Checking — stale Gemfile.lock**

The prism constraint in the `PATH` section of Gemfile.lock still read `>= 0.25.0` after the gemspec was changed to `>= 1.4.0`. Bundler's deployment mode (used by `ruby/setup-ruby` with `bundler-cache: true`) rejects installs when gemspec metadata has changed since the lock was written. Fix: regenerate and commit the updated Gemfile.lock.

**2. Type Checking — `Prism::Translation::Parser::Builder` unknown to Sorbet**

With the lockfile now resolving prism 1.9.0, `sorbet/rbi/gems/prism@1.9.0.rbi` includes `Prism::Translation::Parser::Builder` directly, so no shim is needed. (An earlier iteration of this PR carried a `sorbet/rbi/shims/prism.rbi` shim for when prism 0.27.0 was locked; that shim has been removed.)

**3. Tests Ruby 3.2 / 3.3 — `Object#stub` not defined**

Fresh CI bundle installs (lockfile deleted) resolve **mocha 3.x** on Ruby ≥ 3.2 (mocha 3.0 raised its minimum Ruby requirement to 3.2), and simultaneously **minitest 6.x** (same minimum). Two breaking changes combine:
- Minitest 6 extracted `minitest/mock` (which defines `Object#stub`) to a new standalone gem `minitest-mock`.
- Mocha 3.x no longer loads `minitest/mock` as a side-effect when required.

As a result, `Object#stub` is undefined and `ExtensionLoader.stub(:require, ...)` raises `NoMethodError`. Ruby 3.1 escapes this because both mocha 3.x and minitest 6 require Ruby ≥ 3.2, so it falls back to mocha 2.x + minitest 5.x where `minitest/mock` is built-in.

Fix: add `gem "minitest-mock"` to the Gemfile and `require "minitest/mock"` explicitly in `test/test_helper.rb`. This works for minitest 5 (where mock is still bundled) and minitest 6 (where the standalone gem provides it).

**4. Tests Ruby 3.4 — `Hash.new` keyword argument rejection**

Ruby 3.4 tightened `Hash.new`: bare `Hash.new(name: "boop")` is now treated as passing the unknown keyword `name:`, raising `ArgumentError`. Fix: use explicit braces — `Hash.new({ name: "boop" })`.

### Ruby version matrix update

Drop EoL Ruby 3.1 and 3.2 from the CI matrix and add 3.4 and 4.0, maintaining three consecutive versions. Update the `loading-tests` matrix from (3.1, 3.2) to (3.3, 3.4) in parallel.

### tapioca gems --all on Ruby 3.4 (`Gemfile`, `Gemfile.lock`, `sorbet/tapioca/require.rb`, `sorbet/rbi/gems/`)

Three issues prevented `bundle exec tapioca gems --all` from running on Ruby 3.4:

1. **`mutex_m` removed from stdlib** — Ruby 3.4 removed `mutex_m` from the standard library but Rails still uses it without declaring a gem dependency. Fix: add `gem "mutex_m"` to the Gemfile.

2. **tapioca/sorbet incompatibility** — tapioca 0.13.3 calls `signature.has_rest` on a sorbet `Signature` object, but that method was removed in sorbet 0.6.x. Fix: upgrade tapioca to 0.19.1 which uses `rest_name` instead. sorbet also upgraded from 0.5.x to 0.6.x as a transitive requirement of tapioca 0.19.1.

3. **minitest at_exit hook** — tapioca 0.19.1 loads `sorbet/tapioca/require.rb` in-process, so `require "minitest/autorun"` registers an `at_exit` hook that fires when tapioca exits and tries to parse tapioca's CLI args (e.g. `--all`) as minitest options, causing a non-zero exit code. Fix: change `require "minitest/autorun"` to `require "minitest"` in the require file. Gem RBI generation only needs minitest loaded for type introspection, not its test runner. Also regenerate `sorbet/tapioca/require.rb` via `tapioca require` under tapioca 0.19.1, removing test-support paths that are not on the load path during gem compilation.

### Sorbet 0.6.x type signature fixes (`lib/packwerk/configuration.rb`, `lib/packwerk/rails_load_paths.rb`, `lib/packwerk/run_context.rb`, `test/support/stub_const.rb`)

sorbet 0.6.x (specifically >= 0.6.12698, which introduced `T::Module` in [eaf645232c](https://github.com/sorbet/sorbet/commit/eaf645232c)) now requires `Module` to be parameterized in type signatures — bare `Module` is a malformed type declaration in `# typed: strict` files (enforcement added in sorbet commit 0f3e8b98bf, released 2025-11-13). Applied `srb tc -a` to replace all 12 occurrences of `Module` with `T::Module[T.anything]` across four files.

Bump `sorbet-runtime` minimum in `packwerk.gemspec` from `0.5.9914` to `0.6.12698`. The `T::Module[T.anything]` type annotation is evaluated at runtime when sorbet checks sigs; older versions would raise `NameError` since the class did not exist.

- `lib/packwerk/parsers/ruby.rb:31` — superclass change for `TolerateInvalidUtf8Builder`. Inheritance chain: `TolerateInvalidUtf8Builder < Prism::Translation::Parser::Builder < Parser::Builders::Default`
- `packwerk.gemspec` — Prism lower bound `0.25.0` → `1.4.0` (when `Prism::Translation::Parser::Builder` was introduced); sorbet-runtime lower bound `0.5.9914` → `0.6.12698` (when `T::Module` was introduced, required at runtime by the updated sig annotations)
- `test/test_helper.rb` + `Gemfile` — explicit `minitest/mock` load and `mutex_m` gem for Ruby 3.4 / minitest 6 compatibility
- `test/unit/packwerk/package_test.rb` — `Hash.new` braces fix for Ruby 3.4
- `.github/workflows/ci.yml` — matrix rotation: (3.1, 3.2, 3.3) → (3.3, 3.4, 4.0)
- `Gemfile` — `mutex_m` added; tapioca upgraded 0.13.3 → 0.19.1
- `sorbet/tapioca/require.rb` — `minitest/autorun` → `minitest`; test-support paths removed; regenerated under tapioca 0.19.1
- `lib/packwerk/configuration.rb`, `lib/packwerk/rails_load_paths.rb`, `lib/packwerk/run_context.rb`, `test/support/stub_const.rb` — `Module` → `T::Module[T.anything]` in sigs (sorbet 0.6.x requirement)
- `sorbet/rbi/gems/` — full regeneration reflecting gem version upgrades (tapioca 0.13.3 → 0.19.1, sorbet 0.5.x → 0.6.x, railties 7.0.x → 7.2.x, prism 0.27.0 → 1.9.0, and transitive deps)

## Type of Change

- [x] Bugfix
- [ ] New feature

## Checklist


rafaelfranca:
CI is failing

dduugg:
@rafaelfranca TY for kicking of a build and flagging the failures. I've done what seemed necessary for CI to pass, although it has greatly expanded the scope of this PR. Happy to decompose it into multiple pieces, let me know what would be easiest for the reviewer. 🙇

dduugg:
@rafaelfranca Just checking in again to see what I can do to help this PR along. Please let me know. 🙇

hmcguire-shopify:
Fixed in https://github.com/Shopify/packwerk/pull/451, thank you for the pull request 🙏

PR #455, opened 2026-05-01
Migrate Sorbet signatures to RBS comments

Morriar:
## Summary

Migrate all Sorbet `sig` blocks and type annotations to [RBS comment signatures](https://sorbet.org/docs/rbs-support), allowing us to drop the runtime dependency on `sorbet-runtime`.

The main goal is to **remove the `sorbet-runtime` dependency** from the gem. This eliminates the runtime overhead of signature validation and method wrapping, **boosting performance** for all Packwerk users — especially on large codebases where the cost of `sorbet-runtime` adds up across thousands of file checks.

### What changed

* Translate all `sig` blocks to `#:` RBS comment signatures
* Translate `T.let`, `T.cast`, `T.must`, `T.bind`, `T.unsafe` to RBS comment equivalents (`#: as Type`, `#: as !nil`, etc.)
* Translate `T.type_alias` to `#: type name = ...` RBS syntax
* Migrate `T::Struct` subclasses to plain Ruby classes
* Remove `extend T::Sig`, `extend T::Helpers`, `T::Generics` usage
* Remove `sorbet-runtime` from gemspec dependencies
* Remove `disable_sorbet.rb` shim (no longer needed)


Gemfile.lock
```diff
@@ -201,19 +200,19 @@ GEM
...
       rbi (>= 0.3.7)
       require-hooks (>= 0.2.2)
       rubydex (>= 0.1.0.beta10)
-      sorbet-static-and-runtime (>= 0.5.11087)
+      sorbet-static-and-runtime (>= 0.6.12698)
       spoom (>= 1.7.9)
       thor (>= 1.2.0)
       tsort
-    thor (1.5.0)
+    thor (1.4.0)
```
hmcguire-shopify: random downgrade?

Gemfile.lock
```diff
@@ -151,7 +150,7 @@ GEM
     regexp_parser (2.12.0)
     reline (0.6.1)
       io-console (~> 0.5)
-    require-hooks (0.4.0)
+    require-hooks (0.3.0)
```
hmcguire-shopify: ?

Gemfile.lock
Morriar: Good catch, not sure why bundler decided to downgrade some gems because I remove `sorbet-runtime`

hmcguire-shopify: Downgrades are still there

lib/packwerk/parsers/erb.rb
```diff
@@ -40,18 +39,15 @@ def parse_buffer(buffer, file_path:)
...
       def to_ruby_ast(erb_ast, file_path)
         # Note that we're not using the source location (line/column) at the moment, but if we did
         # care about that, we'd need to tweak this to insert empty lines and spaces so that things
         # line up with the ERB file
-        code_pieces = T.must(code_nodes(erb_ast)).map do |node|
-          T.cast(node, ::AST::Node).children.first
+        code_pieces = code_nodes(erb_ast) #: as !nil
+          .map do |node|
+          node #: as ::AST::Node
+            .children.first
```
paracycle: I think there is an identation problem here:

```suggestion
        code_pieces = code_nodes(erb_ast) #: as !nil
          .map do |node|
            node #: as ::AST::Node
              .children.first
```

lib/packwerk/association_inspector.rb
```diff
@@ -4,39 +4,30 @@
...
-    )
-
-    sig do
-      params(
-        inflector: T.class_of(ActiveSupport::Inflector),
-        custom_associations: CustomAssociations,
-        excluded_files: T::Set[String]
-      ).void
-    end
+    #: type custom_associations = Array[Symbol] | Set[Symbol]
```
paracycle: Honestly, I still prefer `customAssociations` over `custom_associations`.

lib/packwerk/association_inspector.rb
```diff
@@ -4,39 +4,30 @@
...
+    RAILS_ASSOCIATIONS = [
+      :belongs_to,
+      :has_many,
+      :has_one,
+      :has_and_belongs_to_many,
+    ].to_set #: custom_associations
+
+    #: (
+    #|   inflector: singleton(ActiveSupport::Inflector),
+    #|   ?custom_associations: custom_associations,
```
paracycle: Specifically, I think this would read much nicer as:

```suggestion
    #|   ?custom_associations: customAssociations,
```

In its current form, it is very confusing.

paracycle (approved):
Other than the lockfile changes that @hmcguire-shopify pointed out, I think this is great!

lib/packwerk/association_inspector.rb
rafaelfranca: Snake case in Ruby is usually a no go. And I guess we can't do use the case we use for constants because it is also confusing.

I think we should just change this type to be `Set[Symbol]` or `Array[Symbol]`.

paracycle: Sure, I get that snake case isn't used in Ruby, and that's precisely why I am advocating for type aliases to be snake-cased, since there is no chance of them to be confused with anything in Ruby. They are, after all, purely RBS constructs that don't exist in Ruby. That's also the reason why we can't use constant names, since they could silently shadow each other.

See below for a confusing example.

rafaelfranca: Yeah. I think that is what we should do. In this case, we should just use `Set[Symbol]`

PR #467, opened 2026-07-11
feat: allow configuring load_paths for non-Rails (Zeitwerk) apps

swiknaba:

I want to use it in my non-Rails (but Zeitwerk) apps. Packwerk is like 99% there already. This adds the last missing piece to run the check without loading Railties.

There was actually just 1 part that blew up when running this on a non Rails app: file path resolution relied on the Rails load path. It now allows to define load paths in the `packwerk.yml` file.

I also noticed two other minor bugs that blew up when Rails wasn't present. See commit messages :)

=> POC here: https://github.com/swiknaba/kirei/pull/39

=> if #410 get's merged, my PR should be trivial to update. Would need to return a `Zeitwerk::LOader` object instead of the `Hash[String, Module]`.

=> #447 makes my PR obsolete if I see it right, since it also removes any hard dependency on Railties. Feel free to discard my PR if 447 gets merged soon anyway.

no clue what to write here. Tiny change I believe. Thanks for making this open source <3

## Type of Change

- [x] Bugfix <= two tiny 1-liner fixes snugged in
- [x] New feature <= main thing here

### Additional Release Notes

## Checklist

### Before

=> from my dummy test app, see link "POC" above

before-failing

### After

=> from my dummy test app, see link "POC" above

after-works

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -203,7 +203,7 @@ GEM
       rexml (>= 3.2.6)
       sorbet-static-and-runtime (>= 0.5.10187)
       thor (>= 0.19.2)
-    spring (4.0.0)
+    spring (4.7.0)
     stringio (3.1.7)
     tapioca (0.19.1)
       benchmark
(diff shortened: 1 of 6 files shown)


swiknaba:
I have signed the CLA!

Gemfile.lock
```diff
@@ -203,7 +203,7 @@ GEM
       rexml (>= 3.2.6)
       sorbet-static-and-runtime (>= 0.5.10187)
       thor (>= 0.19.2)
-    spring (4.0.0)
+    spring (4.7.0)
```
swiknaba, line 206: this is actually needed in non-rails apps, see msg of [2894127](https://github.com/Shopify/packwerk/pull/467/commits/28941271e834518f7c09e44e657b22ff5f3c1927)

